PR #3, opened 2016-04-08
Add a test rangeproof_verify test for insistent values in Pederson commitment and rangeproof

phedny:
When studying Confidential Transaction, at first I didn't see the link between the value using the Pedersen commitment and the one used in the range proof. As part of finding how this link is present, I create the test case in this PR, which fails if the link doesn't exist. It does not fail, so that's a good thing. Just wanted to share this extra test.

### DIFF
--- a/src/modules/rangeproof/tests_impl.h
+++ b/src/modules/rangeproof/tests_impl.h
@@ -1,0 +1,281 @@
+/**********************************************************************
+ * Copyright (c) 2015 Gregory Maxwell                                 *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef SECP256K1_MODULE_RANGEPROOF_TESTS
+#define SECP256K1_MODULE_RANGEPROOF_TESTS
+
+#include "include/secp256k1_rangeproof.h"
+
+void test_pedersen(void) {
+    unsigned char commits[33*19];
+    const unsigned char *cptr[19];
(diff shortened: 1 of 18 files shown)


src/modules/rangeproof/tests_impl.h
```diff
@@ -162,6 +162,11 @@ void test_rangeproof(void) {
     int j;
     int k;
     secp256k1_rand256(blind);
+    CHECK(secp256k1_pedersen_commit(ctx, commit, blind, 5));
+    len = 5134;
+    CHECK(secp256k1_rangeproof_sign(ctx, proof, &len, 0, commit, blind, commit, 0, 0, 7));
+    CHECK(len <= 5134);
+    CHECK(!secp256k1_rangeproof_verify(ctx, &minv, &maxv, commit, proof, len));
```
apoelstra: Minor nit: can you compare against one like `!= 1` rather than using `!` at the beginning?

src/modules/rangeproof/tests_impl.h
```diff
@@ -162,6 +162,11 @@ void test_rangeproof(void) {
     int j;
     int k;
     secp256k1_rand256(blind);
+    CHECK(secp256k1_pedersen_commit(ctx, commit, blind, 5));
```
apoelstra: Can you add a comment above these lines explaining that you're checking that a when the commit value and claimed commit value don't match, the resulting signature is invalid?

apoelstra:
I'm going to close this because it's out-of-date, and because this test seems cryptographically impossible to fail.

PR #4, opened 2017-04-24
Implement sound rangeproofs

apoelstra:
Implements a [design by Oleg Andreev](https://lists.launchpad.net/mimblewimble/msg00098.html) to create sound rangeproofs with identical size and format to the original Elements ones (though of course, neither will validate as the other).

**Edit:** This is not intended for merge to this repo, only review. See [the secp256k1-mw repo](https://github.com/apoelstra/secp256k1-mw/pull/1) where I had intended to push this, had Github not silently redirected which repo I was on when I clicked "Create Pull Request". But we'll keep discussion here, since it's started.


oleganza:
Note that this multi-generator trick works only for a single-asset blockchain, but if you need to balance values per-asset, and assets are themselves blinded, then you cannot commit to a shared blinding factor as a second part of the commitment. 

So this:

```
V = v*H  +f*G
F = f*J (shared among all digits)
```

won't work if `H = A + b*G`. You'd actually need:

```
Asset commitment (H,B):

H = A + b*G
B = b*J

Value commitment (V,F):
V = v*H + f*G
F = v*B + f*J
```

Now, balancing Vs and Fs will guarantee that `v_i*A_j` balance for per assets (there is an extra requirement for As being random and `v` being of relatively short range, like < 64 bits). But having value encoded in the F commitment makes it non-reusable for individual digits.

Quantum computers are %@#$!@#^@.

apoelstra:
Oleg, if I have `H = A + bG` then a rangeproof with asset generator `H` is just as good as one with asset generator `A`.... which is good, because the latter would require I expose the secret `A`!

Since the soundness changes affect only the rangeproof, and the total balancing equation is independent of this, I don't see why it should matter at all if I have multiple assets.

oleganza:
Hmm...

Lets say you commit to `f` alone in `F`, and we assume the asset commitment is soundly proven to be within a proper set (so since very issuance the )

```
Asset commitment (H,B):
  H = A + b*G
  B = b*J

Value commitment (V,F):
  V = v*H + f*G
  F = f*J
```

Therefore, `V = v*A + (v*b + f)*G`. Balancing Fs and Vs would means that these things balance:

```
v*(a + b)*G   // A = a*G, where dlog `a` is PQ-known 
```

Since `b` (unlike `v`) is scalar with unrestricted range, it is easy for you to find `b'` so that `a' + b' == a + b` and morph the commitment to a different asset `a'`. Due to soundness of the asset range proofs (surjection proofs), you'd need to have a little bit of `a'` on the input, but then in the output you can assign `v` to `a'`.

So the attack looks like this:

1. Create asset `A` that you control.
2. Break ECDLP to extract `a` from `A`.
3. Choose asset `A'` that you do not control.
4. Break ECDLP to extract `a'` from `A'`.
5. Issue `v` units of `A`.
6. Acquire at least one unit of `A'`.
7. Create a tx with `v*A` and `1*A'` on the inputs.
8. Compute `b' = a - a' + b`.
9. Create output with `v*A'` and blind `A'` with factor `b'`.
10. Create output with `1*A` and blind `A` with factor `b`.
11. Now you swapped 1000000 units of dummy asset and 1 unit of USD for 1000000 USD and 1 dummy asset.

I'd be glad to be proven wrong to regain that 20% in space savings.

PR #10, opened 2017-07-10
Add n_keys argument to whitelist_verify

### LINKED ISSUE #9: Whitelisting: ensure that online and offline key arrays have size n_keys
If the array size is smaller than n_keys there is an out of bounds read in `whitelist_sign` and `whitelist_verify`. Either it must be clearly stated in the documentation that before `whitelist_verify` the caller must check that the length of the online and offline arrays are equal to `secp256k1_whitelist_signature_n_keys(sig)`. Or both functions need additional arguments for the sizes of the online and offline key arrays to do that check inside the function.

jonasnick:
Fixes #9 
Interestingly, documentation for the n_keys argument was already present. Commit also adds a test.


apoelstra:
I was going to ask you to update the API tests, but there aren't any for this module :(. I'll write some, maybe that will suss out the remaining API problems.

apoelstra:
ACK

PR #11, opened 2017-07-10
Fix pedersen_blind_generator_blind_sum return value documentation

jonasnick:


instagibbs:
What are the possible failure conditions?

jonasnick:
https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/rangeproof/main_impl.h#L152 or when the blinding values don't fit in a scalar.

instagibbs:
If you could detail that in the comments(at some high level) that'd be great

jonasnick:
Added more detail to comment and added similar comment to other functions in the module.

include/secp256k1_rangeproof.h
```diff
@@ -61,8 +61,10 @@ SECP256K1_API int secp256k1_pedersen_commitment_serialize(
 void secp256k1_pedersen_context_initialize(secp256k1_context* ctx);
 
 /** Generate a pedersen commitment.
- *  Returns 1: commitment successfully created.
- *          0: error
+ *  Returns 1: Commitment successfully created.
+ *          0: Error. The blinding factor is larger than the group order
+ *             (probability for random 32 byte number < 2^-127) or 0. Retry with a
```
instagibbs: "or 0"? That is a valid blinding factor, and we use it in elements.

jonasnick: you're right this is only a problem when value is 0 too. Fixed

instagibbs:
ACK https://github.com/ElementsProject/secp256k1-zkp/pull/11/commits/619b8f16dc6fd5b327a03b076523ae45c727a0f8

jonasnick:
ping @apoelstra

PR #12, opened 2017-08-16
Fix generator makefile

instagibbs:
Include test_impl.h


src/modules/generator/Makefile.am.include
```diff
@@ -1,5 +1,6 @@
 include_HEADERS += include/secp256k1_generator.h
 noinst_HEADERS += src/modules/generator/main_impl.h
+noinst_HEADERS += src/modules/generator/test_impl.h
```
jonasnick: typo, should be tests_impl

instagibbs:
@jonasnick fixed, somehow didn't push the correct fix

jonasnick:
tACK

PR #20, opened 2018-05-16
Minor bugfix. Wrong length due to NUL character.

datavetaren:
I stumbled across this when I compiled secp256k1 for C++. Easiest fix is to change length to 17.
The alternative is to use { ... } syntax for C data initialization.


apoelstra:
IIRC the C standard states that if you assign a 16-character string to a 16-wide char array, it simply won't be NUL-terminated. This is fine.

instagibbs:
https://stackoverflow.com/a/13490908

oleganza:
> IIRC the C standard states that if you assign a 16-character string to a 16-wide char array, it simply won't be NUL-terminated. This is fine.

Sorry! _Hides._

image

apoelstra:
Cool, thanks! I didn't know the difference in C++.

LOL Oleg.

PR #16, opened 2017-11-27
WIP: bulletproofs

apoelstra:
Based on rebase of strauss-multiexp from upstream onto -zkp

~~TODO rangeproof aggregation~~
~~TODO const time proving~~
~~TODO pippenger support~~
TODO 48-bit (and generally non-power-of-2) rangeproofs and aggregates
~~TODO general arithmetic circuit support~~

PR #23, opened 2018-05-25
Bulletproofs (rangeproofs only)

apoelstra:
This is #16 without any of the arithmetic circuit stuff.

At this point there are probably still obvious code cleanliness things to be dealt with, but I think it's ready for review. I'd like to get this merged so that we can get the rangeproofs at least into Elements soon.

cc @sipa @real-or-random @jonasnick can a couple of you review this? I'm happy to split it into multple PRs if that makes sense.


include/secp256k1_commitment.h
```diff
@@ -0,0 +1,165 @@
...
+
+/** Opaque data structure that stores a Pedersen commitment
+ *
+ *  The exact representation of data inside is implementation defined and not
+ *  guaranteed to be portable between different platforms or versions. It is
+ *  however guaranteed to be 33 bytes in size, and can be safely copied/moved.
+ *  If you need to convert to a format suitable for storage or transmission, use
+ *  secp256k1_pedersen_commitment_serialize and secp256k1_pedersen_commitment_parse.
+ *
+ *  Furthermore, it is guaranteed to identical signatures will have identical
```
rodasmith: (nit) s/to/two/

real-or-random:
Discussions about ciphers can be annoying... Is there a specific reason to use chacha20? It's just used to derive randomness, why not rely on SHA256? I don't think that performance is a concern here. If performance is a concern, then I guess that AES is the canonical choice because of its hardware implementations (and constant-time software implementations are pretty good too).

I know @sipa has looked into these things, so I could imagine that all of this has been discussed extensively and I'm just late to the party.

It's also interesting for dicemix, which requires a PRG (or better a PRF) and needs much more randomness, each party needs O(|m| * n^2) bytes for n parties. So the performance of the PRF could be worth to think about and it would be preferable to use something which is already in libsecp256k1 of course. 

Am 26. Mai 2018 00:22:25 MESZ schrieb Andrew Poelstra <notifications@github.com>:

apoelstra:
It was performance motivated. The prover derives 64 * n_gates bytes of randomness, so there was a measurable performance improvement from using chacha. In the rewinding failure case, deriving randomness is a nontrivial portion of the computation (the whole thing is only a couple microseconds and a SHA2 compression is ~300 ns, so that's 10%).

Interesting observation about AES hardware. In general we haven't done any hardware-specific optimizations in -zkp, though we have assembler implementations of some operations upstream. Another motivation was to keep things simple, and chacha is (a) dead simple and (b) we already had code for it laying around in multiple places.

real-or-random:
Okay I see, that makes sense. ChaCha20's software performance is certainly excellent and the way to go if hardware optimizations are not a goal, and they're indeed a doubly-edged sword: If you have AES-NI, AES is super fast and well, if you don't have it, then ChaCha20 is much faster. So ChaCha20 is a choice that does not prefer peers with a specific architecture [1]. And I think that's the right way to go in systems where everybody should be able to participate. 

(That's by the way interesting for online protocols such as DiceMix. Everybody needs to wait for the slowest peer every round, so if you want to optimize for a lower wall-clock time, you can even make the argument that it makes sense to optimize for slower machines. Though I don't think that matters a lot for protocols where bandwidth is the bottleneck and not computation. But that's another story.)

[1] https://calomel.org/aesni_ssl_performance.html

src/scalar_4x64_impl.h
```diff
@@ -955,4 +956,94 @@ SECP256K1_INLINE static void secp256k1_scalar_mul_shift_var(secp256k1_scalar *r,
...
+        x4 = LE32(seed32[0]);
+        x5 = LE32(seed32[1]);
+        x6 = LE32(seed32[2]);
+        x7 = LE32(seed32[3]);
+        x8 = LE32(seed32[4]);
+        x9 = LE32(seed32[5]);
+        x10 = LE32(seed32[6]);
+        x11 = LE32(seed32[7]);
+        x12 = idx;
+        x13 = idx >> 32;
```
real-or-random: This right shift has undefined behavior if `idx`/`size_t` has 32 bits (or fewer). Same below.

(I guess this is rather a concern in the other file.)

src/modules/generator/main_impl.h
```diff
@@ -14,6 +14,32 @@
...
+      0x59, 0xf2, 0x81, 0x5b, 0x16, 0xf8, 0x17, 0x98
+    }
+};
+
+/** Alternate secp256k1 generator, used in Elements Alpha.
+ *  Computed as the hash of the above G, DER-encoded with 0x04 (uncompressed pubkey) as its flag byte.
+ *  import hashlib
+ *  C = EllipticCurve ([F (0), F (7)])
+ *  G_bytes = '0479be667ef9dcbbac55a06295ce870b07029bfcdb2dce28d959f2815b16f81798483ada7726a3c4655da4fbfc0e1108a8fd17b448a68554199c47d08ffb10d4b8'.decode('hex')
+ *  H = C.lift_x(int(hashlib.sha256(G_bytes).hexdigest(),16))
```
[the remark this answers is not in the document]
apoelstra: Interesting, this must be a new change in sage. Thanks for the fix!

src/scalar_4x64_impl.h
```diff
@@ -955,4 +956,94 @@ SECP256K1_INLINE static void secp256k1_scalar_mul_shift_var(secp256k1_scalar *r,
...
+
+#ifdef WORDS_BIGENDIAN
+#define LE32(p) ((((p) & 0xFF) << 24) | (((p) & 0xFF00) << 8) | (((p) & 0xFF0000) >> 8) | (((p) & 0xFF000000) >> 24))
+#define BE32(p) (p)
+#else
+#define BE32(p) ((((p) & 0xFF) << 24) | (((p) & 0xFF00) << 8) | (((p) & 0xFF0000) >> 8) | (((p) & 0xFF000000) >> 24))
+#define LE32(p) (p)
+#endif
+
+static void secp256k1_scalar_chacha20(secp256k1_scalar *r1, secp256k1_scalar *r2, const unsigned char *seed, size_t idx) {
```
real-or-random: This is related to the comment below: `size_t idx` is somewhat strange. The code below is intended to handle for`idx` values that are longer than 32 bits but then `size_t` is not the right type to do this consistently on every platform, i.e., the 32-bit and 64-bit machines will behave differently (even if  `idx >> 32 == 0` on a 32-bit machine).

My suggestion is to use `uint32_t` here and set `x13 = 0` (and accept that we won't need more than 256 GiB of randomness. Another way is simply to use `uint64_t`.

src/scalar_4x64_impl.h
```diff
@@ -955,4 +956,94 @@ SECP256K1_INLINE static void secp256k1_scalar_mul_shift_var(secp256k1_scalar *r,
...
+    size_t n;
+    size_t over_count = 0;
```
real-or-random: I wouldn't use `size_t` for the two counter variables either (because the values do not express memory sizes). But in this case this is probably just personal taste. ;)

real-or-random:
Does not compile for me.
```
src/bench_bulletproof.c:43:5: error: unknown type name ‘secp256k1_bulletproof_circuit’
     secp256k1_bulletproof_circuit **circ;
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
src/bench_bulletproof.c:44:5: error: unknown type name ‘secp256k1_bulletproof_circuit_assignment’
     secp256k1_bulletproof_circuit_assignment *assn;
     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/bench_bulletproof.c:12:
src/bench_bulletproof.c: In function ‘bench_bulletproof_circuit_setup’:
src/bench_bulletproof.c:109:11: warning: implicit declaration of function ‘secp256k1_bulletproof_circuit_prove’; did you mean ‘secp256k1_bulletproof_rangeproof_prove’? [-Wimplicit-function-declaration]
     CHECK(secp256k1_bulletproof_circuit_prove(data->common->ctx, data->common->scratch, data->common->generators, data->circ[0], data->common->proof[0], &data->common->plen, data->assn, NULL, 0, data->common->nonce, &data->common->value_gen[0], NULL, 0) == 1);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

apoelstra:
Derp, I forgot to commit the circuit-removal from the benchmarks. Fixed.

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,228 @@
...
+#include "group.h"
+#include "scalar.h"
+
+#include "modules/commitment/main_impl.h"
+
+struct secp256k1_bulletproof_generators {
+    size_t n;
+    secp256k1_ge *gens;
+    secp256k1_ge *blinding_gen;
+};
```
real-or-random: An explanation of what is stored at what array index would have helped me a lot, including a comment that says that `blinding_gen` points into `gens[]`

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,228 @@
...
+
+#include "modules/bulletproofs/inner_product_impl.h"
+#include "modules/bulletproofs/rangeproof_impl.h"
+#include "modules/bulletproofs/util.h"
+
+secp256k1_bulletproof_generators *secp256k1_bulletproof_generators_create(const secp256k1_context *ctx, const secp256k1_generator *blinding_gen, size_t n, size_t precomp_n) {
```
real-or-random: `s/secp256k1_bulletproof_generators_create/secp256k1_bulletproof_generators_generate`
would be slightly more consistent with the naming in the generators module

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,185 @@
...
+    const unsigned char* const* proof,
+    size_t n_proofs,
+    size_t plen,
+    const uint64_t* const* min_value,
+    const secp256k1_pedersen_commitment* const* commit,
+    size_t n_commits,
+    size_t nbits,
+    const secp256k1_generator* value_gen,
+    const unsigned char* const* extra_commit,
+    size_t *extra_commit_len
```
real-or-random: `const size_t *extra_commit_len`

.gitignore
```diff
@@ -6,6 +6,9 @@ bench_verify
 bench_schnorr_verify
 bench_recover
 bench_internal
+bench_generator
```
real-or-random: since you're adding `bench_generator', you could also add the missing `bench_whitelist`

apoelstra:
@real-or-random I replaced all the comments in the inner-product verification callback function. It should be possible to follow now.

configure.ac
```diff
@@ -507,14 +526,31 @@ if test x"$enable_experimental" = x"yes"; then
...
-      AC_MSG_ERROR([Rangeproof module requires the generator module. Use --enable-module-generator to allow.])
+      AC_MSG_ERROR([Rangeproof module requires the commitment module. Use --enable-module-commitment to allow.])
+    fi
+    if test x"$enable_module_bulletproof" = x"yes"; then
+      AC_MSG_ERROR([Bulletproof module requires the commitment module. Use --enable-module-commitment to allow.])
+    fi
+  fi
+
+  if test x"$enable_module_generator" != x"yes"; then
+    if test x"$enable_module_bulletproof" = x"yes"; then
```
instagibbs: Move this up since you're already doing a "generator not enabled" check here: https://github.com/ElementsProject/secp256k1-zkp/pull/23/files#diff-67e997bcfdac55191033d57a16d1408aR536

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+
+# ifdef __cplusplus
+extern "C" {
+# endif
+
+/** Opaque structure representing a large number of NUMS generators */
+typedef struct secp256k1_bulletproof_generators secp256k1_bulletproof_generators;
+
+/** Version number used in header of circuit and circuit-assignment binary files */
+#define SECP256K1_BULLETPROOF_CIRCUIT_VERSION	1
```
instagibbs: unused

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+
+/* Maximum depth of 31 lets us validate an aggregate of 2^25 64-bit proofs */
+#define SECP256K1_BULLETPROOF_MAX_DEPTH 31
+
+/* Size of a hypothetical 31-depth rangeproof, in bytes */
+#define SECP256K1_BULLETPROOF_MAX_PROOF (160 + 36*32 + 7)
+
+/* Maximum memory, in bytes, that may be allocated to store a circuit representation */
+#define SECP256K1_BULLETPROOF_MAX_CIRCUIT (1024 * 1024 * 1024)
```
instagibbs: include a hand-wavy reason for limit?

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ *  In:   blinding_gen: generator that blinding factors will be multiplied by (cannot be NULL)
+ *                   n: number of NUMS generators to produce
+ */
+SECP256K1_API secp256k1_bulletproof_generators *secp256k1_bulletproof_generators_create(
+    const secp256k1_context* ctx,
+    const secp256k1_generator *blinding_gen,
+    size_t n
+) SECP256K1_ARG_NONNULL(1);
+
+/** Destroys a list of NUMS generators
```
instagibbs: please make it stupid obvious it's freeing this heap memory

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+#define SECP256K1_BULLETPROOF_MAX_PROOF (160 + 36*32 + 7)
+
+/* Maximum memory, in bytes, that may be allocated to store a circuit representation */
+#define SECP256K1_BULLETPROOF_MAX_CIRCUIT (1024 * 1024 * 1024)
+
+/** Allocates and initializes a list of NUMS generators, along with precomputation data
+ *  Returns a list of generators, or NULL if allocation failed.
+ *  Args:          ctx: pointer to a context object (cannot be NULL)
+ *  In:   blinding_gen: generator that blinding factors will be multiplied by (cannot be NULL)
+ *                   n: number of NUMS generators to produce
```
instagibbs: Is there any realistic upper-limit as per above constants, or can we go all the way up `size_t` modulo memory?

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+SECP256K1_API void secp256k1_bulletproof_generators_destroy(
+    const secp256k1_context* ctx,
+    secp256k1_bulletproof_generators *gen
+) SECP256K1_ARG_NONNULL(1);
+
+/** Verifies a single bulletproof (aggregate) rangeproof
+ *  Returns: 1: rangeproof was valid
+ *           0: rangeproof was invalid, or out of memory
+ *  Args:       ctx: pointer to a context object initialized for verification (cannot be NULL)
+ *          scratch: scratch space with enough memory for verification (cannot be NULL)
```
instagibbs: as a user I need some guidance here...

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+    }
+    ecmult_data = (secp256k1_bulletproof_vfy_ecmult_context *)secp256k1_scratch_alloc(scratch, n_proofs * sizeof(*ecmult_data));
+    innp_ctx = (secp256k1_bulletproof_innerproduct_context *)secp256k1_scratch_alloc(scratch, n_proofs * sizeof(*innp_ctx));
+
+    /* In general you cannot memcmp secp256k1_ge's like this because their field
+     * elements may represent the same number differently. In this case it is ok
+     * because (a) a false positive here is no big deal, it will add one mult per
+     * proof to he giant ecmult_multi at the end but not change any semantics;
+     * and (b) typically this list of generators was deterministically decoded
+     * from a list of secp256k1_generators which have a compact encoding, so that
```
instagibbs: trailing whitespace

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ *             gens: generator set with at least 2*nbits*n_commits many generators (cannot be NULL)
+ *  In:       proof: array of byte-serialized rangeproofs (cannot be NULL)
+ *         n_proofs: number of proofs in the above array, and number of arrays in the `commit` array
+ *             plen: length of every individual proof
+ *        min_value: array of arrays of minimum values to prove ranges above, or NULL for all-zeroes
+ *           commit: array of arrays of pedersen commitment that the rangeproofs is over (cannot be NULL)
+ *        n_commits: number of commitments in each element of the above array (cannot be 0)
+ *            nbits: number of bits in each proof
```
instagibbs: >ARG_CHECK(nbits > 0);
>ARG_CHECK(nbits <= 64);

`(must be between 1 and 64)` or similar

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ *           0: rangeproof was invalid, or out of memory
+ *  Args:       ctx: pointer to a context object initialized for verification (cannot be NULL)
+ *          scratch: scratch space with enough memory for verification (cannot be NULL)
+ *             gens: generator set with at least 2*nbits*n_commits many generators (cannot be NULL)
+ *  In:       proof: byte-serialized rangeproof (cannot be NULL)
+ *             plen: length of the proof
+ *        min_value: array of minimum values to prove ranges above, or NULL for all-zeroes
+ *           commit: array of pedersen commitment that this rangeproof is over (cannot be NULL)
+ *        n_commits: number of commitments in the above array (cannot be 0)
+ *            nbits: number of bits proven for each range
```
instagibbs: >ARG_CHECK(nbits > 0);
>ARG_CHECK(nbits <= 64);

`(must be between 1 and 64)` or similar

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+    const secp256k1_generator* value_gen,
+    const unsigned char* extra_commit,
+    size_t extra_commit_len
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(7) SECP256K1_ARG_NONNULL(10);
+
+/** Batch-verifies multiple bulletproof (aggregate) rangeproofs of the same size using same generator
+ *  Returns: 1: all rangeproofs were valid
+ *           0: some rangeproof was invalid, or out of memory
+ *  Args:       ctx: pointer to a context object initialized for verification (cannot be NULL)
+ *          scratch: scratch space with enough memory for verification (cannot be NULL)
```
instagibbs: as a user I need some guidance here...

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ *             gens: generator set with at least 2*nbits*n_commits many generators (cannot be NULL)
+ *  In:       proof: array of byte-serialized rangeproofs (cannot be NULL)
+ *         n_proofs: number of proofs in the above array, and number of arrays in the `commit` array
+ *             plen: length of every individual proof
+ *        min_value: array of arrays of minimum values to prove ranges above, or NULL for all-zeroes
+ *           commit: array of arrays of pedersen commitment that the rangeproofs is over (cannot be NULL)
+ *        n_commits: number of commitments in each element of the above array (cannot be 0)
+ *            nbits: number of bits in each proof
+ *        value_gen: generator multiplied by value in pedersen commitments (cannot be NULL)
+ *     extra_commit: array of additonal data committed to by the rangeproof
```
instagibbs: (individual fields can only be NULL if the corresponding extra_commit_len` equals 0)

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ * extra_commit_len: array of lengths of additional data
```
instagibbs: indentation

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,182 @@
...
+ *  In/out:    plen: pointer to size of `proof`, to be replaced with actual length of proof (cannot be NULL)
+ *  In:       value: array of values committed by the Pedersen commitments (cannot be NULL)
+ *        min_value: array of minimum values to prove ranges above, or NULL for all-zeroes
+ *            blind: array of blinding factors of the Pedersen commitments (cannot be NULL)
+ *        n_commits: number of entries in the `value` and `blind` arrays
+ *        value_gen: generator multiplied by value in pedersen commitments (cannot be NULL)
+ *            nbits: number of bits proven for each range
+ *            nonce: random 32-byte seed used to derive blinding factors (cannot be NULL)
+ *     extra_commit: additonal data committed to by the rangeproof
+ * extra_commit_len: length of additional data
```
instagibbs: indentation

include/secp256k1_commitment.h
```diff
@@ -0,0 +1,165 @@
...
+ *            order (probability for random 32 byte number < 2^-127). Retry with
+ *            different values.
+ *
+ * In:                 ctx: pointer to a context object
+ *                   value: array of asset values, `v` in the above paragraph.
+ *                          May not be NULL unless `n_total` is 0.
+ *         generator_blind: array of asset blinding factors, `r` in the above paragraph
+ *                          May not be NULL unless `n_total` is 0.
+ *                 n_total: Total size of the above arrays
+ *                n_inputs: How many of the initial array elements represent commitments that
```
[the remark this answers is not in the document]
instagibbs: I'd rather not move arguments around unless there's a strong argument, as they're in use already.

include/secp256k1_generator.h
```diff
@@ -24,6 +24,12 @@ typedef struct {
     unsigned char data[33];
 } secp256k1_generator;
 
+/** Standard secp256k1 generator G */
+extern const secp256k1_generator secp256k1_generator_const_g;
+
+/** Alternate secp256k1 generator from Elements Alpha */
```
instagibbs: We really don't need to support Elements Alpha.

include/secp256k1_bulletproofs.h
apoelstra: This constant should just be dropped actually.

include/secp256k1_bulletproofs.h
apoelstra: Not really any upper limit. They take a while to generate, on the order of 1000/sec.

include/secp256k1_bulletproofs.h
apoelstra: What specifically about indentation?

include/secp256k1_generator.h
apoelstra: It's useful to have a standard second generator though for non-CA applications.

include/secp256k1_bulletproofs.h
instagibbs: gimme 3 more whitespaces so it matches?

apoelstra: Ok, sure, but then the `:`s don't line up

instagibbs: oh fine

include/secp256k1_bulletproofs.h
```diff
@@ -0,0 +1,176 @@
...
+ */
+SECP256K1_WARN_UNUSED_RESULT SECP256K1_API int secp256k1_bulletproof_rangeproof_rewind(
+    const secp256k1_context* ctx,
+    const secp256k1_bulletproof_generators* gens,
+    uint64_t* value,
+    unsigned char* blind,
+    const unsigned char* proof,
+    size_t plen,
+    size_t min_value,
```
yeastplume: Think that's slipped through again, should be `uint64_t min_value` (actually found this is causing compilation errors on some setups)

src/scalar.h
```diff
@@ -106,4 +106,7 @@ static void secp256k1_scalar_split_lambda(secp256k1_scalar *r1, secp256k1_scalar
 /** Multiply a and b (without taking the modulus!), divide by 2**shift, and round to the nearest integer. Shift must be at least 256. */
 static void secp256k1_scalar_mul_shift_var(secp256k1_scalar *r, const secp256k1_scalar *a, const secp256k1_scalar *b, unsigned int shift);
 
+/** Generate two scalars from a 32-byte seed and an integer using the chacha20 stream cipher */
+static void secp256k1_scalar_chacha20(secp256k1_scalar *r1, secp256k1_scalar *r2, const unsigned char *seed, uint64_t idx);
```
yeastplume: Discrepancy between declaration and impl here as well `idx`

include/secp256k1_bulletproofs.h
apoelstra: Fixed

src/scalar.h
apoelstra: Fixed. Only discrepancy was in the exhaustive tests impl I think

src/scalar_4x64_impl.h
real-or-random: This is still open.

yeastplume:
We haven't been able to get `secp_bulletproof_rangeproof_verify_multi` working thus far. In our testing, it seems to work fine when passing in a single proof and a single commit but as soon as we pass in multiples (which verify when passing in one-by-one,) it refuses to validate.

Also noticing that none of the tests for `verify_multi` actually specify a length of more than a single commit. Might be worth adding a test or case or two for this?

apoelstra:
@yeastplume The benchmarks test the API-exposed `verify_multi` with aggregates and batch validation simultaneously, if you want to see how to do this (but given that MW does not support aggregation, I'm not sure what value that'd have for you).

In the unit tests, the underlying `rangeproof_verify_impl` function is indeed tested with multiple commits. See the `test_bulletproof_rangeproof_aggregate` function.

src/scalar_4x64_impl.h
apoelstra: `idx` is now a `uint64_t`.

yeastplume:
Apologies, I hadn't looked into the benchmark tests. Thanks!

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+ * and prover cost.
+ *
+ * For the most part, all comments assume this value is at 4.
+ */
+#define IP_AB_SCALARS	4
+
+/* Bulletproof inner products consist of the four scalars and `2[log2(n) - 1]` points
+ * `a_1`, `a_2`, `b_1`, `b_2`, `L_i` and `R_i`, where `i` ranges from 0 to `log2(n)-1`.
+ *
+ * The prover takes as input a point `P` and scalar `c`. It proves that there exist
```
real-or-random: nit: `there exist` -> `it knows`?

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+ * scalars `a_i`, `b_i` for `i` ranging from 0 to `n-1`, such that
```
real-or-random: This is still somewhat confusing with the index 0 of `a` and `b`.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+ * For the remaining invocations it passes through to another callback, `rangeproof_cb_data` which
+ * computes `P`. The reason for this is that in practice `P` is usually defined by another multiexp
+ * rather than being a known point, and it is more efficient to compute one exponentiation.
+ *
+ * Inline we refer to the first `2n` coefficients as `s_i` for `i` from 0 to `2n-1`, since that
+ * is the more convenient indexing. In particular we describe (a) how the indices `j` and `k`,
+ * from the big comment block above, are chosen; and (b) when/how each `x_k^-2` is computed.
+ */
+static int secp256k1_bulletproof_innerproduct_vfy_ecmult_callback(secp256k1_scalar *sc, secp256k1_ge *pt, size_t idx, void *data) {
+    secp256k1_bulletproof_innerproduct_vfy_ecmult_context *ctx = (secp256k1_bulletproof_innerproduct_vfy_ecmult_context *) data;
```
real-or-random: I think it is guaranteed by the caller and nothing really bad happens but I think we should check that `ctx->vec_len` is not 0 here.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+ */
+#define IP_AB_SCALARS	4
+
+/* Bulletproof inner products consist of the four scalars and `2[log2(n) - 1]` points
+ * `a_1`, `a_2`, `b_1`, `b_2`, `L_i` and `R_i`, where `i` ranges from 0 to `log2(n)-1`.
+ *
+ * The prover takes as input a point `P` and scalar `c`. It proves that there exist
+ * scalars `a_i`, `b_i` for `i` ranging from 0 to `n-1`, such that
+ *     `P = sum_i [a_i G_i + b_i H_i]` and `<{a_i}, {b_i}> = c`.
+ * where `G_i` and `H_i` are standard NUMS generators.
```
real-or-random: nit: `.` -> `,`...

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+             * We now explain the index `j`: it is the largest number with one fewer 1-bits than `i`.
+             * Alternately, the most recently returned `s_j` where `j` has one fewer 1-bits than `i`.
+             *
+             * To ensure that `s_j` is available when we need it, on each iteration we define the
+             * variable `cache_idx` which simply counts the 1-bits in `i`; before returning `s_i`
+             * we store it in `ctx->proof[i].xcache[cache_idx]`. Then later, when we want "most
+             * recently returned `s_j` with one fewer 1-bits than `i`, it'll be sitting right
+             * there in `ctx->proof[i].xcache[cache_idx - 1]`.
+             *
+             * Note that `ctx->proof[i].xcache[0]` will always equal `-a_1 * prod_{i=1}^{n-1} x_i^-2`,
```
real-or-random: Could maybe be checked in DEBUG/TEST mode.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+        *sc = ctx->p_offs;
+        *pt = *ctx->g;
+    /* Remaining points are whatever the rangeproof wants */
+    } else if (ctx->shared_g && idx == 2 * (ctx->vec_len + ctx->lg_vec_len * ctx->n_proofs) + 1) {
+        /* Special case: the first extra point is independent of the proof, for both rangeproof and circuit */
+        size_t i;
+        secp256k1_scalar_clear(sc);
+        for (i = 0; i < ctx->n_proofs; i++) {
+            secp256k1_scalar term;
+            if ((ctx->proof[i].proof->rangeproof_cb)(&term, pt, &ctx->randomizer[i], 2 * (ctx->vec_len + ctx->lg_vec_len), ctx->proof[i].proof->rangeproof_cb_data) == 0) {
```
real-or-random: ` != 1` would be slightly more consistent with the rest and may be safer.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+        }
+    } else {
+        size_t proof_idx = 0;
+        size_t real_idx = idx - 2 * (ctx->vec_len + ctx->lg_vec_len * ctx->n_proofs) - 1 - !!ctx->shared_g;
+        while (real_idx >= ctx->proof[proof_idx].proof->n_extra_rangeproof_points - !!ctx->shared_g) {
+            real_idx -= ctx->proof[proof_idx].proof->n_extra_rangeproof_points - !!ctx->shared_g;
+            proof_idx++;
+            VERIFY_CHECK(proof_idx < ctx->n_proofs);
+        }
+        if ((ctx->proof[proof_idx].proof->rangeproof_cb)(sc, pt, &ctx->randomizer[proof_idx], 2 * (ctx->vec_len + ctx->lg_vec_len), ctx->proof[proof_idx].proof->rangeproof_cb_data) == 0) {
```
real-or-random: same here

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+        serproof += 32;
+
+        /* Extract a, b */
+        for (j = 0; j < n_ab; j++) {
+            secp256k1_scalar_set_b32(&ab[j], serproof, &overflow);
+            if (overflow) {
+                secp256k1_scratch_deallocate_frame(scratch);
+                return 0;
+            }
+            /* TODO our verifier currently bombs out with zeros because it uses
```
real-or-random: I don't understand this.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+
+        /* Compute (-a0 * r * x1 * ... * xn)^-1 which will be used to mask out individual x_i^-2's */
+        secp256k1_scalar_negate(&ecmult_data.proof[i].xsqinv_mask, &ecmult_data.proof[i].abinv[0]);
+        secp256k1_scalar_mul(&ecmult_data.proof[i].xsqinv_mask, &ecmult_data.proof[i].xsqinv_mask, &ecmult_data.proof[i].abinv[n_ab - 1]);
+
+        /* Compute each scalar times the previous' inverse, which is used to switch between a's and b's */
+        for (j = n_ab - 1; j > 0; j--) {
+            size_t prev_idx;
+            if (j == n_ab / 2) {
+                prev_idx = j - 1; /* we go from a_n to b_0  */
```
real-or-random: `a_{n-1}`?

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+    size_t pt_idx = 0;
+    secp256k1_scalar dot;
+    size_t half_n_ab = n < IP_AB_SCALARS / 2 ? n : IP_AB_SCALARS / 2;
+
+    if (*proof_len < secp256k1_bulletproof_innerproduct_proof_length(n)) {
+        return 0;
+    }
+    *proof_len = secp256k1_bulletproof_innerproduct_proof_length(n);
+
+    /* Special-case lengths 0 and 1 whose proofs are just expliict lists of scalars */
```
real-or-random: typo in explicit

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+        secp256k1_scalar x;
+        int overflow;
+        size_t j;
+        const size_t n_ab = 2 * vec_len < IP_AB_SCALARS ? 2 * vec_len : IP_AB_SCALARS;
+
+        total_n_points += 2 * ecmult_data.lg_vec_len + proof[i].n_extra_rangeproof_points - !!shared_g; /* -1 for shared G */
+
+        /* Extract dot product, will always be the first 32 bytes */
+        secp256k1_scalar_set_b32(&dot, serproof, &overflow);
+        if (overflow) {
```
real-or-random: maybe `EXPECT` (not sure if it helps though). Same is true for all the other overflow checks.

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+                secp256k1_ecmult_multi_var(ecmult_ctx, scratch, &rj, NULL, &secp256k1_bulletproof_innerproduct_pf_ecmult_callback_h, (void *) &pfdata, 2u << i);
+                pfdata.genh += 2u << i;
+                secp256k1_ge_set_gej(&genh[j], &rj);
+            }
+
+            secp256k1_scalar_sqr(&yinv2, yinv);
+            for (j = 0; j < i; j++) {
+                secp256k1_scalar_sqr(&yinv2, &yinv2);
+            }
+            if (!secp256k1_bulletproof_inner_product_real_prove_impl(ecmult_ctx, scratch, out_pt, pt_idx, g, geng, genh, a_arr, b_arr, &yinv2, ux, halfwidth, commit)) {
```
real-or-random: `!= 1`

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+    secp256k1_scalar_negate(&taux, &taux);
+    secp256k1_scalar_negate(&mu, &mu);
+
+    /* Encode rangeproof stuff */
+    secp256k1_scalar_get_b32(&proof[0], &taux);
+    secp256k1_scalar_get_b32(&proof[32], &mu);
+    secp256k1_bulletproof_serialize_points(&proof[64], out_pt, 4);
+
+    /* Mix this into the hash so the input to the inner product proof is fixed */
+    /* TODO is this necessary? revisit */
```
real-or-random: Hm, good question. It seems to be a good idea to me because it ensures that an inner product proof cannot be copied and reused in a different context. Maybe @bbuenz has more insights... The question is if it's necessary to hash the first part of the range proof into the challenge for the inner product proof.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+        secp256k1_scalar_set_b32(&ecmult_data[i].t, &proof[i][64 + 128 + 1], &overflow);
+        if (overflow || secp256k1_scalar_is_zero(&ecmult_data[i].t)) {
+            secp256k1_scratch_deallocate_frame(scratch);
+            return 0;
+        }
+
+        /* Verify inner product proof */
+        ecmult_data[i].a = age;
+        ecmult_data[i].s = sge;
+        ecmult_data[i].n = nbits * n_commits;
```
real-or-random: Is `n` actually used? Or do we use the fourth argument of `secp256k1_bulletproof_inner_product_verify_impl()` instead`

src/modules/bulletproofs/rangeproof_impl.h
bbuenz: Yes I think so. Otherwise the prover can choose the first IP challenge before he knows P. The way I have always done it is by hashing the last challenge into the next challenge.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+        secp256k1_sha256 sha256;
+        unsigned char commit[32] = {0};
+        unsigned char randomizer61[32] = {0};  /* randomizer for eq (61) so we can add it to eq (62) to save a separate multiexp */
+        secp256k1_scalar taux, mu;
+        secp256k1_ge age, sge;
+        int overflow;
+        size_t j;
+
+        /* Commit to all input data: min value, pedersen commit, asset generator, extra_commit */
+        if (min_value != NULL && min_value[i] != NULL) {
```
real-or-random: This is a branch that influences what will be hashed. This makes me nervous, because different inputs could possibly yield the same hash through different control flow decisions (and I think we don't want to reason about the consequences of such a collision). The obvious way to avoid that is to add prefixes to the hashed values (different constants and counters for the loops) to the hash. For the conditionals, another way is to always hash something. All of this makes the code slightly slower unfortunately...

 (Same issue in lines 202, 215, 229, 223).

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+        secp256k1_sha256_finalize(&sha256, commit);
+
+        secp256k1_sha256_initialize(&sha256);
+        secp256k1_sha256_write(&sha256, commit, 32);
+        secp256k1_sha256_finalize(&sha256, randomizer61);
+        secp256k1_scalar_set_b32(&ecmult_data[i].randomizer61, randomizer61, &overflow);
+        if (overflow || secp256k1_scalar_is_zero(&ecmult_data[i].randomizer61)) {
+            secp256k1_scratch_deallocate_frame(scratch);
+            return 0;
+        }
```
real-or-random: With all that hashing in this function, have you considered leaving the hash (context) open and add more bytes it to it whenever there is a new input value?

src/modules/bulletproofs/inner_product_impl.h
apoelstra: Agreed, at this point in the comment I should 1-index.

src/modules/bulletproofs/inner_product_impl.h
apoelstra: I don't think so, because we don't know `-a_1` explicitly.

src/modules/bulletproofs/inner_product_impl.h
apoelstra: If a_1, a_2, b_1, b_2 are 0, then later in the code where we compute `abinv`, we would try to take a scalar inverse of 0. Fixing this would require changing the whole algorithm, so instead we simply reject such proofs. (They happen with probability ~0 when doing anything zero-knowledge, since in that case every component of the input vectors are uniformly random.)

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+    secp256k1_bulletproof_innerproduct_context *innp_ctx;
+    int ret;
+    size_t i;
+    int same_generators = 1;
+
+    /* sanity-check input */
+    if (POPCOUNT(nbits) != 1 || nbits > MAX_NBITS) {
+        return 0;
+    }
+    if (plen < 64 + 128 + 1) {  /* inner product argument will do a more precise check */
```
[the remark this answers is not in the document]
apoelstra: Oops, actually +32 rather than +4. Good catch.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -0,0 +1,792 @@
...
+    /* get challenges y and z */
+    secp256k1_ge_set_gej(&out_pt[0], &aj);
+    secp256k1_ge_set_gej(&out_pt[1], &sj);
+
+    secp256k1_bulletproof_update_commit(commit, &out_pt[0], &out_pt[1]);
+    secp256k1_scalar_set_b32(&y, commit, &overflow);
+    if (overflow || secp256k1_scalar_is_zero(&y)) {
+        return 0;
+    }
+    secp256k1_bulletproof_update_commit(commit, &out_pt[0], &out_pt[1]); /* TODO rehashing A and S to get a second challenge is overkill */
```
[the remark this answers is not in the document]
apoelstra: The issue is that I have this nice function that hashes two points, and no nice function that'd hash a constant :)

src/modules/bulletproofs/inner_product_impl.h
```diff
@@ -0,0 +1,848 @@
...
+
+    /* Protocol 2: hash dot product to obtain G-randomizer */
+    secp256k1_sha256_initialize(&sha256);
+    secp256k1_sha256_write(&sha256, commit_inp, 32);
+    secp256k1_sha256_write(&sha256, proof, 32);
+    secp256k1_sha256_finalize(&sha256, commit);
+
+    proof += 32;
+
+    secp256k1_scalar_set_b32(&ux, commit, &overflow);
```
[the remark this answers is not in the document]
apoelstra: Yes

src/modules/bulletproofs/rangeproof_impl.h
apoelstra: Yeah, it's used in `secp256k1_bulletproof_rangeproof_vfy_callback` in this file. The inner product code doesn't see this but the rangeproof stuff does.

src/modules/bulletproofs/rangeproof_impl.h
apoelstra: Yeah, my benchmarks show that there's very little speedup to be gained by making the hashes more efficient, so I think it wouldn't be worth the extra complexity and difficulty of reimplementing in a language that didn't have open hash contexts available.

apoelstra: That is, if I were maintaining a running sha2 context, somebody later trying to make a python verifier using `hashlib` would have a hard time.

src/modules/bulletproofs/inner_product_impl.h
apoelstra: It's fine if it is. In fact the inner product unit tests check this case.

apoelstra:
Addressed all comments except for the concern about hashing variable-length stuff into the batch-verification randomizer. Will fix.

src/modules/bulletproofs/rangeproof_impl.h
sipa: FWIW, hashlib has cloneable sha256 state objects :)

```python
>>> import hashlib
>>> x = hashlib.sha256(b"ha")
>>> y = x.copy()
>>> x.update(b"llo")
>>> y.update(b"ll")
>>> y.update(b"o")
>>> x.hexdigest()
'd3751d33f9cd5049c4af2b462735457e4d3baf130bcbb87f389e349fbaeb20b9'
>>> y.hexdigest()
'd3751d33f9cd5049c4af2b462735457e4d3baf130bcbb87f389e349fbaeb20b9'
```

apoelstra:
@real-or-random I think I've fixed the variable hashing. Added a commit which does so.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -209,31 +209,45 @@ static int secp256k1_bulletproof_rangeproof_verify_impl(const secp256k1_ecmult_c
...
+        secp256k1_sha256_write(&sha256, commit, 32);
         for (j = 0; j < n_commits; j++) {
-            secp256k1_bulletproof_update_commit(commit, &commitp[i][j], &value_gen[i]);
+            unsigned char vbuf[8];
+            uint64_t mv;
+            if (min_value != NULL && min_value[i] != NULL) {
+                mv = min_value[i][j];
+            } else {
+                mv = 0;
+            }
```
real-or-random: Do we accept arbitrary min values on the chain? If yes, we should ensure that there is no malleability caused by `min_value[i] == NULL` leading to the same hash as `min_value[i][j] == 0`.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -209,31 +209,45 @@ static int secp256k1_bulletproof_rangeproof_verify_impl(const secp256k1_ecmult_c
...
         if (extra_commit != NULL && extra_commit[i] != NULL) {
-            secp256k1_sha256_initialize(&sha256);
-            secp256k1_sha256_write(&sha256, commit, 32);
+            unsigned char len_buf[4];
+            len_buf[0] = extra_commit_len[i];
+            len_buf[1] = extra_commit_len[i] >> 8;
+            len_buf[2] = extra_commit_len[i] >> 16;
+            len_buf[3] = extra_commit_len[i] >> 24;
+
+            secp256k1_sha256_write(&sha256, len_buf, 4);
```
real-or-random: The length of this buffer is (was?) not my concern. The thing is that there's ambiguity between the cases 
 * `n_commits` = n and an extra commitment (only with length 4 bytes after your commit)
 * `n_commits` = n+1  and no extra commitment

for some n.

I think the simplest way to address this is to write the value of `n_commits` to the hash before the loop. Then we don't need to write the length of the extra commitment to the hash, it's just all the data after the last commitment. (Again, I have no idea of the consensus rules but we need to make sure that a "NULL" extra commitment is not malleable to an extra commitment with an empty string, similar to my comment about the min value above).

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -209,31 +209,45 @@ static int secp256k1_bulletproof_rangeproof_verify_impl(const secp256k1_ecmult_c
         size_t j;
 
         /* Commit to all input data: min value, pedersen commit, asset generator, extra_commit */
```
real-or-random: Is `age` below the asset generator? Just double-checking that the asset generator has not been forgotten.

src/modules/commitment/main_impl.h
```diff
@@ -108,22 +108,22 @@ int secp256k1_pedersen_blind_sum(const secp256k1_context* ctx, unsigned char *bl
...
     VERIFY_CHECK(ctx != NULL);
-    ARG_CHECK(!pcnt || (commits != NULL));
-    ARG_CHECK(!ncnt || (ncommits != NULL));
+    ARG_CHECK(!n_pos || (pos != NULL));
+    ARG_CHECK(!n_neg || (neg != NULL));
     (void) ctx;
     secp256k1_gej_set_infinity(&accj);
-    for (i = 0; i < ncnt; i++) {
-        secp256k1_pedersen_commitment_load(&add, ncommits[i]);
+    for (i = 0; i < n_pos; i++) {
```
apoelstra: `n_neg` and `n_pos` were inadvertently swapped here. Thanks @yeastplume for catching this.

src/modules/bulletproofs/rangeproof_impl.h
apoelstra: On the chain the min value is always 0 or 1 as determined by consensus rules. (Specifically it is 0 for provably unspendable outputs, to allow mimblewimble kernels, and 1 otherwise.)

src/modules/bulletproofs/rangeproof_impl.h
apoelstra: In the consensus rules the `extra_commitment` will always be the scriptpubkey, or something like that ... once again it is determined by consensus and not ever serialized as part of the proof.

src/modules/bulletproofs/rangeproof_impl.h
apoelstra: The asset generator is `value_gen` and it is actually committed, so this is OK.

apoelstra:
@real-or-random updated most of the commits with a couple minor things (specifically I added portable functions for POPCOUNT and CTZL; and fixed yeastplume's bug in the commitment module where I swapped `n_pos` and `n_neg`).

The last commit has the real changes -- I replaced my commitment fix with a new one that simply hashes `n_commit` into the hash.

@yeastplume FYI this commit will hardfork grin, which kinda sucks because for you guys `n_commit` is always a fixed 1 (so it's implicitly committed).

real-or-random:
@apoelstra I only looked at the last commit, nit there: The commit message says that also `age` is added to the hash but the commit does not actually do that. (I think it's actually in the code but introduced by a different commit.) 

Looks good otherwise. :beers:

apoelstra:
Fixed commit message.

Gonna set up travis on `ElementsProject` to try to get some testing on uncommon configurations before merging.

apoelstra:
valgrind is complaining about use of uninitialized memory. I think my `memcmp`ing generators to try and use the "every generator is the same" optimization is to blame. Investigating.

apoelstra:
Added a commit to fix this. @real-or-random can you sanity-check that, then I'll squash it into the history and merge.

apoelstra:
Oops, had it backwards (confirmed by checking benchmark numbers, which unfortunately is the only way to notice if this check fails). Fixed.

src/modules/bulletproofs/rangeproof_impl.h
```diff
@@ -186,15 +186,12 @@ static int secp256k1_bulletproof_rangeproof_verify_impl(const secp256k1_ecmult_c
...
-     * because (a) a false positive here is no big deal, it will add one mult per
-     * proof to he giant ecmult_multi at the end but not change any semantics;
-     * and (b) typically this list of generators was deterministically decoded
-     * from a list of secp256k1_generators which have a compact encoding, so that 
-     * equal group elements actually will compare equal. */
+    /* Check if all generators are equal. If so, we can amortize all their scalar multiplications
+     * by them and save one scalar-ge multiplication per proof. */
     for (i = 1; i < n_proofs; i++) {
-        if (memcmp(&value_gen[i], &value_gen[i - 1], sizeof(value_gen[i])) != 0) {
+        VERIFY_CHECK(!value_gen[i].infinity);
```
real-or-random: `secp256k1_ge_is_infinity()`

real-or-random: Should also check for `i = 0`

src/modules/bulletproofs/util.h
```diff
@@ -0,0 +1,146 @@
...
+    }
+}
+
+SECP256K1_INLINE static void secp256k1_bulletproof_deserialize_point(secp256k1_ge *pt, const unsigned char *data, size_t i, size_t n) {
+    const size_t bitveclen = (n + 7) / 8;
+    const size_t offset = bitveclen + i*32;
+    secp256k1_fe fe;
+
+    secp256k1_fe_set_b32(&fe, &data[offset]);
+    secp256k1_ge_set_xquad(pt, &fe);
```
real-or-random: check return value?

apoelstra: Oof, yep, and I need to potentially abort everywhere this function is called.

apoelstra:
Added the extra `VERIFY_CHECK` to the memcmp commit; added a commit which checks the set_xquad return value.

real-or-random:
Yes, looks good. Two points that you can ignore.
 - It's really a nit but calling `secp256k1_ge_is_infinity()` is nicer I think
 - We could actually enforce that the return value of the deserialize functions is not ignored.

I haven't looked in particular for other places where we could hit invalid points. Do we parse points somewhere else too?

apoelstra:
* Check point deserialization succeeds in the inner product verifier
* Change `infinity` flag checks to use `ge_is_infinity`
* Squash/rebase everything
* ran `git rebase -x 'make && ./tests 1'` to check that every commit compiles

@yeastplume the deserialization checks need to be pulled into grin because they potentially allow invalid proofs to be accepted (though only invalid proofs that were created by a legitimate coin holder, so it's not *critical*). The commit labelled "bulletproofs: add `n_commits` to the hash of all rangeproof input data" will hardfork you, you might want to revert that one since it's not important for a system that doesn't aggregate rangeproofs.

@real-or-random I think that covers everything, and we're good to merge :) Except that I need to reduce the test iterations or something to prevent Travis timing out.

yeastplume:
@apoelstra thanks for the update. We're actually in the process of preparing a fresh 4th testnet (with all sorts of other consensus-breaking changes,) so the plan is just to rebase from upstream as soon as this is merged and deploy on T4. 

You're probably aware we have a few additions to the message hidden in alpha and some changes to the API to enable multi-party proofs, perhaps these might be considered for future enhancements.

real-or-random:
> @real-or-random I think that covers everything, and we're good to merge :) Except that I need to reduce the test iterations or something to prevent Travis timing out.

Yep, ACK. As far as I understand, Travis complains that it doesn't see any output for 10min, so printing more progress information should actually suffice.

apoelstra:
Hmm, after #30 our API does the point decompression before giving the user a `secp256k1_pedersen_commitment` or a `secp256k1_generator`, which kills the "fast rewinding" performance. I think I need to change the API of that function to take a byte-serialized generator and commitment, since that's what would be available on the blockchain.

apoelstra:
Rebased. In the interest of having a straightforward rebase without new code, I didn't change the API of _rewind. @yeastplume this changes the "failed rewind" time from 2.8 µs to 9.9 µs on my laptop, and I'm not going to prioritize fixing it for a while, sorry. The old performance can be recovered without requiring any forks or anything, so it shouldn't be a big deal for a while anyway.

I will add another commit which reduces the test load so Travis doesn't time out (getting output from `make check` is tricky), then we should be good to merge.

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    const secp256k1_ge *commitp_ptr;
+    const uint64_t *minvalue_ptr;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(scratch != NULL);
+    ARG_CHECK(gens != NULL);
+    ARG_CHECK(gens->n >= 2 * nbits * n_commits);
+    ARG_CHECK(proof != NULL);
+    ARG_CHECK(commit != NULL);
+    ARG_CHECK(n_commits > 0);
```
real-or-random: We should additionally have a upper bound for `n_commits` because it's used to calculate the size of many allocs and the number of points in the multi-exponentiation. And maybe a comment that explains that the bound is there for a good reason.

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    ARG_CHECK(nbits > 0);
+    ARG_CHECK(nbits <= 64);
+    ARG_CHECK(value_gen != NULL);
+    ARG_CHECK(extra_commit != NULL || extra_commit_len == 0);
+    ARG_CHECK(secp256k1_ecmult_context_is_built(&ctx->ecmult_ctx));
+
+    if (!secp256k1_scratch_allocate_frame(scratch, 2 * n_commits * sizeof(secp256k1_ge), 1)) {
```
real-or-random: For example here.

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    secp256k1_ge *value_genp;
+    size_t i;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(scratch != NULL);
+    ARG_CHECK(gens != NULL);
+    ARG_CHECK(gens->n >= 2 * nbits * n_commits);
+    ARG_CHECK(commit != NULL);
+    ARG_CHECK(proof != NULL);
+    ARG_CHECK(n_proofs > 0);
```
real-or-random: Same is true for `n_proofs`

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    ARG_CHECK(scratch != NULL);
+    ARG_CHECK(gens != NULL);
+    ARG_CHECK(gens->n >= 2 * nbits * n_commits);
+    ARG_CHECK(proof != NULL);
+    ARG_CHECK(plen != NULL);
+    ARG_CHECK(value != NULL);
+    ARG_CHECK(blind != NULL);
+    ARG_CHECK(value_gen != NULL);
+    ARG_CHECK(nonce != NULL);
+    ARG_CHECK(n_commits > 0 && n_commits);
```
real-or-random: nit:
```suggestion
    ARG_CHECK(n_commits > 0);
```

src/modules/commitment/main_impl.h
```diff
@@ -0,0 +1,217 @@
...
+    secp256k1_pedersen_commitment_load(&ge, commit);
+
+    output[0] = 9 ^ secp256k1_fe_is_quad_var(&ge.y);
+    secp256k1_fe_normalize_var(&ge.x);
+    secp256k1_fe_get_b32(&output[1], &ge.x);
+    return 1;
+}
+
+/* Generates a pedersen commitment: *commit = blind * G + value * G2. The blinding factor is 32 bytes.*/
+int secp256k1_pedersen_commit(const secp256k1_context* ctx, secp256k1_pedersen_commitment *commit, const unsigned char *blind, uint64_t value, const secp256k1_generator* value_gen, const secp256k1_generator* blind_gen) {
```
garyyu: `const secp256k1_context* ctx` is never used in this API. Shall we remove this parameter?

src/modules/commitment/tests_impl.h
```diff
@@ -0,0 +1,230 @@
...
+    secp256k1_context_set_error_callback(sign, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_error_callback(vrfy, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_error_callback(both, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(none, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(sign, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(vrfy, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(both, counting_illegal_callback_fn, &ecount);
+
+    secp256k1_rand256(blind);
+    CHECK(secp256k1_pedersen_commit(none, &commit, blind, val, &secp256k1_generator_const_h, &secp256k1_generator_const_g) == 0);
```
garyyu: these tests will fail, because `secp256k1_pedersen_commit` never use `ctx` parameter,  so whatever we use (`none`, `vrfy`, or `sign`) the result will be same.

src/tests.c
```diff
@@ -5070,6 +5199,27 @@ int main(int argc, char **argv) {
     run_recovery_tests();
 #endif
 
+#ifdef ENABLE_MODULE_GENERATOR
+    run_generator_tests();
+#endif
+
```
garyyu: Add test for pedersen commitment:
```
#ifdef ENABLE_MODULE_COMMITMENT
    run_commitment_tests();
#endif

```

apoelstra: Lol, shit, thanks

apoelstra:
Rebased on #35

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    ARG_CHECK(value_gen != NULL);
+    ARG_CHECK((extra_commit_len == NULL) == (extra_commit == NULL));
+    if (extra_commit != NULL) {
+        for (i = 0; i < n_proofs; i++) {
+            ARG_CHECK(extra_commit[i] != NULL || extra_commit_len[i] == 0);
+        }
+    }
+    ARG_CHECK(secp256k1_ecmult_context_is_built(&ctx->ecmult_ctx));
+
+    if (!secp256k1_scratch_allocate_frame(scratch, n_proofs * (sizeof(*value_genp) + sizeof(*commitp) + n_commits * sizeof(**commitp)), 1 + n_proofs)) {
```
garyyu: According to @jasonxh 's finding in https://github.com/mimblewimble/secp256k1-zkp/pull/38,  here the `1 + n_proofs)` should be `2 + n_proofs)`, since we have `secp256k1_scratch_alloc` at line 148, 149 and `n_proofs` loops at line 150.

apoelstra: Great catch, thanks!

apoelstra: Fixed.

src/modules/commitment/main_impl.h
apoelstra: For reasons of API consistency we prefer to have the context object in every public-facing function.

real-or-random:
As pointed out by drexl in IRC, we should make sure to read https://grin-tech.org/audits/jpa-audit-report.html and  https://github.com/mimblewimble/secp256k1-zkp/pull/37

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,237 @@
...
+    }
+    commitp = (secp256k1_ge *)secp256k1_scratch_alloc(scratch, n_commits * sizeof(*commitp));
+    blinds = (secp256k1_scalar *)secp256k1_scratch_alloc(scratch, n_commits * sizeof(*blinds));
+
+    secp256k1_generator_load(&value_genp, value_gen);
+    for (i = 0; i < n_commits; i++) {
+        int overflow;
+        secp256k1_gej commitj;
+        secp256k1_scalar_set_b32(&blinds[i], blind[i], &overflow);
+        if (overflow || secp256k1_scalar_is_zero(&blinds[i])) {
```
real-or-random:
```suggestion
        if (overflow || secp256k1_scalar_is_zero(&blinds[i])) {
            secp256k1_scratch_deallocate_frame(scratch);
```

src/modules/bulletproofs/main_impl.h
real-or-random: just a nit but not resolved I guess (just stumbled over it again)

real-or-random:
cc @veorq, just for reference

src/modules/commitment/tests_impl.h
```diff
@@ -0,0 +1,230 @@
...
+    for (i = 0; i < n_inputs - 1; i++) {
+        value[i] = secp256k1_rands64(0, total_value);
+        total_value -= value[i];
+    }
+    value[i] = total_value;
+
+    /* Correct for blinding factors and do the commitments */
+    CHECK(secp256k1_pedersen_blind_generator_blind_sum(ctx, value, (const unsigned char * const *) generator_blind, pedersen_blind, n_generators, n_inputs));
+    for (i = 0; i < n_generators; i++) {
+        CHECK(secp256k1_pedersen_commit(ctx, &commit[i], pedersen_blind[i], value[i], &generator[i], &secp256k1_generator_const_h));
```
garyyu: The last parameter should be `secp256k1_generator_const_g` instead of `secp256k1_generator_const_h`.

Related PR in Grin: https://github.com/mimblewimble/secp256k1-zkp/pull/44

dr-orlovsky:
@apoelstra any chances of having this rebased and getting the great work and effort be finally merged?

apoelstra:
@dr-orlovsky yes, very good chance in the next week or two :)

apoelstra:
Closing this. Will retry in a more staged fashion.

apoelstra:
Closing in favor of #108. Note that the new proofs are not remotely compatible with the old ones.

PR #30, opened 2018-10-01
check the value of `ge_set_xquad` everywhere in -zkp modules

apoelstra:


apoelstra:
cc @real-or-random @jonasnick

real-or-random:
utACK.

This is rather for a second PR:

Currently, the parse functions promise to succeed if the bytestring contains a "valid" generator / commitment. That's not quite right. It's kind of safe, because the user is forced to use the generators and commitments through our API functions, which check that the point is actually valid. I say "kind of" because our API promises that parsed objects can be memcmp'd, so the user could memcmp invalid points.

I think it will be less surprising, less error-prone, and more efficient to check full validity at parse time. This will also be consistent with the pubkey_parse function, which checks everything and then the pubkey_load merely does a memcpy. Probably we should just steal from there; the only difference is that it accepts uncompressed encoding and uses a different prefix.
If you agree, I can create a PR for that (or you do it here).
If not, we should clarify in the documentation what "valid" means.

apoelstra:
I don't understand. What further validity checks are needed?

I see that we should check the `x' coordinate for overflow, which we currently don't, because that's a malleability vector, but it's fairly benign.

Also we should not be telling users they can `memcmp` generators that have been parsed.

real-or-random:
Sorry, maybe my comment was confusing. I'm not saying that more checks are needed. (Except that you're right, we should check for x-overflow.)

I'm proposing to move the checks from the `generator_load` functions to the `generator_parse` functions (and the same for commitments) Advantages are
 * We fail earlier, so there is less room for internal functions and users to work with invalid points, e.g., memcmp them as we explain here: https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/include/secp256k1_generator.h#L20
 * We perform the checks only once instead of every time we load a point.
 * We are consistent with the corresponding functions for pubkeys. Here, the `_parse` function performs all the validation and does not postpone it to the `_load` function: https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/eckey_impl.h#L17

apoelstra:
Ah! Yes, you're right.

apoelstra:
@real-or-random  I completely rewrote the PR.

real-or-random:
utACK

src/modules/rangeproof/main_impl.h
```diff
@@ -43,22 +44,38 @@ static void secp256k1_pedersen_commitment_save(secp256k1_pedersen_commitment* co
...
+    secp256k1_ge ge;
+
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(output != NULL);
     ARG_CHECK(commit != NULL);
-    memcpy(output, commit->data, sizeof(commit->data));
+
+    secp256k1_pedersen_commitment_load(&ge, commit);
+
+    output[0] = 11 ^ secp256k1_fe_is_quad_var(&ge.y);
```
jonasnick: Shouldn't that be `output[0] = 8 ^ secp256k1_fe_is_quad_var(&ge.y);` ?

jonasnick: and if so we need a test for this :)

apoelstra: Yes, and yes :)

PR #31, opened 2018-10-08
Fix pedersen commitment serialization after #30 broke it

apoelstra:


instagibbs:
looks better, will test

instagibbs:
ACK

PR #33, opened 2018-11-19
Add points_module with general interface for operations with points (Resolve #32)

### LINKED ISSUE #32: General interface for curve points
Libsecp256k1 provides an interface for operations on curve secp256k1. However, following the principle "Be difficult to use insecurely", only a specific API which directly describes operands was exposed. For instance, instead of functions for sum and multiply by scalars on curve points, only specific functions for sum and multiplications for public keys were made available.

However, during development the number of objects which inherently are points on curve has been growing and presently there are public keys, pedersen commitments and generators.

Our application (leer) requires (de)serialization for curve points independent of the nature of those points (we are developing a virtual machine where stack may contain curve points and has native opcodes for curve operations). Besides that, such functionality is required by libsecp256k1 modules by themselves: rangeproof module uses its own point serialization format https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/rangeproof/rangeproof_impl.h#L54 .

Hence it looks expedient to provide API for point (de)serialization as well as for bidirectional casting between points and other objects with specific purposes (that are inherently points) .

### Chosing of header byte
Currently there are these rules for headers byte.
Header bytes `0b00000000`,`0b00000001`,`0b00000110` and `0b00000111` are reserved for special form of public keys for historical compatibility with bitcoin code.
For other object:
 * Lowest bit always encodes whether y-coordinate of point is odd or even.
 * `0b0000001x` header is used for public keys
 * `0b0000101x` header is used for generators
 * `0b0000100x` header is used for pedersen commitments

Since point object is a uniting type, and there will probably be other specific objects which represent points in the future, it is pertinent to use a distinctive header, such as `0b1000000x` or `0b1111111x`.

WTRMQDev:


apoelstra:
Thanks for your work but we don't want to go this direction. Many of our "points" are quite different objects - some use different ways to distinguish between even and odd points, some are always 32 bytes, etc.

PR #34, opened 2018-11-19
Add explanation about how BIP32 unhardened derivation can be used to …

jonasnick:
…simplify whitelisting

### DIFF
--- a/src/modules/whitelist/whitelist.md
+++ b/src/modules/whitelist/whitelist.md
@@ -93,4 +93,13 @@ the remaining public keys are verified out-of-band when setting up the system,
 so there is no direct benefit to this. We do it only to reduce fragility and
 increase safety of unforeseen uses.
 
-
+Having to access the offline key `Q_i` to compute the secret to the sum `W +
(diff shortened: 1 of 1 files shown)


src/modules/whitelist/whitelist.md
```diff
@@ -93,4 +93,12 @@ the remaining public keys are verified out-of-band when setting up the system,
 so there is no direct benefit to this. We do it only to reduce fragility and
 increase safety of unforeseen uses.
 
-
+Having to access the offline key `Q_i` to compute the secret to the sum `W +
+Q_i` for every authorization is onerous.  Instead, if the whitelisted keys are
+created using
+[BIP32](https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki)
+unhardened derivation, the sum can be computed on an online machine. In order
+to achieve that, the offline key `Q_j` is set to the negated BIP32 master
```
instagibbs: Not the master key, but the last hardened derivation parent key(typically, the pubkey corresponding to the `xpub`?

jonasnick: thanks, fixed

PR #35, opened 2018-12-22
Add MuSig module

jonasnick:
This PR adds a module that implements a Schnorr-based multi-signature scheme called [MuSig](https://eprint.iacr.org/2018/068.pdf).
It is based on an original implementation by @apoelstra and builds on the schnorrsig module (https://github.com/bitcoin-core/secp256k1/pull/558) and trivial `ecmult_multi` (https://github.com/bitcoin-core/secp256k1/pull/580).

~~The PR is WIP because it needs more state machine tests (there are quite a few lines that are currently not covered).~~


include/secp256k1_musig.h
```diff
@@ -0,0 +1,417 @@
...
+ *
+ * 1. This structure is initialized with `secp256k1_musig_pubkey_combine`, which sets
+ *    the `pubkey` and `index` fields, and zeros out all other fields.
+ *
+ * 2. If the signer data is used in a non-public session, it should be updated with
+ *    `secp256k1_musig_get_public_nonce` which in addition to returning the nonce
+ *    takes an array of signer data structs and an array of commitments, and sets
+ *    each signer data struct's `nonce_commitment` field to the corresponding
+ *    commitment. If the signer data is used in a public session, the
+ *    `nonce_commitment` is set set in musig_session_initialize_public.
```
apoelstra: I think this comment should be rearranged so that it says, in order,

1. In a public session, use `musig_session_initialize_public` to initialize the `nonce_commitment` field.
2. In a non-public session (signing session? participatory session?) the function `secp256k1_musig_get_public_nonce` should also be used, which as a side-effect also returns the signer's public nonce. This ensures that the public nonce is not exposed until all commitments have been received.

include/secp256k1_musig.h
```diff
@@ -0,0 +1,417 @@
...
+
+/** Computes a combined public key, hash of keys for use with MuSig, and initializes
+ *  an array of signer data structures
+ *
+ *  Returns: 1 when the public keys were successfully combined, 0 otherwise
+ *  Args:        ctx: pointer to a context object initialized for verification
+ *                    (cannot be NULL)
+ *           scratch: scratch space used to compute the combined pubkey by
+ *                    multiexponentiation. If NULL, an inefficient algorithm is used.
+ *  Out: combined_pk: the MuSig-combined public key (cannot be NULL)
```
apoelstra: What do you think about allowing this to be NULL, in case the public key has already been computed (e.g. in a previous signing session)? In that case we would avoid doing the expensive ecmult_multi.

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,566 @@
...
+        secp256k1_ec_pubkey_serialize(ctx, commit, &commit_size, &session->nonce, SECP256K1_EC_COMPRESSED);
+        secp256k1_sha256_write(&sha, commit, commit_size);
+        secp256k1_sha256_finalize(&sha, nonce_commitment32);
+    }
+
+    return 1;
+}
+
+
+int secp256k1_musig_session_get_public_nonce(const secp256k1_context* ctx, secp256k1_musig_session *session, secp256k1_pubkey *nonce, secp256k1_musig_signer_data *signers, const unsigned char *const *commitments, size_t n_commitments) {
```
apoelstra: Interesting that all the failure modes here come from API misuse. I wonder if it's reasonable to turn some of these `return 0`s into `ARG_CHECK`s?

My feeling is no, I prefer to use `ARG_CHECK` only for things that could be prevented with Rust's type system, but I could go either way.

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,566 @@
...
+    ARG_CHECK(partial_sig != NULL);
+    ARG_CHECK(session != NULL);
+
+    if (!session->nonce_is_set || !session->has_secret_data) {
+        return 0;
+    }
+
+    secp256k1_scalar_set_b32(&sk, session->seckey, &overflow);
+    if (overflow) {
+        return 0;
```
apoelstra: We should probably zero out `sk` (and `k` below) in this and subsequent early-returns. Or even maintain a `ret` variable so there is only a single return.

apoelstra:
ACK except

* nits
* missing QA/tests
* changes to adaptor signature equation needed to match the Malavolta et al paper

I should also mention that the use of large non-opaque structs may be something we regret; but on the other hand, it is possible for most users to use them as though they were opaque, and IMHO the effort required to opaque-ify them (e.g. replacing the whole struct with a giant `unsigned char` array and then de/serializing objects throughout the entire rest of the code) would not be worth the resulting obfuscation.

include/secp256k1_musig.h
jonasnick: I feel like it's a bit weird that we describe a "workflow of a data structure" which kind of explains how to use the module but not entirely. How to use the module isn't explained anywhere else at all, maybe we should do that instead.

include/secp256k1_musig.h
jonasnick: Hm, right, if you want to initialize new signer_data but know the pubkey already you still have to call `secp256k1_musig_pubkey_combine` - that's confusing. I'm thinking about renaming that or somehow integrating it with `session_initialize`.

src/modules/musig/main_impl.h
jonasnick: Why would it be better to have an `ARG_CHECK` exactly for that?
In general for me it's a bit difficult to imagine what exactly rusts type system can prevent, but it won't be a big problem switching from `return 0` to `ARG_CHECK` once we have the rust wrapper.

jonasnick:
> changes to adaptor signature equation needed to match the Malavolta et al paper

the adaptor signature implementation should already match the paper

include/secp256k1_musig.h
apoelstra: Agreed, it'd be good to write a `.md` with some instructions. I'm working on a threshold signature scheme that reuses as much of this module as possible, and it's weird to find myself referring back to data structures' comments to remember how everything fits together.

src/modules/musig/main_impl.h
jonasnick: I didn't manage to get a readable implementation with a `ret` variable and didn't find a use of goto fail; in libsecp, but I added a commit that should best-effort-clear all secret scalars.

include/secp256k1_musig.h
apoelstra: For threshold signatures we need a new version of `session_initialize` but not a new version of `pubkey_combine`, so I like the current separation.

Maybe we should just tell users who want to avoid repeated computations that they have to store both the combined pubkey and the freshly-initialized `signer_data` structures?

The workflow I'm considering for threshold signatures is
1. Users call `secp256k1_musig_pubkey_combine` to compute the total MuSig key and initialize `signer_data` structures with each users' original public keys.
2. Each user calls `secp256k1_thresholdsig_keysplit` to split their key into shards, which they send to each user (and also broadcast "public coefficients" which act as a witness that the sharding was done correctly).
3. Recipients of the shards call `secp256k1_thresholdsig_verify_shard` which (a) verifies the public coefficients are consistent with the shards; (b) verifies that the public coefficients are consistent with the original public keys; (c) **updates in place** the `signer_data` structures since after the key-splitting ceremony everybody winds up with modified keys.

So both steps 1 and 3 are kinda expensive and are things that users might want to cache. It would be simple/consistent to tell them "store the `signer_data` structures", though I'm also a bit wary of inviting users to de/serialize this structure because there might be replay attacks.

jonasnick: Actually I think it's just super weird to initialize the signers in combined_pk, I hadn't realized that. It should happen in session_init functions.

apoelstra: All the steps I just described happen at key-generation time BTW, and in principle only need to happen once.

src/modules/musig/main_impl.h
apoelstra: My thinking is that users can assume `ARG_CHECK` failures never happen unless they're misusing the API, while "normal" failures are something they should check for.

include/secp256k1_musig.h
jonasnick: I rephrased the comment according to your suggestion.

include/secp256k1_musig.h
jonasnick: I moved the signer initialization from pubkey_combine to session_initialize.

jonasnick:
I added state machine tests and overflow tests. Now the tests cover the musig module as much as possible (https://htmlpreview.github.io/?https://raw.githubusercontent.com/jonasnick/secp256k1-zkp/24049468577e3821035b3e71d56b2201123096f0/coverage.src_modules_musig_main_impl.h.html).

I think the only thing that's missing right now is documentation for using the module.

jonasnick:
The MuSig code should now have the invariant that for any session identified by a unique session id, either `compute_messagehash` returns an error or the result is constant.

src/modules/musig/main_impl.h
apoelstra: Works for me.

include/secp256k1_musig.h
```diff
@@ -71,15 +71,15 @@ typedef struct {
...
+ *    signers' nonce_commitments.
  *
- * 2. If the signer data is used in a non-public session, it should be updated with
- *    `secp256k1_musig_get_public_nonce` which in addition to returning the nonce
- *    takes an array of signer data structs and an array of commitments, and sets
- *    each signer data struct's `nonce_commitment` field to the corresponding
- *    commitment. If the signer data is used in a public session, the
- *    `nonce_commitment` is set set in musig_session_initialize_public.
+ * 2. In a non-public session the nonce_commitments are set with the function
+ *    secp256k1_musig_get_public_nonce, which as a also returns the signer's public
```
apoelstra: nit: missing backticks around `secp256k1_musig_get_public_nonce`

include/secp256k1_musig.h
```diff
@@ -89,8 +89,7 @@ typedef struct {
  *
  * Fields:
  *   present: indicates whether the signers nonce is set
- *     index: index of the signer in the MuSig; set to be consistent with the order
- *            of the pubkeys provided to `secp256k1_musig_pubkey_combine`
+ *     index: index of the signer in the MuSig
```
apoelstra: Can you say "index of the signer in the MuSig key aggregation"? In the threshold case there will be a potential ambiguity, since there may be an incomplete list of signers which a user may believe requires re-indexing.

src/modules/musig/main_impl.h
```diff
@@ -58,7 +58,18 @@ static int secp256k1_musig_pubkey_combine_callback(secp256k1_scalar *sc, secp256
     return secp256k1_pubkey_load(ctx->ctx, pt, &ctx->pks[idx]);
 }
 
-int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scratch_space *scratch, secp256k1_pubkey *combined_pk, unsigned char *pk_hash32, secp256k1_musig_signer_data *signers, const secp256k1_pubkey *pubkeys, size_t n_signers) {
+
+void secp256k1_musig_signers_init(secp256k1_musig_signer_data *signers, const secp256k1_pubkey *pubkeys, size_t n_signers) {
```
apoelstra: This should be `static`, right?

apoelstra:
These tests look great! Very thorough. It took me a few reads to understand what you were doing by "combining nonces with a different signer set than was initialized", but I'm not sure the comments could be improved. This is a somewhat subtle thing and the reader probably just needs to read the code.

If you fix the nits and rebase, I believe we'll be ready to merge.

jonasnick:
addressed nits, improved comments in tests, fixed bug in test

jonasnick:
I noticed that the current API doesn't support sign-to-contract. I think we'd need to add a noncedata and noncefp argument to `session_initialize`. This would probably use the `s2c_commit_context` that's introduced in the sign-to-contract commit for libsecp (https://github.com/bitcoin-core/secp256k1/pull/572/files#diff-222d6d707e263c79c462fb223b0e3ddcR83). I'm ok with adding that in another PR.

jonasnick:
Picked most recent Schnorr commits from upstream and squashed MuSig.

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+    return 1;
+}
+
+/* Compute r = SHA256(ell, idx) */
+static void secp256k1_musig_coefficient(secp256k1_scalar *r, const unsigned char *ell, size_t idx) {
+    secp256k1_sha256 sha;
+    unsigned char buf[32];
+    secp256k1_sha256_initialize(&sha);
+    secp256k1_sha256_write(&sha, ell, 32);
+    while (idx > 0) {
```
real-or-random: I don't think it's wrong but it seems somewhat weird not to write leading zeros to the hash. Or is there a good reason, maybe to keep it platform-independent? Just checking if this is intentional.

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+        unsigned char c = idx;
+        secp256k1_sha256_write(&sha, &c, 1);
+        idx /= 0x100;
```
real-or-random: I somehow feel that a right shift by 8 is much more readable. :D But it's correct.

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+    /* The probability that this not a valid secret key is approximately 2^-128 */
+    } while (!secp256k1_ec_seckey_verify(ctx, seckey));
+    fclose(frand);
+    ret = secp256k1_ec_pubkey_create(ctx, pubkey, seckey);
+    return ret;
+}
+
+/* Sign a message hash with the given key pairs and store the result in sig */
+int sign(const secp256k1_context* ctx, unsigned char seckeys[][32], const secp256k1_pubkey* pubkeys, const unsigned char* msg32, secp256k1_schnorrsig *sig) {
+    secp256k1_musig_session musig_session[3];
```
real-or-random: replace `3` by  `N_SIGNERS´?

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+        FILE *frand;
+        unsigned char session_id32[32];
+        unsigned char pk_hash[32];
+        secp256k1_pubkey combined_pk;
+
+        /* Create combined pubkey and initialize signer data */
+        if (!secp256k1_musig_pubkey_combine(ctx, NULL, &combined_pk, pk_hash, pubkeys, N_SIGNERS)) {
+            return 0;
+        }
+        /* Create random session ID. It is absolutely necessary that the session ID
```
real-or-random: I think we should again note that the secret key will be leaked.
I can hear someone saying "Ha, those guys are wrong. I called it twice with the same session ID and it worked".

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+         * is unique for every call of secp256k1_musig_session_initialize. */
+        frand = fopen("/dev/urandom", "r");
+        if (frand == NULL || !fread(session_id32, 32, 1, frand)) {
```
real-or-random: fclose like above

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+        if (!create_key(ctx, seckeys[i], &pubkeys[i])) {
+            printf("FAILED\n");
+            return 1;
+        }
+    }
+    printf("ok\n");
+    printf("Combining public keys...");
+    if (!secp256k1_musig_pubkey_combine(ctx, NULL, &combined_pk, NULL, pubkeys, N_SIGNERS)) {
+        printf("FAILED\n");
+        return 0;
```
real-or-random: return 1

src/modules/musig/example.c
real-or-random: Actually "is unique for every call of secp256k1_musig_session_initialize" is somewhat misleading: strictly speaking, the example code violates this (but because it simulates 3 signers). This could be clarified. And/or maybe say "unique per signer" or "unique per secret key" or maybe rephrase.

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+        frand = fopen("/dev/urandom", "r");
+        if (frand == NULL || !fread(session_id32, 32, 1, frand)) {
+            return 0;
+        }
+        fclose(frand);
+        /* Initialize session */
+        if (!secp256k1_musig_session_initialize(ctx, &musig_session[i], signer_data[i], nonce_commitment[i], session_id32, msg32, &combined_pk, pk_hash, pubkeys, N_SIGNERS, i, seckeys[i])) {
+            return 0;
+        }
+        nonce_commitment_ptr[i] = nonce_commitment[i];
```
real-or-random: nit:
I think `nonce_commitment_ptr[i] = &nonce_commitment[i][0];` expresses the intention better.

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+            return 0;
+        }
+    }
+    /* Communication round 2: Exchange nonces */
+    for (i = 0; i < N_SIGNERS; i++) {
+        for (j = 0; j < N_SIGNERS; j++) {
+            if (!secp256k1_musig_set_nonce(ctx, &signer_data[i][j], &nonce[j])) {
+                /* Signer j's nonce does not match the nonce commitment. Wait
+                 * until the correct nonce is received or restart the protocol
+                 * (with a different session ID of course). */
```
real-or-random: We shouldn't suggest waiting for the correct nonce; this just makes the state machine overly complex for no reason. If the nonce is wrong, then it's just over. I think we shouldn't  even suggest restarting but aborting  (with the hint that it is safe to restart the protocol with a different session ID).

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+    }
+    for (i = 0; i < N_SIGNERS; i++) {
+        if (!secp256k1_musig_partial_sign(ctx, &musig_session[i], &partial_sig[i])) {
+            return 0;
+        }
+    }
+    /* Communication round 3: Exchange partial signatures */
+    for (i = 0; i < N_SIGNERS; i++) {
+        for (j = 0; j < N_SIGNERS; j++) {
+            if (!secp256k1_musig_partial_sig_verify(ctx, &musig_session[i], &signer_data[i][j], &partial_sig[j])) {
```
real-or-random: Maybe add a note on the need/non-need for partial verification.

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+    return secp256k1_musig_partial_sig_combine(ctx, &musig_session[0], sig, partial_sig, N_SIGNERS);
+}
+
+ int main(void) {
+    secp256k1_context* ctx;
+    int i;
+    unsigned char seckeys[N_SIGNERS][32];
+    secp256k1_pubkey pubkeys[N_SIGNERS];
+    secp256k1_pubkey combined_pk;
+    unsigned char msg[32] = "this_should_actually_be_msg_hash";
```
real-or-random: Nit: it is okay to sign messages, which are not hashes.
`this_could_be_the_hash_of_a_msg!` ;)

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
+#ifndef SECP256K1_MUSIG_H
+#define SECP256K1_MUSIG_H
+
+/** This module implements a Schnorr-based multi-signature scheme called MuSig
+ * (https://eprint.iacr.org/2018/068.pdf).
+ */
```
real-or-random: maybe mention the example file

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *                   summing the participants' nonces. Needed to ensure the nonce's y
+ *                   coordinate has a quadratic-residue y coordinate
+ *              msg: The 32-byte message (hash) to be signed
+ *       msg_is_set: Whether the above message has been set
+ *  has_secret_data: Whether this session object has a signers' secret data; if this
+ *                   is `false`, it may still be used for verification purposes.
+ *           seckey: If `has_secret_data`, the signer's secret key
+ *         secnonce: If `has_secret_data`, the signer's secret nonce
+ *            nonce: If `has_secret_data`, the signer's public nonce
+ *        nonce_commitments_hash: If `has_secret_data` and
```
real-or-random: whitespace

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *    been received.
+ *
+ * 3. Each individual data struct should be updated with `musig_set_nonce` once a
+ *    nonce is available. This function takes a single signer data struct rather than
+ *    an array because it may fail in the case that the provided nonce does not match
+ *    the commitment. In this case, it is desirable to identify the exact party whose
+ *    nonce was inconsistent.
+ *
+ * Fields:
+ *   present: indicates whether the signers nonce is set
```
real-or-random: signer's

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *  `musig_partial_signature_serialize` and `musig_partial_signature_parse`
+ *  functions.
+ */
+typedef struct {
+    unsigned char data[32];
+} secp256k1_musig_partial_signature;
+
+/** Computes a combined public key and the hash of the given public keys
+ *
+ *  Returns: 1 when the public keys were successfully combined, 0 otherwise
```
real-or-random: when -> if

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *                     require sharing nonces before the message is known because it
+ *                     reduces nonce misuse resistance. If NULL, must be set with
+ *                     `musig_session_set_msg` before signing and verifying.
+ *        combined_pk: the combined public key of all signers (cannot be NULL)
+ *          pk_hash32: the 32-byte hash of the signers' individual keys (cannot be
+ *                     NULL)
+ *            pubkeys: input array of public keys. Array length must equal to
+ *                     `n_signers`
+ *          n_signers: length of signers and pubkeys array. Number of signers
+ *                     participating in the MuSig
```
real-or-random: must be > 0 ? (Or say that pubkeys cannot be NULL)

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *    `musig_get_public_nonce`, which as a also returns the signer's public nonce.
+ *    This ensures that the public nonce is not exposed until all commitments have
+ *    been received.
+ *
+ * 3. Each individual data struct should be updated with `musig_set_nonce` once a
+ *    nonce is available. This function takes a single signer data struct rather than
+ *    an array because it may fail in the case that the provided nonce does not match
+ *    the commitment. In this case, it is desirable to identify the exact party whose
+ *    nonce was inconsistent.
+ *
```
real-or-random: Maybe add that a session struct can be reused for a second session if it's initilaized again.

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *  and partial signatures. It does not have secret key material and therefore can not
+ *  be used to create signatures.
+ *
+ *  Returns: 1 when session is successfully initialized, 0 otherwise
+ *  Args:        ctx: pointer to a context object (cannot be NULL)
+ *  Out:     session: the session structure to initialize (cannot be NULL)
+ *           signers: an array of signers' data to be initialized. Array length must
+ *                    equal to `n_signers`(cannot be NULL)
+ *  In:        msg32: the 32-byte message to be signed If NULL, must be set with
+ *                    `musig_session_set_msg` before using the session for signing
```
real-or-random: this session cannot be used for signing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+    secp256k1_musig_partial_signature *partial_sig
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Checks that an individual partial signature is valid
+ *
+ *  This function is essential when using protocols with adaptor signatures.
+ *  However, it is not essential for regular MuSig's, in the sense that if any
+ *  partial signatures are invalid, the full signature will also be invalid, so the
+ *  problem will be caught. But this function allows determining the specific party
+ *  who produced an invalid signature, so that signing can be restarted without them.
```
real-or-random: With dicemix in mind, it's great to see this!

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *  Args:        ctx: pointer to a context object (cannot be NULL)
+ *           session: session to update with the combined nonce (cannot be NULL)
+ *           signers: an array of signers' data, which must have had nonces set with
+ *                    `musig_set_nonce`. Array length must equal to `n_signers`
+ *                    (cannot be NULL)
+ *         n_signers: the length of the signers array. Must be the total number of
+ *                    signers participating in the MuSig.
+ *  Out: nonce_is_negated: a pointer to an integer that indicates if the combined
+ *                    nonce had to be negated.
+ *           adaptor: point to add to the combined nonce.
```
real-or-random: "In:" 
can be NULL?

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+        memcpy(session->msg, msg32, 32);
+        session->msg_is_set = 1;
+    } else {
+        session->msg_is_set = 0;
+    }
+    memcpy(&session->combined_pk, combined_pk, sizeof(*combined_pk));
+    memcpy(session->pk_hash, pk_hash32, 32);
+    session->nonce_is_set = 0;
+    session->has_secret_data = 1;
+    if (n_signers == 0 || my_index >= n_signers) {
```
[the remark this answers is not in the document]
real-or-random: Okay it is actually ARG_CHECKED implicitly because pubkeys cannot be NULL

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+    int nonce_is_negated;
+    unsigned char msg[32];
+    int msg_is_set;
+    int has_secret_data;
+    unsigned char seckey[32];
+    unsigned char secnonce[32];
+    secp256k1_pubkey nonce;
+    unsigned char nonce_commitments_hash[32];
+    int nonce_commitments_hash_is_set;
+} secp256k1_musig_session;
```
real-or-random: Can/Should a session struct be reused for a second session if it's initilaized again?

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+        if (!signers[i].present) {
+            return 0;
+        }
+        secp256k1_sha256_write(&sha, signers[i].nonce_commitment, 32);
+        secp256k1_pubkey_load(ctx, &noncep, &signers[i].nonce);
+        secp256k1_gej_add_ge_var(&combined_noncej, &combined_noncej, &noncep, NULL);
+    }
+    secp256k1_sha256_finalize(&sha, nonce_commitments_hash);
+    if (session->has_secret_data
+            && memcmp(session->nonce_commitments_hash, nonce_commitments_hash, 32) != 0) {
```
real-or-random: We should access `session->nonce_commitments_hash` only if `     session->nonce_commitments_hash_is_set` (undefined behavior otherwise)

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ * nonce_commitment: commitment to the nonce, or all-bits zero if a commitment
+ *                   has not yet been set
+ */
+typedef struct {
+    int present;
+    size_t index;
+    secp256k1_pubkey pubkey;
+    secp256k1_pubkey nonce;
+    unsigned char nonce_commitment[32];
+} secp256k1_musig_session_signer_data;
```
real-or-random: How about not giving this to every function but instead only to the init functions. I think otherwise one way to misuse the current API is to use a session with the wrong signers' data.

src/modules/musig/main_impl.h
real-or-random: Maybe you actually wanted to test for that and not for `has_secret_data`?

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+    const secp256k1_pubkey *pubkeys,
+    const unsigned char *const *commitments,
+    size_t n_signers
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5) SECP256K1_ARG_NONNULL(6) SECP256K1_ARG_NONNULL(7) SECP256K1_ARG_NONNULL(8);
+
+/** Checks a signer's public nonce against a commitment to said nonce, and update
+ *  data structure if they match
+ *
+ *  Returns: 1: commitment was valid, data structure updated
+ *           0: commitment was invalid, nothing happened
```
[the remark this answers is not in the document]
real-or-random: Hm now that I think about it again, destroying may be weird if a similar API should be used for threshold signatures, where we may want to continue if a single signer fails.

src/modules/musig/tests_impl.h
```diff
@@ -0,0 +1,704 @@
+/**********************************************************************
+ * Copyright (c) 2018 Andrew Poelstra                                 *
```
real-or-random: Maybe change the name ;)

src/modules/musig/main_impl.h
```diff
@@ -0,0 +1,586 @@
...
+    secp256k1_musig_signers_init(signers, pubkeys, n_signers);
+
+    memcpy(session->pk_hash, pk_hash32, 32);
+    session->nonce_is_set = 0;
+    session->msg_is_set = 0;
+    if (msg32 != NULL) {
+        memcpy(session->msg, msg32, 32);
+        session->msg_is_set = 1;
+    }
+    session->has_secret_data = 0;
```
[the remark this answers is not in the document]
real-or-random: Maybe it's a good idea to memset the entire struct to zero in the beginning.

src/modules/musig/tests_impl.h
```diff
@@ -0,0 +1,704 @@
...
+    CHECK(musig_state_machine_missing_msg_test(pk, &combined_pk, pk_hash, nonce_commitment[0], &nonce[0], &partial_sig[0], sk[1], session_id[1], NULL) == 0);
+    CHECK(musig_state_machine_missing_msg_test(pk, &combined_pk, pk_hash, nonce_commitment[0], &nonce[0], &partial_sig[0], sk[1], session_id[1], msg) == 1);
+
+    /* Can't verify and combine partial sigs until nonces are combined */
+    CHECK(musig_state_machine_missing_combine_test(pk, &combined_pk, pk_hash, nonce_commitment[0], &nonce[0], &partial_sig[0], msg, sk[1], session_id[1], &partial_sig[1], 0) == 0);
+    CHECK(musig_state_machine_missing_combine_test(pk, &combined_pk, pk_hash, nonce_commitment[0], &nonce[0], &partial_sig[0], msg, sk[1], session_id[1], &partial_sig[1], 1) == 1);
+}
+
+void scriptless_atomic_swap(secp256k1_scratch_space *scratch) {
+    /* Thoughout this test "a" and "b" refer to two hypothetical blockchains,
```
real-or-random: Throughout

include/secp256k1_musig.h

src/modules/musig/main_impl.h

include/secp256k1_musig.h

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+
+/** This module implements a Schnorr-based multi-signature scheme called MuSig
+ * (https://eprint.iacr.org/2018/068.pdf).
+ */
+
+/** Data structure containing data related to a signing session resulting in a single
+ * signature.
+ *
+ * This structure is not opaque, but it is strongly discouraged to copy it or read or
+ * write to it directly. A signer who is online throughout the whole process and can
```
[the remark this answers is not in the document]
jonasnick: The idea with "strongly discouraged" was that that would still allow the possibility of offline (for pre-sharing nonces) or memoryless signers. I suggested to @apoelstra that we make this text shorter and just say that no one should ever copy or serialize the session. I am changing the text "strongly discouraged" to MUST NOT.

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ * keep this structure in memory can use the provided API functions for a safe
+ * standard workflow.
+ *
+ * A signer who goes offline and needs to import/export or save/load this structure
```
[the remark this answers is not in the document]
jonasnick: Yeah I suggest we remove that part for now

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ * **must** take measures prevent replay attacks wherein an old state is loaded and
+ * the signing protocol forked from that point. One straightforward way to accomplish
+ * this is to attach the output of a monotonic non-resettable counter (hardware
+ * support is needed for this). Increment the counter before each output and
+ * encrypt+sign the entire package. If a package is deserialized with an old counter
+ * state or bad signature it should be rejected.
```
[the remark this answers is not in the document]
jonasnick: That's what I was thinking, either remove the part or implement that functionality

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *
+ * The workflow for this structure is as follows:
+ *
+ * 1. This structure is initialized with `musig_session_initialize` or
+ *    `musig_session_initialize_verifier`, which set the `pubkey` and `index` fields,
+ *    and zero out all other fields. The public session is initialized with the
+ *    signers' nonce_commitments.
+ *
+ * 2. In a non-public session the nonce_commitments are set with the function
+ *    `musig_get_public_nonce`, which as a also returns the signer's public nonce.
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *  Out:      session: the session structure to initialize (cannot be NULL)
+ *            signers: an array of signers' data to be initialized. Array length must
+ *                     equal to `n_signers` (cannot be NULL)
+ * nonce_commitment32: filled with a 32-byte commitment to the generated nonce
+ *                     (cannot be NULL)
+ *  In:  session_id32: a *unique* 32-byte ID to assign to this session (cannot be
+ *                     NULL). If a non-unique session_id32 was given then a partial
+ *                     signature will LEAK THE SECRET KEY.
+ *              msg32: the 32-byte message to be signed. Shouldn't be NULL unless you
+ *                     require sharing nonces before the message is known because it
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *                  with `musig_session_initialize_verifier`.
+ *  In:     nonce: signer's alleged public nonce (cannot be NULL)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_set_nonce(
+    const secp256k1_context* ctx,
+    secp256k1_musig_session_signer_data *signer,
+    const secp256k1_pubkey *nonce
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Updates a session with the combined nonce of all signers. The combined nonce
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+    secp256k1_musig_session *session,
+    const secp256k1_musig_session_signer_data *signers,
+    size_t n_signers,
+    int *nonce_is_negated,
+    const secp256k1_pubkey *adaptor
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4);
+
+/** Sets the message of a session if previously unset
+ *
+ *  Returns 1 if message is successfully set and 0 otherwise
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *  Out:   out32: pointer to a 32-byte array to store the serialized signature
+ *  In:      sig: pointer to the signature
+ */
+SECP256K1_API int secp256k1_musig_partial_signature_serialize(
+    const secp256k1_context* ctx,
+    unsigned char *out32,
+    const secp256k1_musig_partial_signature* sig
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Parse and validate a MuSig partial signature.
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ *
+ *  Returns: 1 when the signature could be parsed, 0 otherwise.
+ *  Args:    ctx: a secp256k1 context object
+ *  Out:     sig: pointer to a signature object
+ *  In:     in32: pointer to the 32-byte signature to be parsed
+ *
+ *  After the call, sig will always be initialized. If parsing failed or the
+ *  encoded numbers are out of range, signature validation with it is
```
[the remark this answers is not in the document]
jonasnick: fixing

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+    const secp256k1_context* ctx,
+    const secp256k1_musig_session *session,
+    const secp256k1_musig_session_signer_data *signer,
+    const secp256k1_musig_partial_signature *partial_sig
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Combines partial signatures
+ *
+ *  Returns: 1: all partial signatures had valid data. Does NOT mean the resulting
+ *              signature is valid.
```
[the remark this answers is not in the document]
jonasnick: I'm mostly changing it to "verify". I left some traces of "valid data" and "invalid signature" because they can't be replaced with verify easily

include/secp256k1_musig.h
```diff
@@ -0,0 +1,428 @@
...
+ */
+SECP256K1_API int secp256k1_musig_partial_sig_adapt(
+    const secp256k1_context* ctx,
+    secp256k1_musig_partial_signature *adaptor_sig,
+    const secp256k1_musig_partial_signature *partial_sig,
+    const unsigned char *sec_adaptor32,
+    int nonce_is_negated
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Extracts a secret adaptor from a MuSig, given all party's partial
```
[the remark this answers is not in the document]
real-or-random: yes, sorry, too many quotation marks in my comment

src/modules/musig/main_impl.h
jonasnick: I don't think this has anything to do with platform independence (@apoelstra). I'm happy to fill the remaining bytes with 0's so we're always writing the full size_t.

src/modules/musig/main_impl.h
jonasnick: agreed

src/modules/musig/example.c
```diff
@@ -0,0 +1,145 @@
...
+#include <secp256k1_musig.h>
+
+ /* Number of public keys involved in creating the aggregate signature */
+#define N_SIGNERS 3
+ /* Create a key pair and store it in seckey and pubkey */
+int create_key(const secp256k1_context* ctx, unsigned char* seckey, secp256k1_pubkey* pubkey) {
+    int ret;
+    FILE *frand = fopen("/dev/urandom", "r");
+    do {
+        if (frand == NULL || !fread(seckey, 32, 1, frand)) {
```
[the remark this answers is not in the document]
jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

src/modules/musig/example.c
jonasnick: I can just remove the "Wait until the correct nonce is received or" part.

src/modules/musig/example.c
jonasnick: Do you have a suggestion? In practice I think you'll always want to verify partial sigs if just for debugging.

src/modules/musig/example.c
jonasnick: fixing

include/secp256k1_musig.h
jonasnick: fixing

include/secp256k1_musig.h
jonasnick: fixing

include/secp256k1_musig.h
jonasnick: both. fixing

include/secp256k1_musig.h
jonasnick: fixing

src/modules/musig/main_impl.h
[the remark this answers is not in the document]
jonasnick: you don't need it because you won't be able to call `get_public_nonce` with a verifier session, but I can be more explicit. Memsetting the struct seems like a good idea.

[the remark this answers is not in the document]
real-or-random: Oh yes, you're right

src/modules/musig/main_impl.h
real-or-random: maybe (VERIFY_)CHECK it

include/secp256k1_musig.h
jonasnick: I agree that it's a problem that callers may use the session with the wrong signers. I hope I've caught and tested all cases where this could be a security issue.

I can't really imagine how the session would reasonably manage the signers though. The session can't allocate the signers on the stack because we don't have variable length arrays and we don't want to use them. Keeping a pointer in the session to signer data which is not manage by the session itself seems dangerous.

src/modules/musig/tests_impl.h
jonasnick: fixing

src/modules/musig/tests_impl.h
jonasnick: adding my name

src/modules/musig/main_impl.h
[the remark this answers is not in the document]
jonasnick: Following @apoelstra's rule of thumb ("I prefer to use ARG_CHECK only for things that could be prevented with Rust's type system") I think that it's overkill to use ARG_CHECK here.

include/secp256k1_musig.h
jonasnick: yes if we memset it to 0

src/modules/musig/main_impl.h
jonasnick: adding VERIFY_CHECK

src/modules/musig/main_impl.h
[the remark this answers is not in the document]
jonasnick: I'm adding the memset for both session types

include/secp256k1_musig.h
jonasnick: Added a memset, so yes. I'll add a test too.

src/modules/musig/example.c
real-or-random: agreed, I was wrong here

real-or-random:
The changes look good to me. :+1:

src/modules/musig/example.c
real-or-random: sounds good to me

include/secp256k1_musig.h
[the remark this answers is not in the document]
real-or-random: agreed

src/modules/musig/main_impl.h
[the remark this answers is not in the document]
real-or-random: yep

include/secp256k1_musig.h
real-or-random: Hm, yes, I can't find a case where this could a security problem. For the commitments, we have the commitments_hash. And for the public keys, the  combined_pk is set in stone already during initialization.

A little bit strange is that the pubkeys in the signer_data are only used for partial verification, and the init functions get the pubkeys and the combined_pk. If the pubkeys are anyway used only for partial verification, couldn't we drop them from init? (Or alternatively let init compute the combined_pk, at least in a singing session. Just for verification, the user can still use the combine function.)

src/modules/musig/example.c
real-or-random: "To check whether signing was successful, it suffices to either verify the the combined signature with the combined public key, or verify all partial signatures of all signers individually. Verifying the combined signature is cheaper but verifying the individual signatures has the advantage that it can be used to determine which of the signers misbehaved, i.e., sent invalid partial signatures in the protocol. " or something

I think in practice, it's also good to first verify the combined sig, and only verify the individual sigs if it does not work.

src/modules/musig/example.c
real-or-random: I think removing that part is best then.

jonasnick: fixing

src/modules/musig/example.c
jonasnick: fixing

include/secp256k1_musig.h
real-or-random: If you agree that moving pubkeys to `partial_verify` makes sense, then we could at least get rid of them in `session_initialize`.

The more I think about it, the more it makes sense to me. Maybe I'm oversimplifying but doesn't this even mean that you don't need to store them at all (if you don't care about partial verification)? I mean if the two of us just want to spend these coins under our combined public key pk, then we neither need to remember each other's public key nor send our public keys over the wire again. We just need to know H(pk0,pk1) and agree on our indices, but these can be canonical such as 0=software wallet, 1=hardware wallet. So it can even make sense to run the protocol without even knowing the other public keys (if I'm not mistaken, which could very well be).

src/modules/musig/example.c
```diff
@@ -85,9 +85,9 @@ int sign(const secp256k1_context* ctx, unsigned char seckeys[][32], const secp25
     for (i = 0; i < N_SIGNERS; i++) {
         for (j = 0; j < N_SIGNERS; j++) {
             if (!secp256k1_musig_set_nonce(ctx, &signer_data[i][j], &nonce[j])) {
-                /* Signer j's nonce does not match the nonce commitment. Wait
-                 * until the correct nonce is received or restart the protocol
-                 * (with a different session ID of course). */
+                /* Signer j's nonce does not match the nonce commitment. In this
+                 * case, restart the protocol by creating a new session (with a fresh
+                 * session ID of course). */
```
real-or-random: nit: I'd even say "In this case, abort the protocol. It's possible to restart the protocol..."

src/modules/musig/example.c
real-or-random: nit: sorry, the text was my suggestion but now I realize that "misbehaved" is too much. There could be a network attacker flipping bits etc.
"Verifying the combined signature is cheaper but verifying the individual partial signatures has the advantage that it can be used to determine which of the partial signatures are invalid (if any), i.e., which of the partial signatures cause the combined signature to be invalid and thus the protocol run to fail."

include/secp256k1_musig.h
jonasnick: I moved pubkeys to partial verify

jonasnick:
Added commit with nit fix

src/modules/musig/main_impl.h
jonasnick: I just noticed that this *is* kind of crucial because we want the MuSig coefficient to be simple and easily implemented by others to come to some kind of interoperable MuSig standard.

apoelstra: Sure, we can always write 4 bytes. I did this so I wouldn't have to think about maximum signer size sets.

src/modules/musig/example.c
apoelstra: I think "misbehaved" is fine, though it does have a bit of a connotation that there was deliberate action.

src/modules/musig/main_impl.h
jonasnick: so I'm somewhat siding with the original implementation now. Added a comment `The serialization of idx is least significant byte first and variable-length such that the last byte is non-zero`.

jonasnick: ... although, if you mess up the serialization you become immediately attackable via key cancellation. So maybe 4 bytes is better. Apparently it's to late for me right now to decide :).

real-or-random: Okay, so if Andrew didn't have a strong reason to use this format, I'm fine with both of course.
My slight preference is writing the 4 bytes because it's easier to re-implement (no while loop, no bit fiddling, you typically have some 32-bit type, etc...)

include/secp256k1_musig.h
```diff
@@ -161,7 +163,7 @@ SECP256K1_API int secp256k1_musig_pubkey_combine(
  *          pk_hash32: the 32-byte hash of the signers' individual keys (cannot be
  *                     NULL)
  *          n_signers: length of signers array. Number of signers participating in
- *                     the MuSig. Must be greater than 0.
+ *                     the MuSig. Must be greater than 0 and smaller than 2^32 - 1.
```
real-or-random: "at most"

src/modules/musig/tests_impl.h
```diff
@@ -122,6 +122,12 @@ void musig_api_tests(secp256k1_scratch_space *scratch) {
     CHECK(ecount == 8);
     CHECK(secp256k1_musig_session_initialize(sign, &session[0], signer0, nonce_commitment[0], session_id[0], msg, &combined_pk, pk_hash, 0, 0, sk[0]) == 0);
     CHECK(ecount == 8);
+    /* If more than UINT32_MAX fits in a size_t, test that session_initialize
+     * rejects n_signers that high. */
+    if (SIZE_MAX > ((size_t) UINT32_MAX) + 1) {
```
real-or-random: here we overflow :)

jonasnick:
fixed commit with fixed musig coefficient index and added commit to use a tagged hash

src/modules/musig/tests_impl.h

include/secp256k1_musig.h
```diff
@@ -262,7 +264,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_set_nonce(
  *                    set with `musig_set_nonce`. Array length must equal to `n_signers`
  *                    (cannot be NULL)
  *         n_signers: the length of the signers array. Must be the total number of
- *                    signers participating in the MuSig.
+ *                    signers participating in the MuSig. Must be greater than 0 and
+ *                    smaller than 2^32.
```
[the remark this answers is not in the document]

include/secp256k1_musig.h

src/modules/musig/main_impl.h
jonasnick: using 4 bytes now

real-or-random:
ACK c5e9fa2

jonasnick:
While playing with rust-bitcoin/bitcoin_hashes I noticed that `secp256k1_musig_sha256_init_tagged` is wrong because the length of the already hashed data is not initialized (called `bytes` in the sha256 struct). I thought it wouldn't matter, but this length is hashed as part of the padding (https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/hash_impl.h#L157). As a result, right now `hash(tag, data) != hash_tag(data)` where `hash_tag` is the hash function initialized by `init_tagged`.

jonasnick:
Added a fix

real-or-random:
Good catch! ACK 950054e

apoelstra:
ACK 950054e

jonasnick:
squashed

apoelstra:
ACK 2fc700a

real-or-random:
We should also include the fixes in https://github.com/bitcoin-core/secp256k1/pull/580

markblundeberg:
Congrats, guys! I've been watching from afar and I like the session ID approach.

apoelstra:
Will pull in https://github.com/bitcoin-core/secp256k1/pull/580 after it is merged, as part of a rebase.

Issue #36, opened 2018-12-24
secp256k1_surjectionproof_initialize cannot be used from python's ctypes ffi


dgpv:
I'm porting BlindTransaction() function from Elements Core (https://github.com/ElementsProject/elements/blob/5467be3b16f014b371ecf6bced73be0254755cfa/src/blind.cpp#L193) to python, using ctypes FFI.
(I want https://github.com/Simplexum/python-bitcointx to be able to blind/unblind Elements sidechain txs on its own).

working with `secp256k1_*` functions was easy until now, but I hit a problem with `secp256k1_surjectionproof_initialize`. It requires a reference to `secp256k1_surjectionproof` struct, but the size of this structure may change for different platforms or versions. This means that the only way to reliably use surjectionproof functions is to call them from C/C++ code that can include the header file for a particular version of the library. And then you probably want to ship this library with your code, or link with it statically, or else there might be size mismatch if you use the code compiled with old-version `secp256k1_surjectionproof.h` but have new-version dynamic library installed.

The calling code that do not have access to relevant header file might try to just alloc a big buffer and hope that the structure will not ever be bigger than that. But that is unclean and unreliable.

the library could export the size of the structure as a global symbol
`const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);`

or export a function that will return this size, or somehow expose this size in another way (but of course the caller itself should allocate the buffer of the specified size, not the library)

I understand that secp256k1-zkp is work in progress, and to be used on one's own risk. But it would be excellent to be able to use it (on my own risk) from python :-)

dgpv:
Note that other opaque structures in the library usually have their maximum size specified in the comments, with text like `not guaranteed to be portable, but guaranteed to be 64 bytes in size`, so it was easy to work with that - just allocate a buffer of needed size.

dgpv:
After conversation on #secp256k1 irc channel, it was determined that better way is to add two functions: one that would allocate+initialize the structure, and another one destroy it, so that client would not need to know the size.

I prepared PR #55 that implements those functions (currently it is not final, because there are questions regarding naming for the allocate+initialize function, and the need for guard values to prevent accidental misuse of destroy function on stack-allocated struct)

apoelstra:
Closed by #55

PR #37, opened 2018-12-24
export sizeof(secp256k1_surjectionproof) as SECP256K1_SURJECTIONPROOF_RAW_SIZE

dgpv:
the rationale is explained in #36 - knowing the struct size used in particular version of the library enables to alloc the struct without access to `secp256k1_surjectionproof.h`

For example, with python:

```
>>> import ctypes
>>> secp256k1 = ctypes.cdll.LoadLibrary('.libs/libsecp256k1.so')
>>> surjectionproof_size = ctypes.c_int.in_dll(secp256k1, 'SECP256K1_SURJECTIONPROOF_RAW_SIZE').value
>>> surjectionproof_size
8264
>>> proof = ctypes.create_string_buffer(surjectionproof_size)

# secp256k1.secp256k1_surjectionproof_initialize(secp256k1_blind_context, proof, ...)

```

### DIFF
--- a/include/secp256k1_surjectionproof.h
+++ b/include/secp256k1_surjectionproof.h
@@ -33,6 +33,11 @@ extern "C" {
  *  and secp256k1_surjectionproof_serialize to encode/decode proofs into a
  *  well-defined format.
  *
+ *  If you need to allocate new secp256k1_surjectionproof structure
+ *  for use with secp256k1_surjectionproof_initialize, you should allocate
+ *  a buffer of size SECP256K1_SURJECTIONPROOF_RAW_SIZE, and then treat the
+ *  data in this buffer as opaque data, as described above.
+ *
  *  The representation is exposed to allow creation of these objects on the
  *  stack; please *do not* use these internals directly.
  */
@@ -48,6 +53,8 @@ typedef struct {
     /** Borromean signature: e0, scalars */
     unsigned char data[32 * (1 + SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS)];
 } secp256k1_surjectionproof;
+
+SECP256K1_API const int SECP256K1_SURJECTIONPROOF_RAW_SIZE;
 
 /** Parse a surjection proof
  *
--- a/include/secp256k1_surjectionproof.h
+++ b/include/secp256k1_surjectionproof.h
@@ -54,7 +54,7 @@ typedef struct {
     unsigned char data[32 * (1 + SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS)];
 } secp256k1_surjectionproof;
 
-SECP256K1_API const int SECP256K1_SURJECTIONPROOF_RAW_SIZE;
+#define SECP256K1_SURJECTIONPROOF_RAW_SIZE sizeof(secp256k1_surjectionproof)
 
 /** Parse a surjection proof
  *
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -740,5 +740,11 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
 
 #ifdef ENABLE_MODULE_SURJECTIONPROOF
 # include "modules/surjection/main_impl.h"
-const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);
+/* In the headers, SECP256K1_SURJECTIONPROOF_RAW_SIZE is #defined
+ * as sizeof(secp256k1_surjectionproof), so it can be used in C code
+ * to do allocate the structure on stack, using that name.
+ * In this file, this #define is not be used, and we can safely
+ * #undef it to be able to define the exported symbol with the same name. */
+#undef SECP256K1_SURJECTIONPROOF_RAW_SIZE
+SECP256K1_API const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);
 #endif
--- a/src/secp256k1.c
+++ b/src/secp256k1.c
@@ -740,4 +740,5 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
 
 #ifdef ENABLE_MODULE_SURJECTIONPROOF
 # include "modules/surjection/main_impl.h"
+const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);
 #endif


instagibbs:
cc @apoelstra

dgpv:
If the library authors decide that this is better done with a function that returns the size, I will submit a PR for that. I am preparing to create a release of python-bitcointx with support for Elements sidechain, but having a requirement to apply the patch to the secp256k1-zkp source before using python-bitcointx for Elements is not very convenient. I would appreciate a comment from the authors about how they think this should be handled.

apoelstra:
cc @gmaxwell @sipa

apoelstra:
We had some comment on IRC and weren't clear on the motivation for this; how are you using this struct at all without the include file?

dgpv:
I ported the transaction blinding code from C++ in Elements (https://github.com/ElementsProject/elements/blob/7f6351782a1555696e69d5ffc5304005a07876cd/src/blind.cpp#L224) to python (https://github.com/Simplexum/python-bitcointx/blob/31f95df76ae163a5139ef0bf6a9ec556c9c09cfd/bitcointx/sidechain/elements.py#L777)

In SurjectOutput(), the secp256k1_surjectionproof structure is allocated on the stack, and then passed to secp256k1_surjectionproof_initialize(). To port this code cleanly to python, I needed to know the size of the structure.

>  how are you using this struct at all without the include file?

By using the change presented in this PR.

In this PR, this is done by exporting a special symbol that is statically assigned with 
`const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);`

Here is the lines in python code that do the allocation using this symbol: https://github.com/Simplexum/python-bitcointx/blob/31f95df76ae163a5139ef0bf6a9ec556c9c09cfd/bitcointx/sidechain/elements.py#L1473

dgpv:
rebased on current tip of secp256k1-zkp branch.

jonasnick:
If I understand correctly @apoelstra is asking because #36 can be interpreted as you're not using the include file. But it seems more like you are using it but you're concerned that the include file might be inconsistent with the (dynamically loaded) library. If that's the case I think it makes more sense to ship it with your code, link it statically or otherwise ensure that include file and library are consistent (as is the case with other libraries not only libsecp). If the files are inconsistent you might have more problems than just a different surjectionproof size.

dgpv:
It is normal to use dymanically loaded library without the header, especially if your are using the library through FFI of some higher-level language. Linux distos usually ship headers in a separate *-dev package or the like. If you have certain values defined in a header file (that are part of the API), you just take this constants and copy to your higher-level language library code. For example, secp256k1_generator struct in include/secp256k1_generator.h is guaranteed to be 64 bytes in size (as per the comments in the file), so I can use SECP256K1_GENERATOR_SIZE = 64 in python code. 

If there is some value that cannot be defined in a API header file (and cannot be copied verbatim), then there is usually a function to get the needed value, if it is calculated, or exported symbol with the said value, if it is defined at build-time, or other methods, but the idea is that you do not need the header file at runtime.  Otherwise, the purpose of dynamic library is lost for any languages beside C/C++ (or maybe a couple of other langs that can parse C headers natively) - everyone would need to include their copy of the library with known build-time values. This might be OK for commercial software distribution, where you probably shipping everything as a bundle, but if this is a *dynamic* library that is going to be included in distros (and secp256k1 is inclued, with headers in separate secp256k1-dev package in debian, for example), then the functions that are part of the API should be usable without headers, in my opinion.

Let's say I add C extension to python-bitcointx package with the only purpose of getting this value from the header. Then: 1) to install python-bitcointx on your system you would need C compiler installed - which is not always true even for linux systems; 2) you install python-bitcointx with this C extenstion that gets built at the time of installation, and then upgrade secp256k1-zkp on your system, and this particular value is changed in new version - you will probably get memory corruption bugs without knowing whats going on - because the extension was not rebuilt.

So the only options without having this value exposed without header seems to be:

1. Just allocate a lot of space and hope it wont ever be larger than 100K bytes, for example (terrible option)
2. Include full library with python-bitcointx package and build it at installation time (bad option)
3. Include statically-linked library (very bad option - shipping binary executables that are hard to audit)

jonasnick:
Sorry for the late reply and thanks for making your case so clearly. There was some discussion about your approach in #secp256k1 on freenode and I suggest you bring up your proposal there because I'm not knowledgeable enough in this area to mediate.

Fwiw I've had to overcome a similar problem when writing a rust bindings for MuSig including the the MuSig session type. Instead of allocating as many bytes as the size of a session, in rust it's possible to recreate the struct with the same layout as in C (see https://github.com/jonasnick/rust-secp256k1-zkp/blob/dea9e50b41bd70aae6cd705892f06e722755976d/src/ffi.rs#L39, that's what `#[repr(C)]` does). This is definitily more tedious, but has the advantage that the whole thing can be allocated on the stack (which is a design goal due to hardware wallet compatibility). I'm not sure if the same can be achieved with python.

dgpv:
> There was some discussion about your approach in #secp256k1 on freenode and I suggest you bring up your proposal there

OK, noted.

> in rust it's possible to recreate the struct with the same layout as in C 

This is only marginally better than just measuring the size of the struct by `printf("%d\n", sizeof(secp256k1_surjectionproof);` and using the resulting value in your code. This is because the value can change at any moment in the future, and you will need to always track the changes. The size depends on the constant and compilation option (SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS, #ifdef VERIFY).
There is no guarantee that the max_n_inputs constant won't be changed, other constants/#ifdefs would not be added, or new fields would not be added to accomodate some new features/requirements. You would need to track that and adjust your code. At this point, getting new size and just using it would be simpler.

And then you have to make sure that old version of your code (which uses smaller size for the struct) is never used with new version of the library. Otherwise you will get memory corruption.

dgpv:
Changed the approach a bit: in the header, I do

`#define SECP256K1_SURJECTIONPROOF_RAW_SIZE sizeof(secp256k1_surjectionproof)`

while in secp256k1.c, I #undef that name, and then declare the const variable that would be exported:

```
#undef SECP256K1_SURJECTIONPROOF_RAW_SIZE
SECP256K1_API const int SECP256K1_SURJECTIONPROOF_RAW_SIZE = sizeof(secp256k1_surjectionproof);
```

dgpv:
After conversation on #secp256k1 channel it was determined that better approach is to introduce create/destroy functions for this struct, and the create function would also call initialize function, so the caller won't need to do exctra call to init the struct. The code that wants to allocate the struct on the stack would still be able to do so, and would need to call initialize function directly, as before.

I'm closing this PR.

PR #40, opened 2019-01-03
use proper types for rangeproof min/max in unit test case

instagibbs:
Squashes some warnings, also I think was causing issues with 32-bit builds in downstream projects.


apoelstra:
ACK

PR #43, opened 2019-02-08
WIP: scratch space instead of stack for large allocations in rangeproof_sign

jonasnick:
This is a WIP because I'd like to add the ability to provide NULL instead of a scratch space - but only if this PR makes sense in general.

This reduces stack usage of `secp256k1_rangeproof_sign` from about 30,000 bytes to 1,520 bytes (estimated with `-fstack-usage`).


jonasnick:
No immediate need for this PR anymore. Closing...

PR #44, opened 2019-02-08
WIP: Borromean ring sigs with callback

real-or-random:
Different approach to reduce stack usage #43.

This is more like a proof-of-concept to show that the callback-style is doable without too ugly changes. But this has a bug. Some of the generated `s` values are wrong. I'm giving up for now after some debugging.


apoelstra:
I think we're not going to use this solution

PR #45, opened 2019-02-12
musig: add user documentation

apoelstra:


src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
+MuSig - Rogue-Key-Resistant Multisignatures Module
```
jonasnick: Should this PR also remove some of the redundant documentation from the include file? In particular the things that @real-or-random mentioned in the original MuSig PR and that we were not explicitly discouraging serializing the state.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+
+In MuSig all participants contribute key material to a single signing key,
+using the equation
+
+    P = sum_i µ_i * P_i
+
+where `P_i` is the public key of the `i`th participant and `µ_i` is a so-called
+_MuSig coefficient_ computed according to the following equation
+
+    C = H(P_1 || P_2 || ... || P_n)
```
jonasnick: C should be better called L like in the paper and in the code (`ell`)

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+        e = H(R || P || m)
+       and partial signature
+        s_i = k_i + µ_i*x_i*e
+       where `x_i` is the secret key corresponding to `P_i`.
+
+The complete signature is then the `(s, R)` where `s = sum_i s_i` and `R = sum_i R_i`.
+
+# API Usage
+
+It is essential to security that signers use a unique uniformly random none for all
```
jonasnick: none -> nonce

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+        R = sum_i R_i
+       and shared challenge
+        e = H(R || P || m)
+       and partial signature
+        s_i = k_i + µ_i*x_i*e
+       where `x_i` is the secret key corresponding to `P_i`.
+
+The complete signature is then the `(s, R)` where `s = sum_i s_i` and `R = sum_i R_i`.
+
+# API Usage
```
jonasnick: cross reference example.c?

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+and hash `C` in the out-pointer `pk_hash32`, if this pointer is non-NULL.
+
+## Signing
+
+A participant who wishes to sign a message (as opposed to observing/auditing the
+signature process, which is also a supported mode) acts as follows.
+
+### Signing Participant
+
+    1. Starts the session by calling `secp256k1_musig_session_initialize`. This
```
jonasnick: Starts -> Start or "The signer starts"

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+    3. Once all signers nonce commitments have been received, the signer records
+       these commitments with the function `secp256k1_musig_session_get_public_nonce`.
+       This function updates in place
+           * the session state `session`
+           * the array of signer data `signers`
+       taking in as input the list of commitments `commitments` and outputting the
+       signer's public nonce `R_i` in the out-pointer `nonce`.
+    4. The signer then communicates `R_i` to all other signers, and receives `R_j`
+       from each signer `j`. On receipt of a nonce `R_j` he calls the function
+       `secp256k1_musig_set_nonce` to record this fact. This function checks that
```
jonasnick: the function must also be called for the signer's own nonce.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+       the received nonce is consistent with the previously-received nonce and will
+       return 0 in this case.
+       These nonces `R_i` are secp256k1 public keys; they should be serialized using
+       `secp256k1_ec_pubkey_serialize` and parsed with `secp256k1_ec_pubkey_parse`.
+    5. Once all nonces have been exchanged in this way, signers are able to compute
+       their partial signatures. They do so by calling `secp256k1_musig_session_combine_nonces`
+       which updates in place
+           * the session state `session`
+           * the array of signer data `signers`
+       It outputs an auxillary integer `nonce_is_negated` and has an auxillary input
```
jonasnick: auxillary -> auxiliary

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+    2. Together, the parties execute steps 1-4 of the signing protocol above.
+    3. At step 5, when combining the two parties' public nonces, both parties call
+       `secp256k1_musig_session_combine_nonces` with `adaptor` set to `T` and `nonce_is_negated`
+       set to a non-NULL pointer to int.
+    4. Steps 6 and 7 proceed as before. Step 8, verifying the partial signatures, is now
+       essential to the security of the protocol and must not be omitted!
+
+The above steps are executed identically for both signing sessions. However, step 9 will
+not work as before, since the partial signatures will not add up to a valid total signature.
+Additonal steps must be taken, and it is at this point that the two signing sessions
```
jonasnick: Additonal -> Additional

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+       signatures to verify their correctness. If only the validity of the final signature
+       is important, not assigning blame, this step can be skipped.
+    9. Any signer, or central coordinator, may combine the partial signatures to obtain
+       a complete signature using `secp256k1_musig_partial_sig_combine`. This function takes
+       a signing session and array of MuSig partial signatures, and outputs a single
+       Schnorr signature.
+
+### Non-signing Participant
+
+A participant who wants to verify the signing process but not actually contribute a
```
jonasnick: Maybe mention that verifying the signing process means verifying nonce commitments and partial signatures.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+### Verifier
+
+The final signature is simply a valid Schnorr signature using the combined public key. It
+can be verified using the `secp256k1_schnorrsig_verify` with the correct message and
+public key output from `secp256k1_musig_pubkey_combine`.
+
+## Atomic Swaps
+
+The signing API supports the production of "adaptor signatures", modified partial signatures
+which have an auxiliary secret encrypted to them. A partial signature is revealed to a party
```
jonasnick: Encrypted to whom? partial signatures?

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+who learns the secret; conversely, a party with the corresponding secret will learn the
+secret.
```
jonasnick: there seems to be something wrong with this sentence

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+Additonal steps must be taken, and it is at this point that the two signing sessions
+diverge. From here on we consider "Session A" which benefits Alice (e.g. which sends her
+coins) and "Session B" which benefits Bob (e.g. which sends him coins).
+
+    5. In Session B, Bob calls `secp256k1_musig_partial_sig_adapt` with his partial signature
+       and `t`, to produce an adaptor signature. He can then call `secp256k1_musig_partial_sig_combine`
+       with this adaptor signature and Alice's partial signature, to produce a complete
+       signature for blockchain B.
+    6. Alice reads this signature from blockchain B. She calls `secp256k1_musig_extract_secret_adaptor`,
+       passing the complete signature along with her and Bob's partial signatures. This function
```
jonasnick: Mention that partial signatures are from session B like "her and Bob's partial signatures from session B"

src/modules/musig/musig.md
```diff
@@ -0,0 +1,189 @@
...
+public key output from `secp256k1_musig_pubkey_combine`.
+
+## Atomic Swaps
+
+The signing API supports the production of "adaptor signatures", modified partial signatures
+which have an auxiliary secret encrypted to them. A partial signature is revealed to a party
+who learns the secret; conversely, a party with the corresponding secret will learn the
+secret.
+
+Using these adaptor signatures, two 2-of-2 MuSig signing protocols can be executed in
```
jonasnick: Maybe before explaining atomic swap in detail we could more abstractly explain what the adaptor signature is. Something like
```
Given a signing session where combine_nonces was called with public key `T = t*G` as `adaptor`, and partial signatures from all participants (verified)
    1. This session's combined signature must be made by one party adapting their partial signature with `t` and combining it with the remaining partial signatures.
    2. All participants can compute `t` from the session's combined signature and all previously received (adaptorless/un-adapted?) partial signatures
```

src/modules/musig/musig.md
apoelstra: I think the redundancy is fine. Users will probably read this document zero or one times, vs the include file which is the obvious location of the API reference.

src/modules/musig/musig.md
apoelstra: Reworded.

src/modules/musig/musig.md
instagibbs: Indeed, I will get very confused since I check the include file.

jonasnick: I mean mostly the part about the memoryless operation with the counter (https://github.com/ElementsProject/secp256k1-zkp/pull/35#discussion_r251845939). Imo we shouldn't mention this when we don't really have support for it. Sounds a bit confusing for people who *just* read the include file and not not the `.md` because its at such a prominent position. Maybe include file should also mention this (`.md) file?

apoelstra: Agreed, we should add a reference to this document in the include file (and maybe we could eliminate some of the memoryless operation stuff there). Will add a commit.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+7. The signer then communicates the partial signature `s_i` to all other signers, or
+   to a central coordinator. These partial signatures should be serialized using
+   `musig_partial_signature_serialize` and parsed using `musig_partial_signature_parse`.
+8. Each signer calls `secp256k1_musig_partial_sig_verify` on the other signers' partial
+   signatures to verify their correctness. If only the validity of the final signature
+   is important, not assigning blame, this step can be skipped.
+9. Any signer, or central coordinator, may combine the partial signatures to obtain
+   a complete signature using `secp256k1_musig_partial_sig_combine`. This function takes
+   a signing session and array of MuSig partial signatures, and outputs a single
+   Schnorr signature.
```
grangerh: Just a small question: why is there no function for session closing / destruction?

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+To produce a multisignature `(s, R)` on a message `m` using verification key
+`P`, signers act as follows:
+
+1. Each computes a nonce, or ephemeral keypair, `(k_i, R_i)`. Every participant
+   communinicates `H(R_i)` to every other participant.
+2. Upon receipt of every `H(R_i)`, each participant communicates `R_i` to every
+   other participant. The recipients check that each `R_i` is consistent with
+   the previously-communicated hash.
+3. Each participant computes a combined nonce
+    R = sum_i R_i
```
ysangkok: This line is not getting formatted as intended in the rendered Markdown view on Github.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+   and shared challenge
+    e = H(R || P || m)
+   and partial signature
+    s_i = k_i + µ_i*x_i*e
```
ysangkok: These asterisks are triggering italics.

src/modules/musig/musig.md
apoelstra: Oh, yes, we should add one that zeroes out secret fields. I think we just overlooked this.

src/modules/musig/musig.md
apoelstra: Do you think I should use the Unicode × or is there a way to make the * renders properly?

real-or-random: Unicode is fine but please use the Unicode \cdot equivalent at least. :stuck_out_tongue_winking_eye:

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+    L = H(P_1 || P_2 || ... || P_n)
+    µ_i = H(L || i)
+
+where H is a hash function modelled as a random oracle.
+
+To produce a multisignature `(s, R)` on a message `m` using verification key
+`P`, signers act as follows:
+
+1. Each computes a nonce, or ephemeral keypair, `(k_i, R_i)`. Every participant
+   communinicates `H(R_i)` to every other participant.
```
real-or-random: typo

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+===========================
+
+This module implements the MuSig [1] multisignature scheme. The majority of
+the module is an API designed to be used by signing or auditing participants
+in a multisignature scheme. This involves a somewhat complex state machine
+and significant effort has been taken to prevent accidental misuse of the
+API in ways that could lead to accidental signatures or loss of key material.
+
+# Theory
+
```
real-or-random: This paragraph uses "signers" and "participants" for the same thing. That's not necessarily bad it actually reminded me that there can be participants who are not signers (just listening -- and the code supports that). So maybe we should make that clearer.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
+MuSig - Rogue-Key-Resistant Multisignatures Module
+===========================
+
+This module implements the MuSig [1] multisignature scheme. The majority of
+the module is an API designed to be used by signing or auditing participants
+in a multisignature scheme. This involves a somewhat complex state machine
+and significant effort has been taken to prevent accidental misuse of the
+API in ways that could lead to accidental signatures or loss of key material.
```
real-or-random: It should be mentioned that MuSig creates a Schnorr signature (compatibe with the Schnorr BIP) here or somewhere else.

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+
+The complete signature is then the `(s, R)` where `s = sum_i s_i` and `R = sum_i R_i`.
+
+# API Usage
+
+The following sections describe use of our API, and are mirrored in code in `src/modules/musig/example.c`.
+
+It is essential to security that signers use a unique uniformly random nonce for all
+signing sessions, and that they do not reuse these nonces even in the case that a
+signing session fails to complete. To that end, all signing state is encapsulated
```
real-or-random: well they should not reuse it, no matter if the signing sessions fails to complete or  not. ;)

src/modules/musig/musig.md
real-or-random: ah I see that distinction is made below, so yes, I think it should be made here too

src/modules/musig/musig.md
```diff
@@ -0,0 +1,196 @@
...
+   return 0 in this case. The signer must also call this function with his own
+   nonce and his own index `i`.
+   These nonces `R_i` are secp256k1 public keys; they should be serialized using
+   `secp256k1_ec_pubkey_serialize` and parsed with `secp256k1_ec_pubkey_parse`.
+5. Once all nonces have been exchanged in this way, signers are able to compute
+   their partial signatures. They do so by calling `secp256k1_musig_session_combine_nonces`
+   which updates in place
+   - the session state `session`
+   - the array of signer data `signers`
+   It outputs an auxilary integer `nonce_is_negated` and has an auxilary input
```
real-or-random: s/auxilary/auxiliary/g

apoelstra: :( I think I accidentally added the wrong spelling to my spellchecker at some point.

src/modules/musig/musig.md
apoelstra: I think the word "even" communicates this.

real-or-random: agreed

src/modules/musig/musig.md
apoelstra: I changed this paragraph.

src/modules/musig/musig.md
apoelstra: Added reference.

real-or-random:
LGTM

apoelstra:
Removed some docs from include file; added link to Blockstream blog post. @real-or-random @jonasnick thoughts?

jonasnick:
ACK

real-or-random:
ACK

PR #46, opened 2019-02-20
WIP: Threshold signatures

apoelstra:


apoelstra:
Missing a `session_initialize_verifier` function. Will add this in a second commit.

apoelstra:
Some more algebra is at https://download.wpsoftware.net/bitcoin/wizardry/2019-02-sfdevs-threshold/slides.pdf

include/secp256k1_thresholdsig.h
```diff
@@ -0,0 +1,147 @@
...
+    const unsigned char *pk_hash,
+    int continuing,
+    const secp256k1_thresholdsig_keyshard *privshard,
+    size_t my_idx,
+    size_t other_idx,
+    const secp256k1_pubkey *pubcoeff,
+    size_t n_coeffs
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(6) SECP256K1_ARG_NONNULL(11);
+
+/** Initializes a threshold signing session
```
jonasnick: Would be good to mention that this is done once the participating signers are known. And that `signers` array refers to the participating signers.

configure.ac
```diff
@@ -139,6 +139,11 @@ AC_ARG_ENABLE(module_musig,
     [enable_module_musig=$enableval],
     [enable_module_musig=no])
 
+AC_ARG_ENABLE(module_threshold,
```
jonasnick: Maybe call the module thresholdsig for consistency with musig and with the api names?

src/modules/thresholdsig/example.c
```diff
@@ -0,0 +1,193 @@
...
+        if (!secp256k1_thresholdsig_keysplit(ctx, shards[i], pubcoeff[i], seckeys[i], N_SIGNERS, N_SIGNERS_TOTAL)) {
+            printf("FAILED\n");
+            return 1;
+        }
+    }
+    printf("ok\n");
+    printf("Verifying and combining shards...");
+    for (i = 0; i < N_SIGNERS_TOTAL; i++) {
+        size_t j;
+        /* Note that on every iteration, the inner loop will overwrite the `signing_pubkeys` array */
```
guggero: I know the `threshold.md` and `secp256k1_thresholdsig.h` state that the `seckeys[i]` is modified here too.
But wouldn't it be nice to mention it here again, since we have a comment anyway?
Maybe something like ``Note that on every iteration, the inner loop will overwrite seckeys[i] and the `signing_pubkeys` array``?

jonasnick:
Perhaps worth noting that there's this other threshold scheme http://cacr.uwaterloo.ca/techreports/2001/corr2001-13.ps that @real-or-random dug out which is a bit more complicated and doen't mix as nice with the MuSig API. But it has the very desirable property that the set of partipating signers can change even after the partial signatures have been collected. This avoids having to implement annoying and fragile protections against a single signer DoSing a signing session by first indicating presence and then dropping out.

apoelstra:
Rebased; also added a .md document that tries to summarize where we're at with this

real-or-random:
Cool, I'll have a look. Given our experience with subtle attacks in interactive signatures, my take is that we should not merge this before we have some security proof. But the updated MD helps definitively here.

src/modules/thresholdsig/main_impl.h
```diff
@@ -0,0 +1,366 @@
...
+
+static int secp256k1_thresholdsig_verify_shard_ecmult_callback(secp256k1_scalar *sc, secp256k1_ge *pt, size_t idx, void *data) {
+    secp256k1_thresholdsig_verify_shard_ecmult_context *ctx = (secp256k1_thresholdsig_verify_shard_ecmult_context *) data;
+
+    secp256k1_scalar_mul(sc, &ctx->musig_coeff, &ctx->idxn);
+    secp256k1_scalar_mul(&ctx->idxn, &ctx->idxn, &ctx->idx);
+    secp256k1_pubkey_load(NULL, pt, &ctx->pubcoeff[idx]);
+    return 1;
+}
+
```
mizhang: Is "my_index" the "threshold_index", "other_index" the "musig_index"?

src/modules/thresholdsig/threshold.md
```diff
@@ -0,0 +1,282 @@
...
+which begins with the same keys `X_i`. In particular, it does not depend
+on the threshold value `k`.
+
+Key Generation works as follows.
+
+1. All signers agree on their initial public keys `X_i` through some external
+   procedure this API does not cover.
+2. Each signer uses `secp256k1_musig_pubkey_combine` to compute the combined
+   public key `P`. This function also outputs a list `Y_i` of tweaked public
+   keys, which will be needed to validate partial signatures.
```
mizhang: Didn't get how secp256k1_musig_pubkey_combine outputs a list of tweaked public keys, could be I missed anything?

src/modules/thresholdsig/threshold.md
```diff
@@ -0,0 +1,282 @@
...
+3. Each signer modifies her secret key using `secp256k1_musig_tweak_secret_key`
+   to produce a secret key `y_i` compatible with her public key `Y_i`.
+
```
mizhang: Don't see secp256k1_musig_tweak_secret_key called in this threshold module.  Maybe related to above jonasnick's comment, the public key are not tweaked beforehand anymore.

src/modules/thresholdsig/threshold.md
```diff
@@ -0,0 +1,282 @@
...
+   to someone (or everyone) for aggregation.
+   [TODO] Things will work with more than `k` signers; if people disagree on the
+   set of signers this is a problem, so "just take the first `k`" is actually
+   likely to fail in the case of `>k`.
+6. Some participant receives all the partial signatures and combines them using
+   `secp256k1_musig_combine_partial_sigs`. The output of this function is a
+   complete signature. If the signature is invalid, at least one of the partial
+   signatures was invalid. The culprit may be identified using the function
+   `secp256k1_musig_partial_sig_verify` on the individual partial signatures.
+
```
mizhang: Each participant checks if other partial signatures are valid.  Whether the combined pubkey is correct is not checked, right?  I don't see any risk except wasting time in trying to produce final signature.  If would like to avoid this situation, can add checking combined pubkey the same time when checking nonce hash at the beginning of this procedure.

apoelstra:
@mizhang thanks a ton for your interest in this PR! Unfortunately it's still in a badly WIP state, and the documents aren't in sync with the PR (they're both basically braindumps of what I was thinking after having implemented an API, decided I didn't like it, then tried to come up with a better one).

mizhang:
Andrew,

           Any mailing list I can join?  I would like to contribute if
WIP.  How should I start?

thanks,

miya

mizhang:
Are you around Austin?  Will be possible to meet?  :-)

apoelstra:
Hey Miya, yes, I'm back in town :) - send me an email at apoelstra@blockstream.com and we can figure something out. There isn't a good mailing list for this sort of development unfortunately.

src/modules/thresholdsig/design.md
```diff
@@ -0,0 +1,134 @@
+# Status
+
+This PR has been open for a while and is blocked on progress on some
+supporting infrastructure; in particular the key setup protocol requires
+both a broadcast channel and private communication channels between
```
jonasnick: I don't think the broadcast channel requirement should be viewed as a blocker for this PR because we can't solve that problem in this library. Assuming security of this scheme holds up, to get unforgeability it should be sufficient to make the implementer ensure that the pubcoeffs are the same for every participant. Obviously this is an enormous footgun.

We could add a function to compare pubcoeff hashes and signatures (independently). But this wouldn't solve the problem in general as it's probably easy to misuse and not straightforward to implement (how do you prevent replay?). Additionally, iit would be overkill for some situations. For example, sometimes there is that trusted party that can check the output on the displays of various hardware wallets match. Or a blockchain could be used as a broadcast channel.

The broadcast requirement at signing time is not as important because it's only DoS-preventing, so it could be safely offloaded to the implementer.

I agree that more in depth security analysis is a blocker.

real-or-random: I'd say as long as we trust the broadcast channel only for robustness (= we terminate properly) and we do not trust it for unforgeability, this is not a blocker. And this should be true for this scheme. But yes, a security analysis should show that this is indeed the case.

apoelstra:
Closing as out of date. We'll redo this on top of musig2

PR #51, opened 2019-03-28
Improve explanation of key cancellation attack in whitelist.md

jonasnick:
The "garbage key" does not result in burned coins if attacker and participant cooperate.

### DIFF
--- a/src/modules/whitelist/whitelist.md
+++ b/src/modules/whitelist/whitelist.md
@@ -34,11 +34,13 @@ A less obvious scheme is to have a participant sign an arbitrary message with
 the sum of her key `P` and the whitelisted key `W`. Such a signature with the key
 `P + W` proves knowledge of either (a) discrete logarithms of both `P` and `W`;
 or (b) neither. This makes directly attacking participants' signing schemes much
-harder, but allows an attacker to whitelist arbitrary "garbage" keys by computing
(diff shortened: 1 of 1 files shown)


src/modules/whitelist/whitelist.md
```diff
@@ -60,8 +62,8 @@ delegation. However, we can get this back by introducing a new key, `P'`,
 and signing with the key `P + H(W + P')(W + P')`. This gives us the best
 of both worlds: `P'` does not need to be online to delegate, allowing it
 to be securely stored and preventing real-time attacks; `P` does need to
-be online, but its compromise only allows an attacker to whitelist "garbage
-keys", not attacker-controlled ones.
+be online, but its compromise only allows an attacker to whitelist keys it does
```
dgpv: s/it/he/ ?

dgpv:
One nitpick: attacker: it -> he.

Maybe also mention that "until attacker and the legitimate participant owning cooperate." creates incentive for attacker to demand random for giving out the cancellation key.

otherwise, ACK. (I feel like I am not in a position to ack/nack anything here, because I am an outsider to the project, but the author asked me out-of-band to review :-))

src/modules/whitelist/whitelist.md
jonasnick: fixed

PR #57, opened 2019-04-30
surjectionproof: reduce stack usage and limit proofs to anonymity set of 16

apoelstra:
This would be a softfork of Elements. Alternately we could make this constant settable at compile time for limited-memory wallets.

### DIFF
--- a/include/secp256k1_surjectionproof.h
+++ b/include/secp256k1_surjectionproof.h
@@ -12,7 +12,7 @@ extern "C" {
 #define SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS 256
 
 /** Maximum number of inputs that may be used in a surjection proof */
-#define SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS 16
+#define SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS 256
 
 /** Number of bytes a serialized surjection proof requires given the
  *  number of inputs and the number of used inputs.
@@ -52,6 +52,7 @@ typedef struct {
     unsigned char data[32 * (1 + SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS)];
 } secp256k1_surjectionproof;
 
+#ifndef USE_REDUCED_SURJECTION_PROOF_SIZE
 /** Parse a surjection proof
  *
  *  Returns: 1 when the proof could be parsed, 0 otherwise.
@@ -73,6 +74,7 @@ SECP256K1_API int secp256k1_surjectionproof_parse(
   const unsigned char *input,
   size_t inputlen
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+#endif
 
 /** Serialize a surjection proof
  *
@@ -241,6 +243,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_surjectionproof_generat
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5) SECP256K1_ARG_NONNULL(7) SECP256K1_ARG_NONNULL(8);
 
 
+#ifndef USE_REDUCED_SURJECTION_PROOF_SIZE
 /** Surjection proof verification function
  * Returns 0: proof was invalid
  *         1: proof was valid
@@ -258,6 +261,7 @@ SECP256K1_API int secp256k1_surjectionproof_verify(
   size_t n_ephemeral_input_tags,
   const secp256k1_generator* ephemeral_output_tag
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5);
+#endif
 
 #ifdef __cplusplus
 }
--- a/include/secp256k1_surjectionproof.h
+++ b/include/secp256k1_surjectionproof.h
@@ -11,6 +11,9 @@ extern "C" {
 /** Maximum number of inputs that may be given in a surjection proof */
 #define SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS 256
 
+/** Maximum number of inputs that may be used in a surjection proof */
+#define SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS 16
+
 /** Number of bytes a serialized surjection proof requires given the
  *  number of inputs and the number of used inputs.
  */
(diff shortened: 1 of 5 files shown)


real-or-random:
utACK 30760f8

apoelstra:
No, I don't think there's any reason for limited-memory wallets to verify.

src/modules/surjection/main_impl.h
```diff
@@ -221,8 +222,6 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     size_t ring_input_index = 0;
     secp256k1_gej ring_pubkeys[SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS];
     secp256k1_scalar borromean_s[SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS];
```
romanz: Would it be possible to also change the above arrays' size to `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS`?

real-or-random:
> No, I don't think there's any reason for limited-memory wallets to verify.

I agree. And then I think a configurable constant is better than a softfork.

instagibbs:
ACK wallet policy only. Really easy to support.

src/modules/surjection/main_impl.h
apoelstra: Yes. I don't know how I missed those.

romanz:
Ping :)

apoelstra:
Sorry! Been in conference mode. Almost over..

real-or-random:
The API docs of `secp256k1_surjectionproof_verify` should document the fact that the function depends on `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS`.

apoelstra:
To be clear - I should make a `./configure` flag that changes this?

real-or-random:
Hmm, I don't know. This would be the first flag that changes the behavior of the code. If we don't want that (together with a big warning in the output of configure) then I guess it's also enough to #define MAX_INPUTS to the current value only if it's not already defined.

real-or-random:
@romanz why do we actually need this if the Trezor never verifies?

romanz:
> @romanz why do we actually need this if the Trezor never verifies?

You're right, we don't need verification to run on the device - so we can leave it as is for now.

However, the generation code also uses stack to allocate `SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS` values:
https://github.com/ElementsProject/secp256k1-zkp/blob/30760f861a4991e8ca6f19c0ff5b0293059a9698/src/modules/surjection/main_impl.h#L223-L224

Since the generation needs to run on the device, would it be possible to change the constants above to `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS`?

apoelstra:
Yes, though I think dropping `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS` below 256 is still a fork.

I could use different constants for generation and verification but then the size of `secp256k1_surjectionproof` will still be huge.

romanz:
Would it be possible to allow the user to change `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS` in compile-time?
This way, we can compile only the surjection proof generation code for the TREZOR, with the constant set to a lower value (similar to https://github.com/romanz/secp256k1-zkp/commit/bdb2de91d37add6e643aea694fd396aaac4e2b8d, but only for generation-related code).

> I could use different constants for generation and verification but then the size of secp256k1_surjectionproof will still be huge.

IIUC, currently the main memory issue is with the stack, since `secp256k1_surjectionproof_generate()` [allocates](https://github.com/ElementsProject/secp256k1-zkp/blob/30760f861a4991e8ca6f19c0ff5b0293059a9698/src/modules/surjection/main_impl.h#L223-L224) around 5 * 32 * 256 = ~41kB, and we'd like to keep the stack utilization as small as possible.
We currently [allocate](https://github.com/romanz/trezor-firmware/blob/liquid/core/embed/extmod/modtrezorcrypto/modtrezorcrypto-secp256k1_zkp.h#L423-L427) `secp256k1_surjectionproof` as a global variable (taking around 32 * 256 = ~8kB). 
It's not optimal, but it works for now - and we can optimize it in a following PR :)

apoelstra:
Hmm, so, `SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS` is defined in `secp256k1_surjectionproof.h` which is actually exposed to the user. We can't really "change this at build time" because the file is included in users' projects which we don't control.

romanz:
You're right, sorry for the confusion...

I have tried your suggestion (adding a flag that reduces the "effective" stack allocation and disables parsing and verification) here: https://github.com/romanz/secp256k1-zkp/commit/f487cc82ddf0d54b8dfee7a03cc0dd21ace54dc3.

What do you think?

real-or-random:
Thanks, this LGTM in general!

I think it's better to really remove verification method entirely in this case to make it a compile-time error if you try to verify (otherwise we have consensus failures too, they're just more obvious). In other words: If we remove `secp256k1_surjectionproof_verify` entirely, its behavior remains correct whenever we offer this function; we just can't offer it in some configurations.
And I think we don't need to disable parsing, or do we?

romanz:
Sounds good, will update the branch.

apoelstra:
Can we just make it `static` if the flag is set?

romanz:
Tested ACK https://github.com/ElementsProject/secp256k1-zkp/commit/7a4e6b5b16a5862e7d9c72b3f4c4a501738372f8 (both on emulator and device using https://github.com/romanz/trezor-firmware/commit/779c518c59481f03c8d0f2bea79389386454d6b7).

Stack usage seem to be fine (computed using https://github.com/trezor/trezor-firmware/commit/8769a2a9caee3f38827483cb7c5df02652e9504d):
stack-usage

romanz:
> Can we just make it `static` if the flag is set?

Just to be sure - you meant making `secp256k1_surjectionproof_verify()` static conditionally on `USE_REDUCED_SURJECTION_PROOF_SIZE` and directly including the `surjection/main_impl.h` in `surjection/test_impl.h`?

real-or-random:
oh sure, `static` is probably exactly the thing we want.

apoelstra:
I'm working on this - there are a couple other issues with @romanz's commit which cause the tests to fail without `--enable-reduced-surjectionproof-size`. Will update the PR once I've fixed everything.

apoelstra:
@romanz can you try this?

romanz:
Many thanks for the fixes (and sorry about breaking the tests)!
Tested ACK 73f9dfc3afaff44edbb73053302511f733237b5a on TREZOR model T (both emulator and device).

include/secp256k1_surjectionproof.h
```diff
@@ -143,7 +148,8 @@ SECP256K1_API size_t secp256k1_surjectionproof_serialized_size(
  *                        e.g. in a coinjoin with others' inputs, an ephemeral tag can be given;
  *                        this won't match the output tag but might be used in the anonymity set.)
  *          n_input_tags: the number of entries in the fixed_input_tags array
- *      n_input_tags_to_use: the number of inputs to select randomly to put in the anonymity set
+ *   n_input_tags_to_use: the number of inputs to select randomly to put in the anonymity set
```
romanz: A small nit: the indentation here seems a bit off.

apoelstra: I don't think it's off. the `:`s line up

romanz: You're right, sorry for the confusion...

real-or-random:
I think this introduces a bug, I'll have more details later.

apoelstra:
Should rebase on #70 once it's merged. I think that will fix Tim's bug.

real-or-random:
I pushed a commit to fix "my" bug.

apoelstra:
ACK https://github.com/ElementsProject/secp256k1-zkp/pull/57/commits/f7e4d08dae231fa4bec0cd2a785f78e985232d15

romanz:
Tested https://github.com/ElementsProject/secp256k1-zkp/commit/f7e4d08dae231fa4bec0cd2a785f78e985232d15 on TREZOR model T (both emulator and device).

PR #59, opened 2019-05-12
Clarify how to derive alternative generator H

jonasnick:
I'm suggesting to link to this procedure from the taproot BIP (https://github.com/sipa/bips/pull/43) to create a internal taproot key without known dlog. Importing hashlib and defining F are necessary to just copy and paste and run this.


real-or-random:
tACK

If you want you could even remove the `*` at the beginning of the lines of the  sage script to make this easier to copy and paste. (Those `*` are anyway optional for typical tools that parse comments, they just need the `/**`.)

jonasnick:
good idea

PR #65, opened 2019-05-26
Improve comments for surctionproof init+alloc/destroy funcs

dgpv:
Improve comments for surctionproof init+alloc/destroy funcs
    
    The comments with 'XXX' was intended to indicate that the listed
    concerns was subject to review and change, but the code with these
    comments was merged straight away. This commit replaces comments
    with more complete text describing the issues.
    
    This also signifies that the commit that this code was introduced in is
    not anymore 'work in progress'.

### DIFF
--- a/src/modules/surjection/main_impl.h
+++ b/src/modules/surjection/main_impl.h
@@ -151,7 +151,10 @@ static size_t secp256k1_surjectionproof_csprng_next(secp256k1_surjectionproof_cs
     }
 }
 
-/* XXX secp256k1_surjectionproof_create is not a good name, because it can be confused with secp256k1_surjectionproof_generate */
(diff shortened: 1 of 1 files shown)


dgpv:
It seems important to me to remove that to avoid potential confusion for people that will read the code in the future

apoelstra:
Sounds good! I looked at these comments when merging the old PR and I thought they were fine to keep in for some long-future improvements, but I have no strong feelings about it.

dgpv:
At least the comments should have more information and no 'XXX' tag. Let me prepare better version for the comments.

dgpv:
Added detailed comments describing the issues concerned, instead.

apoelstra:
Sounds good. Honestly I think it's a bit too wordy, but again, I don't have strong feelings :)

dgpv:
ACK

PR #66, opened 2019-05-30
Rebase on upstream `e541a90e`

apoelstra:
PR'ing to run CI and to get an ack from @real-or-random who did this rebase in parallel (in about ten stages), confirming at each stage that our diffs were identical.


apoelstra:
Closing in favor of #67

PR #70, opened 2019-06-03
surjectionproof: fix malleability in surjection proof parsing

apoelstra:
Not really a big deal since surjection proofs have long since been moved to witness data in Elements, but this is not something we want to exist.


src/modules/surjection/main_impl.h
```diff
@@ -55,6 +55,15 @@ int secp256k1_surjectionproof_parse(const secp256k1_context* ctx, secp256k1_surj
...
     }
 
+    /* Check that the bitvector of used inputs is of the claimed
+     * length; i.e. the final byte has no "padding bits" set */
+    if (n_inputs % 8 != 0) {
+        const unsigned char padding_mask = ((1 << (8 - (n_inputs % 8))) - 1) << (n_inputs % 8);
+        if ((input[2 + (n_inputs + 7) / 8 - 1] & padding_mask) != 0) {
+            return 0;
+        }
+    }
```
real-or-random: Correct! But the mask expression was pretty hard to parse and understand.

I suggest
```
    if (n_inputs % 8 != 0) {
        const unsigned char padding_mask = (~0U) << (n_inputs % 8);
        if ((input[2 + (n_inputs + 7) / 8 - 1] & padding_mask) != 0) {
            return 0;
        }
    }
```
or even eliminate the if by
```
    {
        /* Check that the bitvector of used inputs is of the claimed
         * length; i.e. the final byte has no "padding bits" set */
        const unsigned char padding_mask = (n_inputs % 8 != 0) ? ((~0U) << (n_inputs % 8)) : 0;
        if ((input[2 + (n_inputs + 7) / 8 - 1] & padding_mask) != 0) {
            return 0;
        }
    }
```

apoelstra: First suggestion taken. I like having the `if` there as it gives more of a justification to the block beyond "C89 demands it" and I think it's also clearer to read.

apoelstra:
cc @real-or-random re-ack?

real-or-random:
utACK

PR #72, opened 2019-06-21
Fix schnorrsig and musig modules after rebase

jonasnick:
Schnorrsig and musig modules don't compile right now. This is also why travis complains in the open pull requests.


real-or-random:
Oh ups.

ACK read the code

PR #73, opened 2019-06-21
 Require message in musig protocol in an earlier state

jonasnick:
This PR removes the `set_msg` function from the musig module and adds a `msg` argument `get_public_nonce` which must be non-null if the message has not been set during initialization.

The reason for this is that I think that our current `set_msg` logic is vulnerable to Wagner's attack. The main problem is that by grinding the message an attacker has control over the final hash being signed.

Assume the combined pubkey of Alice and the attacker is `P`.
The attacker wants to forge a signature on message `m'` and opens 3 parallel sessions with Alice.
They exchange nonce commitments and nonces.
Let's call the combined nonces `Ri` for sessions `i = 0,1,2`.
Now the attacker draws `ka` and sets `R' = R0 + R1 + R2 + ka*G` and runs Wagner's algorithm grinding `mi` to find
```
H(P, R',m') = H(P, R1, m1) + H(P, R2, m2) + H(P, R3, m3)
```
with about `2^100` work (with more parallel sessions requires *much* less work).

Then the attacker and Alice set the message in session `i` to `mi` and Alice creates partial signatures `si = ki + H(P, Ri, mi)`.
Now `(R', s' = s0 + s1 + s2)` is a valid partial signature under Alice's public key `A` as well:
```
s'*G = (k0 + k1 + k2)*G + (H(P, R1, m1) + H(P, R2, m2) + H(P, R3, m3))*A
     = R' - ka*G + H(P, R', m')*A
 ```
This signature is completed by the attacker adding the partial signature using nonce `ka` which results in the forgery.

This PR does keeps allows to set the message after nonce commitments have been exchanged, unfortunately the possibility of pre-sharing nonces while not knowing the messages is removed.

### DIFF
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -147,9 +147,9 @@ SECP256K1_API int secp256k1_musig_pubkey_combine(
  *                     NULL). If a non-unique session_id32 was given then a partial
  *                     signature will LEAK THE SECRET KEY.
  *              msg32: the 32-byte message to be signed. Shouldn't be NULL unless you
- *                     require sharing public nonces before the message is known
+ *                     require sharing nonce commitments before the message is known
  *                     because it reduces nonce misuse resistance. If NULL, must be
- *                     set with `musig_session_set_msg` before signing and verifying.
+ *                     set with `musig_session_get_public_nonce`.
  *        combined_pk: the combined public key of all signers (cannot be NULL)
  *          pk_hash32: the 32-byte hash of the signers' individual keys (cannot be
  *                     NULL)
@@ -186,6 +186,8 @@ SECP256K1_API int secp256k1_musig_session_initialize(
  *  In:  commitments: array of 32-byte nonce commitments (cannot be NULL)
  *     n_commitments: the length of commitments and signers array. Must be the total
  *                    number of signers participating in the MuSig.
+ *             msg32: the 32-byte message to be signed. Must be NULL if already
+ *                    set with `musig_session_initialize` otherwise can not be NULL.
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_session_get_public_nonce(
     const secp256k1_context* ctx,
@@ -193,7 +195,8 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_session_get_publi
     secp256k1_musig_session_signer_data *signers,
     secp256k1_pubkey *nonce,
     const unsigned char *const *commitments,
-    size_t n_commitments
+    size_t n_commitments,
+    const unsigned char *msg32
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
 /** Initializes a verifier session that can be used for verifying nonce commitments
@@ -205,9 +208,7 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_session_get_publi
  *  Out:     session: the session structure to initialize (cannot be NULL)
  *           signers: an array of signers' data to be initialized. Array length must
  *                    equal to `n_signers`(cannot be NULL)
- *  In:        msg32: the 32-byte message to be signed If NULL, must be set with
- *                    `musig_session_set_msg` before using the session for verifying
- *                    partial signatures.
+ *  In:        msg32: the 32-byte message to be signed (cannot be NULL)
  *       combined_pk: the combined public key of all signers (cannot be NULL)
  *         pk_hash32: the 32-byte hash of the signers' individual keys (cannot be NULL)
  *       commitments: array of 32-byte nonce commitments. Array length must equal to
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by real-or-random** [comment]
Oh nice observation. That the message is fixed upfront is apparently not an artefact of the formal model either...

Concept ACK.

File: src/modules/musig/tests_impl.h [hunk 1]
```diff
@@ -440,34 +433,35 @@ int musig_state_machine_diff_signers_combine_nonce_test(secp256k1_pubkey *combin
...
-    int partial_sign, partial_verify;
 
-    CHECK(secp256k1_musig_session_initialize(ctx, &session, signers, nonce_commitment, session_id, msg, combined_pk, pk_hash, 2, 1, sk) == 1);
+    CHECK(secp256k1_musig_session_initialize(ctx, &session, signers, nonce_commitment, session_id, NULL, combined_pk, pk_hash, 2, 1, sk) == 1);
     ncs[0] = nonce_commitment_other;
     ncs[1] = nonce_commitment;
-    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2) == 1);
+
+    /* Trying to get the nonce without providing a message fails. */
+    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2, NULL) == 0);
```
real-or-random: Should we `ARG_CHECK` this violation?

src/modules/musig/tests_impl.h
```diff
@@ -440,34 +433,35 @@ int musig_state_machine_diff_signers_combine_nonce_test(secp256k1_pubkey *combin
...
+    /* Providing a messge should make get_public_nonce succeed. */
```
real-or-random: typo

src/modules/musig/tests_impl.h
jonasnick: If the rationale for `ARG_CHECK`ing is still "everything that the rust typesystem would have caught" then I guess no.

src/modules/musig/tests_impl.h

real-or-random:
ACK read the code, but did not test it

include/secp256k1_musig.h
```diff
@@ -149,7 +149,7 @@ SECP256K1_API int secp256k1_musig_pubkey_combine(
  *              msg32: the 32-byte message to be signed. Shouldn't be NULL unless you
  *                     require sharing public nonces before the message is known
  *                     because it reduces nonce misuse resistance. If NULL, must be
- *                     set with `musig_session_set_msg` before signing and verifying.
+ *                     set with `musig_session_get_public_nonce`.
```
apoelstra: We should probably cut this comment down even further to get rid of the "unless you require" clause.

src/modules/musig/tests_impl.h
apoelstra: We could prevent this with session types, i.e. have a `Session<NonceUnset>` and `Session<NonceSet>`

src/modules/musig/tests_impl.h
```diff
@@ -440,34 +433,35 @@ int musig_state_machine_diff_signers_combine_nonce_test(secp256k1_pubkey *combin
...
     ncs[0] = nonce_commitment_other;
     ncs[1] = nonce_commitment;
-    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2) == 1);
+
+    /* Trying to get the nonce without providing a message fails. */
+    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2, NULL) == 0);
+    /* Providing a message should make get_public_nonce succeed. */
+    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2, msg) == 1);
+    /* Trying to set the message again fails. */
+    CHECK(secp256k1_musig_session_get_public_nonce(ctx, &session, signers, &nonce, ncs, 2, msg) == 0);
```
apoelstra: I don't have a strong opinion but I feel like this should also be an ARG_CHECK (and that rust-secp should prevent it statically)

apoelstra:
I see a bunch of stuff when I run `valgrind ./tests 4` with this code.

include/secp256k1_musig.h
jonasnick: Hm, not sure what exactly you're suggesting. Something like that?
```
 *              msg32: the 32-byte message to be signed. Shouldn't be NULL. If NULL, must be
 *                           set with `musig_session_get_public_nonce`.
```
Seems to be less clear. Also I assume sharing nonce commitments won't be a very exotic thing. Lightning will likely use it to avoid roundtrips. I have an open PR in the scriptless scripts repo (https://github.com/apoelstra/scriptless-scripts/pull/6) that explains how to do this because I was contacted by someone concerned that MuSig will end up being very expensive in communication costs.

src/modules/musig/tests_impl.h
jonasnick: By using something like a state machine/session types? Yes, fixing.

include/secp256k1_musig.h
apoelstra: I think if you change "require sharing public nonces" to "require sharing nonce commitments" I'll be happy

src/modules/musig/tests_impl.h

include/secp256k1_musig.h

src/modules/musig/tests_impl.h
jonasnick: I considered this but thought that this is not exactly the rust type system as you could do something similar with C. But actually this is what we'll be doing in rust. Fixing.

jonasnick: Fixed

jonasnick:
rebased

jonasnick:
ping @apoelstra

real-or-random:
ACK 6a57be0

apoelstra:
ACK 6a57be09e308d87e8ff7edb04af55b5a0a1b59f2

PR #74, opened 2019-06-22
Add tweak32 parameter to musig_partial_sig_combine

jonasnick:
With this PR it's possible to create MuSig signatures for public keys with pay to contract commitments:

```
secp256k1_musig_pubkey_combine(.., &P, pk_hash, pk, ...);
secp256k1_ec_pubkey_tweak_add(ctx, &Q, ec_commit_tweak);
/* Note same pk_hash as in pubkey combine but with tweaked key Q */
secp256k1_musig_session_initialize(..., &Q, pk_hash, ...);
...
secp256k1_musig_partial_sig_combine(..., ec_commit_tweak);
```

If the signers want to create a signature for the the normal combined key `P` they run the protocol just as before (i.e. initialize the session with `P` and give `NULL` to the `partial_sig_combine` `tweak32` argument.

### DIFF
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -366,13 +366,20 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_partial_sig_verif
  *  Out:          sig: complete signature (cannot be NULL)
  *  In:  partial_sigs: array of partial signatures to combine (cannot be NULL)
  *             n_sigs: number of signatures in the partial_sigs array
+ *            tweak32: if `combined_pk` was tweaked with `ec_pubkey_tweak_add` after
+ *                     `musig_pubkey_combine` and before `musig_session_initialize` then
+ *                     the same tweak must be provided here in order to get a valid
+ *                     signature for the tweaked key. Otherwise `tweak` should be NULL.
+ *                     If the tweak is larger than the group order or 0 this function will
+ *                     return 0. (can be NULL)
  */
 SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_partial_sig_combine(
     const secp256k1_context* ctx,
     const secp256k1_musig_session *session,
     secp256k1_schnorrsig *sig,
     const secp256k1_musig_partial_signature *partial_sigs,
-    size_t n_sigs
+    size_t n_sigs,
+    const unsigned char *tweak32
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
 
 /** Converts a partial signature to an adaptor signature by adding a given secret
--- a/src/modules/musig/main_impl.h
+++ b/src/modules/musig/main_impl.h
@@ -473,7 +473,7 @@ int secp256k1_musig_partial_sign(const secp256k1_context* ctx, const secp256k1_m
     return 1;
 }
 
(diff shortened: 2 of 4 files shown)


include/secp256k1_musig.h
```diff
@@ -366,13 +366,18 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_partial_sig_verif
  *  Out:          sig: complete signature (cannot be NULL)
  *  In:  partial_sigs: array of partial signatures to combine (cannot be NULL)
  *             n_sigs: number of signatures in the partial_sigs array
+ *            tweak32: if `combined_pk` was tweaked with `tweak_add` after
```
real-or-random: nits:
 * `tweak_add` is not the full function name
 * s/if/If

src/modules/musig/main_impl.h
```diff
@@ -502,6 +502,21 @@ int secp256k1_musig_partial_sig_combine(const secp256k1_context* ctx, const secp
...
+    if (tweak32 != NULL) {
+        unsigned char msghash[32];
+        secp256k1_scalar e, scalar_tweak;
+        int overflow = 0;
+
+        if (!secp256k1_musig_compute_messagehash(ctx, msghash, session)) {
+            return 0;
+        }
+        secp256k1_scalar_set_b32(&e, msghash, NULL);
+        secp256k1_scalar_set_b32(&scalar_tweak, tweak32, NULL);
```
real-or-random: Hm, `ec_pubkey_tweak_add` checks for overflow... Maybe we should do the same just for consistency. (Not that it will matter...)

jonasnick: I did this first and then removed it because of the additional complexity and the additional test for no good reason because `ec_pubkey_tweak_add` will already fail with the same tweak. However it's possible to trigger this condition if the tweak is not output of a hash function. Perhaps I should just add the check.

include/secp256k1_musig.h
jonasnick: why capital letter for `if`? That's inconsistent with the rest.

src/modules/musig/main_impl.h
jonasnick: done

include/secp256k1_musig.h
real-or-random: It's a full sentence. But I'm not going to start an argument about this. :man_shrugging:

jonasnick:
Rearranged privkey tweaking a small bit to reuse more code from `eckey_privkey_mul`. Squashed because I'm building a PR on top of this.

apoelstra:
ACK.

This API is a little bit dangerous because it allows tweaking even without doing the p2c/taproot construction, but I think that's OK; it'd be a layer violation to put the specific p2c construction into secp-zkp.

jonasnick:
> This API is a little bit dangerous because it allows tweaking even without doing the p2c/taproot construction,

Hm, not sure if we can care about layer violations if this lib should remain high level.
I suggested a couple of weeks ago to add a taproot tweak function to libsecp upstream that does the right thing for normal/single taproot keys but Pieter argued that the existing `tweak_add` is sufficient.

Perhaps we could add an optional `tweak_func` argument to `partial_sig_combine` that by default is secure, i.e. hashes a prefix, key and tweak, with the prefix being something like "secp-zkp default musig tweak" and not the taproot prefix.

apoelstra:
I think having a default that wasn't Taproot would be a bit silly. But yeah, I would be more comfortable with such an API.

apoelstra:
On the other hand having a default that *was* Taproot is also potentially dangerous, if people were mistakenly adding Taproot spend conditions to their keys..

jonasnick:
Next step would be #77 which has to move the tweak to pubkey_combine, but is still easy to misuse.

PR #77, opened 2019-07-02
WIP: Make musig module always create and sign for pubkeys with even y-coordinates

jonasnick:
With this PR the sign of pubkeys and secret keys is flipped if necessary which is transparent to the user. The motivation for this is to allow just passing around 32 bytes of combined public keys instead of 33. With this PR the first byte of the serialized MuSig public key can be removed and then a `SECP256K1_TAG_PUBKEY_EVEN` byte (`2`) can be added before deserializing.

Builds on #74.

Tests should cover the reachable branches and were run under valgrind.

### DIFF
--- a/src/modules/musig/main_impl.h
+++ b/src/modules/musig/main_impl.h
@@ -7,6 +7,7 @@
 #ifndef _SECP256K1_MODULE_MUSIG_MAIN_
 #define _SECP256K1_MODULE_MUSIG_MAIN_
 
+#include <stdint.h>
 #include "include/secp256k1.h"
 #include "include/secp256k1_musig.h"
 #include "hash.h"
@@ -97,10 +98,12 @@ static void secp256k1_musig_signers_init(secp256k1_musig_session_signer_data *si
     }
 }
 
-int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scratch_space *scratch, secp256k1_pubkey *combined_pk, unsigned char *pk_hash32, const secp256k1_pubkey *pubkeys, size_t n_pubkeys) {
+static const uint64_t keygen_aux_magic = 0xf4adbbdf7c7dd304UL;
+int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scratch_space *scratch, secp256k1_pubkey *combined_pk, secp256k1_musig_keygen_aux *keygen_aux, const secp256k1_pubkey *pubkeys, size_t n_pubkeys, const unsigned char *tweak32) {
     secp256k1_musig_pubkey_combine_ecmult_data ecmult_data;
     secp256k1_gej pkj;
     secp256k1_ge pkp;
+    char pubkey_is_negated = 0;
 
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(combined_pk != NULL);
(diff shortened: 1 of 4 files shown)


apoelstra:
ACK

src/modules/musig/main_impl.h
```diff
@@ -561,14 +600,24 @@ int secp256k1_musig_partial_sig_verify(const secp256k1_context* ctx, const secp2
...
      * to multiplying the signer's public key by the coefficient, except
      * much easier to do. */
-    secp256k1_musig_coefficient(&mu, session->pk_hash, signer->index);
+    secp256k1_musig_coefficient(&mu, session->keygen_aux.pk_hash, signer->index);
     secp256k1_scalar_mul(&e, &e, &mu);
 
     if (!secp256k1_pubkey_load(ctx, &rp, &signer->nonce)) {
         return 0;
     }
+    memcpy(&pubkey_tmp, pubkey, sizeof(pubkey_tmp));
```
real-or-random: A structure assignment is better style here (and I think in general, reading possible struct padding bytes is UB).

src/modules/musig/main_impl.h
```diff
@@ -117,15 +120,36 @@ int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scrat
...
+    if (tweak32 != NULL) {
+        secp256k1_scalar scalar_tweak;
+        int overflow = 0;
+        secp256k1_scalar_set_b32(&scalar_tweak, tweak32, &overflow);
+        if (overflow || secp256k1_scalar_is_zero(&scalar_tweak)) {
+            /* This mimics the behavior of secp256k1_ec_pubkey_tweak_add regarding
+             * overflow and tweak32 being 0.*/
+            return 0;
+        }
+        secp256k1_eckey_pubkey_tweak_add(&ctx->ecmult_ctx, &pkp, &scalar_tweak);
```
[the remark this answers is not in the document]

src/modules/musig/main_impl.h
```diff
@@ -97,10 +98,12 @@ static void secp256k1_musig_signers_init(secp256k1_musig_session_signer_data *si
     }
 }
 
-int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scratch_space *scratch, secp256k1_pubkey *combined_pk, unsigned char *pk_hash32, const secp256k1_pubkey *pubkeys, size_t n_pubkeys) {
+static uint64_t keygen_aux_magic = 0xf4adbbdf7c7dd304UL;
```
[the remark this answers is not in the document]

src/modules/musig/main_impl.h
```diff
@@ -561,14 +600,24 @@ int secp256k1_musig_partial_sig_verify(const secp256k1_context* ctx, const secp2
...
 
     if (!secp256k1_pubkey_load(ctx, &rp, &signer->nonce)) {
         return 0;
     }
+    memcpy(&pubkey_tmp, pubkey, sizeof(pubkey_tmp));
+    if (session->keygen_aux.pubkey_is_negated) {
+        int ret = 0;
+        secp256k1_ge p;
+        ret = secp256k1_pubkey_load(ctx, &p, &pubkey_tmp);
+        if (ret) {
```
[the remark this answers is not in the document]
jonasnick: looks like I was copy and pasting this from somewhere and not careful .. fixed

src/modules/musig/main_impl.h
jonasnick: structure assignment is better style, fixed

src/modules/musig/main_impl.h
```diff
@@ -561,14 +602,24 @@ int secp256k1_musig_partial_sig_verify(const secp256k1_context* ctx, const secp2
...
     secp256k1_scalar_mul(&e, &e, &mu);
 
     if (!secp256k1_pubkey_load(ctx, &rp, &signer->nonce)) {
         return 0;
     }
+    /* Verify against negated public key if the combined key was negated. */
+    pubkey_tmp = *pubkey;
+    if (session->keygen_aux.pubkey_is_negated) {
+        secp256k1_ge p;
+        if (!secp256k1_pubkey_load(ctx, &p, &pubkey_tmp)) {
```
real-or-random: On a second thought... Have you considered adding a `nonce_is_negated` argument to `schnorrsig_real_verify` ? This would save the load/save cycle here.

jonasnick: Thanks for bringing this up. It's actually even easier. We can just negate `e`. Pushed.

jonasnick:
Changed to WIP because verifying pay to contract commitments does not work with this PR.

ChristopherA:
Does this branch now conform to "Completely switch to 32-byte public keys in bip-schnorr/taproot/tapscript" https://github.com/sipa/bips/pull/55 ?

I'd like to create something like minisign https://github.com/jedisct1/minisign that signs files with a secp schnorr signature.

Or is it too premature to do this?

jonasnick:
@ChristopherA No it doesn't conform to the current version of BIP-Schnorr. I will do that in the coming week(s). Closing for now.

PR #80, opened 2019-08-18
Fix explanation of H derivation. It doesn't use DER encoding.

jonasnick:


real-or-random:
ACK b1fb2ce62507ba64635a4471db04bbef15ddd62f

PR #82, opened 2019-08-19
Musig: add cplusplus directive and fix ARG_NONNULL macro

### LINKED ISSUE #81: MuSig build issues
As of now secp256k1-zkp builds successfully when the MuSig module is enabled, though when trying to use it there are two problems:

1 - the module is missing the extern header/footer, so it must be externed manually - not sure if this is intentional since the module is experimental, but other experimental modules like schnorrsig do include this.

2 - On line 272, the last ARG_NONNULL flag does not match the function definition (args 1-3 must be non-null, not 1-2 & 4). This results in this error when building a project that includes secp256k1-zkp built with the musig module:

secp256k1-zkp/.libs/include/secp256k1.h:153:70: error: nonnull argument references non-pointer operand (argument 1, operand 4)
 define SECP256K1_ARG_NONNULL(_x)  __attribute__ ((__nonnull__(_x)))
                                                                      ^
secp256k1-zkp/.libs/include/secp256k1_musig.h:272:53: note: in expansion of macro ‘SECP256K1_ARG_NONNULL’
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4);
                                                     ^~~~~~~~~~~~~~~~~~~~~

I've fixed this by just changing the 4 to a 3 on that line.

jonasnick:
Fixes #81

### DIFF
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -273,7 +273,7 @@ SECP256K1_API int secp256k1_musig_session_combine_nonces(
     size_t n_signers,
     int *nonce_is_negated,
     const secp256k1_pubkey *adaptor
-) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4);
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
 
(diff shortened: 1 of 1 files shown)


real-or-random:
ACK 217151df1c106684c064bdaa79a770c2566bc940 I looked at the code

PR #84, opened 2019-08-20
port of secp256k1#566 "jni: Use only Guava for hex encoding and decoding"

real-or-random:
This removes a dependency on javax.xml.bind, which is no longer
available in JDK >= 11, see
https://openjdk.java.net/jeps/320#Java-EE-modules .

This fixes Travis, whose Java builds were failing recently.


jonasnick:
ACK 682a821759a436f63457780e4cca7fe83cab3a52

PR #86, opened 2019-11-15
Update Schnorrsig and MuSig

jonasnick:
The current schnorrsig and musig module are not compatible with bip-schnorr anymore. This PR cherry-picks the following commits from https://github.com/bitcoin-core/secp256k1/pull/558:
```
84308855d6e6562bd8d52677f681ab68735d3467 f remove nonce_is_negated parameter
639a8599ec97121970566c3a1f9e6fb763471c83 f use tagged hashes in nonce derivation and signature hash
2e76c5dc6d6ff4e30aac481820648ae95353abcf f don't allow counter != 0 in nonce function
7eeb3aa49832015a8aa8caf17a008d263f373fc0 f add xonly pubkey struct which is serialized as 32 byte and whose Y coordinate is a quadratic residue
add8e78542e29c3aab0d462425a4988689cdb331 f use xonly_pubkeys in schnorrsig sign and verify
2e4ed392e1fd8cb7c64787bde9b67ddc0b463e3d f add tweak functions for xonly_pubkeys
84fe42737d9f915a53e3e682ebd11021e24bfd52 f address some of pieter's comments
e28b61c290f4549dfeb94b421310ff68b661fea0 f const casting
5764b2b2b7e9d96c7704364e55c17ed064217398 f test that pubkey is zeroed after xonly_pubkey_parse returned 0
74bb3b4318b4c17b8cb01fa9d42ee776094b1743 f don't use secp256k1_pubkeys in xonly_tweak api and instead use is_positive flag
dfce048bdec171fa76384a28136c59acd80f9b5a f remove xonly_pubkey_to_pubkey
2b36bf7db30d399878f8b32abf3fe29039f79809 f sign -> is_positive
26f95c556a088dc9e83e587d96534429a1974737 f feed seckey that is actually signed with (i.e. perhaps negated) into nonce functions
538052c95039e84c2f6b2790b826b86765f8d2c3 f reenable test vectors
b5fcf541495d78595c0bcbecdba58219a866b7f3 f uninitialized memory in tests
affa6e87e972b7d2f48a60b79b9360f8b15db021 f rename is_positive in xonly_pubkey api to has_square_y
44881b21e3876fd63a4674fdc1375709e31cd694 f rename is_positive in schnorrsig module to has_square_y
9d5373307cfad5fa1b53820d417a8155d2d557dc f add new bip-schnorr test vector 14
[...]

PR #87, opened 2019-12-03
MuSig state machine simplifictions, renames and API improvements

jonasnick:
A couple of improvements to musig that didn't quite fit into #86.


jonasnick:
Will open new PR with rebased and improved version of this that also reintroduces taproot tweaking.

PR #93, opened 2020-09-23
[upstream PR #773]: Fix some compile problems on weird/old compilers.

apoelstra:


apoelstra:
Generated this PR automatically by taking the upstream merge commit, merging that into secp256k1-zkp, and PR'ing the result. This means a total of 3 merge commits (the one upstream, the local one, and the one created by the PR).

For an alternate approach (where we just look at the upstream PR itself and try to re-PR it to secp-zkp) see https://github.com/ElementsProject/secp256k1-zkp/pull/94

apoelstra:
Closing this in favor of #94 -- I think that we shouldn't use this (heavier) approach unless there are conflicts that force us to.

PR #94, opened 2020-09-23
[upstream PR #773]: Fix some compile problems on weird/old compilers.

apoelstra:


apoelstra:
Generated this PR automatically by looking at the list of merged PRs upstream and creating new PRs (from the same commit) here.

real-or-random:
@apoelstra What's the plan here? Is this just a test for your scripts? Or should we review this?

apoelstra:
@real-or-random both. If we can merge this I can do a series of these PRs, until we hit a merge conflict, then we can decide on a process for how to deal with that.

My proposal here is that we start continuously merging PRs from upstream into libsecp-zkp, rather than rebasing (and rather than doing the subtree-based schemed that @jonasnick has been experimenting with).

jonasnick (approved):
ACK 1309c03c45beece646a7d21fdb6a0e3d38adee2b

PR #95, opened 2020-09-25
[upstream PR #741]: Remove unnecessary sign variable from wnaf_const

apoelstra:

PR #96, opened 2020-09-26

apoelstra:


real-or-random:
Is weird/wrong to have these merge commits in this PR?

jonasnick:
I wonder if it'd be possible to put multiple upstream pull requests into one PR here, otherwise this will take a while (and it'll be useful in the future when we open PR for all upstream PRs once a month or so).

apoelstra:
The most straightforward way to do that would be to create a PR from upstream `master` rather than from `HEAD` of individual PRs. We can certainly do that, though the result would be one extra merge commit in our history.

Do you think we should do this (and do "batches" every month)? My thought had been that once we caught up with upstream, we'd stay up-to-date in real time, creating a mirror PR on secp-zkp shortly after upstream PRs got merged.

I agree that this is going to take a very long time, all next week and possibly into the one after that, at the current rate of running Travis on every PR :)

apoelstra:
@real-or-random yeah it is weird that these merge commits showed up in this PR. The issue is that upstream's history isn't as linear as my scripts are pretending it is ... this wouldn't be a problem if I were PR'ing from upstream `master` rather than PR heads.

jonasnick:
Hm, on the other hand this PR's commit could be cherry-picked without conflicts on top of -zkp master. Downside: requires reviewers to run the same script to create the PR instead of searching for the commit in upstream secp. Upside: does not end up with weird history (not sure if it'll get more crazy after this PR).

apoelstra:
I'd rather not cherry-pick because losing the commit IDs will make tracing history much harder. If there is a bug ideally we could quickly find the upstream PR.

apoelstra:
Heh, so, the reason that we're having trouble here is that my idea of "copy the PRs by merging the commit ID of the PR tip" isn't the right approach. Each PR builds on top of upstream master, so if secp256k1-zkp isn't sync'd with upstream master, we're going to wind up with extra commits.

In particular, we merged #94 and #95, bringing the commits from the upstream PRs into libsecp-zkp, but not bringing the upstream merge commits (`66bb9320` and `3e5cfc5c`). So when we then try to copy upstream PR 774 here, which builds on those two commits, it copies the commits in.

**The right approach** is to instead merge from upstream master. See https://github.com/ElementsProject/secp256k1-zkp/pull/99 for an example of how to do this. The result is a bit ugly - two merge commits per PR (one "normal" one, one from upstream) (and in fact this particular one also has the two missing merge commits from #94 and #95, meaning it has one real commit and four merge commits :)), but I think this will result in the most sensible history.

real-or-random:
> **The right approach** is to instead merge from upstream master. See #99 for an example of how to do this. The result is a bit ugly - two merge commits per PR (one "normal" one, one from upstream) (and in fact this particular one also has the two missing merge commits from #94 and #95, meaning it has one real commit and four merge commits :)), but I think this will result in the most sensible history.

Yes, I think that's reasonable even with two merge commits.

jonasnick:
@apoelstra how did you create #99? Will need to adjust the script in #98.

apoelstra:
@jonasnick my new script is
```
#!/bin/sh

COMMIT=$1
PRCOMMIT=$(git rev-list -1 "$1"^2)
TITLE=$(git log -1 "$COMMIT" --pretty=format:%s | sed s/'Merge #\([0-9]*\)'/'[upstream PR #\1]'/)
PRNUM=$(git log -1 "$COMMIT" --pretty=format:%s | sed s/'Merge #\([0-9]*\).*'/'\1'/)
OLDCOMMIT=$(git rev-list -1 HEAD)

echo $PRNUM

git fetch upstream &&
git rebase upstream/secp256k1-zkp secp256k1-zkp &&
git checkout -b temp-merge-$PRNUM &&
#git reset --hard "$PRCOMMIT" &&
git merge --no-edit $1 &&
git push origin temp-merge-$PRNUM &&
hub pull-request -b ElementsProject:secp256k1-zkp -h apoelstra:temp-merge-$PRNUM -m "$TITLE"

git checkout secp256k1-zkp
git branch -D temp-merge-$PRNUM
```
I replaced the `reset --hard` line with a `merge` line and also replaced the `PRCOMMIT` variable with a `PRNUM` variable which i think is easier to read.

ysangkok:
I think every new shell script may as well use `set -euo pipefail`. Then you wouldn't be so confused if an environment variable was misspelled, and you wouldn't need all the `&&`.

PR #98, opened 2020-09-27
Add contrib/sync-upstream.sh script to automate syncing PRs

jonasnick:
Based on @apoelstra's original script but adapted to work with github-cli instead of "hub".

### DIFF
--- a/contrib/sync-upstream.sh
+++ b/contrib/sync-upstream.sh
@@ -1,0 +1,111 @@
+#!/bin/bash
+
+set -eou pipefail
+
+help() {
+    echo "$0 range [end]"
+    echo "    merges every merge commit missing merge commit."
+    echo "    If the optional [end] commit is provided, only merges up to [end]."
+    echo
+    echo "$0 select <commit> ... <commit>"
+    echo "    merges every selected merge commit"
+    echo
+    echo "This tool creates a branch and a script that can be executed to create the"
+    echo "PR automatically. The script requires the github-cli tool (aka gh)."
+    echo ""
+    echo "Tip: \`git log --oneline upstream/master --merges\` shows merge commits."
+    exit 1
+}
+
+if [ "$#" -lt 1 ]; then
+    help
+fi
+
+REMOTE=upstream
+REMOTE_BRANCH=$REMOTE/master
+# Makes sure you have a remote "upstream" that is up-to-date
+setup() {
+    ret=0
+    git fetch $REMOTE &> /dev/null || ret=$?
+    if [ ${ret} == 0 ]; then
+        return
+    fi
+    echo "Adding remote \"$REMOTE\" with URL git@github.com:bitcoin-core/secp256k1.git. Continue with y"
+    read -r yn
+    case $yn in
+        [Yy]* ) ;;
+        * ) exit 1;;
+    esac
+    git remote add $REMOTE git@github.com:bitcoin-core/secp256k1.git &> /dev/null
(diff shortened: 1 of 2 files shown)


contrib/sync-upstream.sh
```diff
@@ -0,0 +1,40 @@
+#!/bin/sh
+
+# - $@ are the upstream merge commits (use `git log --oneline master --merges` to show merges)
+# - Requires github-cli tool (a.k.a gh)
+# - Assumes you have an up-to-date (recently fetched) remote bitcoin-core/secp256k1
```
apoelstra, line 5: Would be good to echo this comment (or something like it) if the script is called without any arguments

jonasnick: Added help text if called with the wrong arguments.

apoelstra:
I think this actually doesn't have the "range of PRs" logic in it?

I also think a simpler API might be to just give an upstream commit ID, the script checks that it's from upstream master, and then infers the list of PRs you want to pull in.

jonasnick:
Oh sorry, that's because I forgot to push. Fixed now.

Yeah, if you could only give it one commit and the script would infer the list of PRs that would be very nice. Not sure what's the best way to determine reliably if secp256k1-zkp already includes a PR.

apoelstra:
As long as we are merging every PR, `git merge-base bitcoin/master secp256k1-zkp` will reliably give us the commit of the most recent PR that we merged.

jonasnick:
Cool, that works! I'll update the script to do that.

contrib/sync-upstream.sh
```diff
@@ -0,0 +1,94 @@
...
+set -eou pipefail
+
+help() {
+    echo "$0 range <branch> <startcommit> [end]"
+    echo "    merges every merge commit in the range from the <start> merge commit (excluding) to the optional [end] commit (including)."
+    echo
+    echo "$0 select <commit> ... <commit>"
+    echo "    merges every selected merge commit"
+    echo
+    echo "Note that this tool requires the github-cli tool (aka gh)."
```
real-or-random, line 12: Maybe say that this is only to create the final PR (and the user could do this manually).

Also it would be good to explain that invoking this script will only create a branch with name foo and not already create a PR. (To make sure people know they can play around with it locally.)

jonasnick: Fixed

jonasnick:
The range command doesn't require a start commit anymore because now it'll be automatically determined with `git merge-base`. In order to do that, the script now assumes that there's an `upstream` remote, and if there isn't, it'll try to create one. The script also makes sure that all the relevant remotes are up-to-date.

Not work-in-progress anymore.

real-or-random:
ACK 49856f2daf97e10033fe9d5e878c9fa20efe2b3c The code looks fine. I haven't tested this script but it's not the end of the world if the script has a bug.

real-or-random (approved):
ACK 8295d64396a8560453180b87851f37a3113a72d2 The code looks fine. I haven't tested this script but it's not the end of the world if the script has a bug.

jonasnick:
squashed

real-or-random:
ACK 7eeacd7725fa8c895c2f58850b151e66199137cf  The code looks fine. I haven't tested this script but it's not the end of the world if the script has a bug.

 https://github.com/ElementsProject/secp256k1-zkp/compare/7eeacd7725fa8c895c2f58850b151e66199137cf..8295d64396a8560453180b87851f37a3113a72d2

PR #99, opened 2020-09-28

apoelstra:


jonasnick (approved):
ACK 1789183cbab6062e1ed0a12e04b6362e0531d00f

PR #101, opened 2020-09-28
Upstream PRs  #772 #779 #778

jonasnick:
[upstream PR #772]: Improve constant-timeness on PowerPC
[upstream PR #779]: travis: Fix argument quoting for ./configure
[upstream PR #778]: secp256k1_gej_double_nonzero supports infinity

Created using a new version of #98. Could have done more than 3 commits but wanted to show how this looks like first. By the way `git rerere` is really useful for doing these kind of merges because it remembers how you resolved merge conflicts.

### DIFF
--- a/contrib/travis.sh
+++ b/contrib/travis.sh
@@ -3,10 +3,6 @@
 set -e
 set -x
 
-if [ -n "$HOST" ]
-then
-    export USE_HOST="--host=$HOST"
-fi
 if [ "$HOST" = "i686-linux-gnu" ]
 then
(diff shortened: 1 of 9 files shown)


real-or-random:
The history graphs looks good to me, I think that's nice, then we can batch some of these until and catch up quicker.

apoelstra (approved):
ACK. Was able to reproduce exactly.

Agreed, it's good to batch, though I think we should only batch "easy" PRs which have small/easy conflicts, and do the complicated ones individually.

PR #105, opened 2020-11-13
MuSig state machine simplifictions, API improvements and taproot tweaking

jonasnick:
This PR is mainly a rebase of #87. It also adds taproot tweaking which was part of #86 but then got lost. The PR has a few changes that are seem unnecessary if we upgrade to MuSig2 soon, but overall I think they'll make that switch easier.

### DIFF
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -142,7 +142,7 @@ typedef struct {
  *   In:     pubkeys: input array of public keys to combine. The order is important;
  *                    a different order will result in a different combined public
  *                    key (cannot be NULL)
- *         n_pubkeys: length of pubkeys array
+ *         n_pubkeys: length of pubkeys array. Must be greater than 0.
  */
 SECP256K1_API int secp256k1_musig_pubkey_combine(
     const secp256k1_context* ctx,
@@ -176,7 +176,8 @@ SECP256K1_API int secp256k1_musig_pubkey_combine(
  *                     `musig_pubkey_combine` (cannot be NULL)
  *          n_signers: length of signers array. Number of signers participating in
  *                     the MuSig. Must be greater than 0 and at most 2^32 - 1.
- *           my_index: index of this signer in the signers array
+ *           my_index: index of this signer in the signers array. Must be less
+ *                     than `n_signers`.
  *             seckey: the signer's 32-byte secret key (cannot be NULL)
  */
 SECP256K1_API int secp256k1_musig_session_initialize(
@@ -193,7 +194,10 @@ SECP256K1_API int secp256k1_musig_session_initialize(
     const unsigned char *seckey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5) SECP256K1_ARG_NONNULL(7) SECP256K1_ARG_NONNULL(8) SECP256K1_ARG_NONNULL(11);
 
-/** Gets the signer's public nonce given a list of all signers' data with commitments
+/** Gets the signer's public nonce given a list of all signers' data with
+ *  commitments.  Called by participating signers after
+ *  `secp256k1_musig_session_initialize` and after all nonce commitments have
+ *  been collected
  *
  *  Returns: 1: public nonce is written in nonce
  *           0: signer data is missing commitments or session isn't initialized
@@ -204,7 +208,7 @@ SECP256K1_API int secp256k1_musig_session_initialize(
  *                    `musig_session_initialize`. Array length must equal to
  *                    `n_commitments` (cannot be NULL)
  *  Out:       nonce: the nonce (cannot be NULL)
- *  In:  commitments: array of 32-byte nonce commitments (cannot be NULL)
+ *  In:  commitments: array of pointers to 32-byte nonce commitments (cannot be NULL)
  *     n_commitments: the length of commitments and signers array. Must be the total
  *                    number of signers participating in the MuSig.
  *             msg32: the 32-byte message to be signed. Must be NULL if already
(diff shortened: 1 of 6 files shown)


include/secp256k1_musig.h
```diff
@@ -154,6 +161,41 @@ SECP256K1_API int secp256k1_musig_pubkey_combine(
     size_t n_pubkeys
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5);
 
+/** Tweak an x-only public key by adding the generator multiplied with tweak32
+ *  to it. The resulting output_pubkey with the given internal_pubkey and tweak
+ *  passes `secp256k1_xonly_pubkey_tweak_test`.
+ *
+ *  This function is only useful before initializing a signing session.
+ *  Otherwise just use `secp256k1_xonly_pubkey_tweak_add`. Can only be called
```
apoelstra, line 169: Rather than "Otherwise just use" you could say "If you are only computing a public key, but not yet signing with it, you can just use".

As written, I wasn't clear on where in the workflow I was supposed to use each function (though it did become clear after reading a few more doccomments)

src/modules/musig/tests_impl.h
```diff
@@ -170,6 +170,42 @@ void musig_api_tests(secp256k1_scratch_space *scratch) {
     CHECK(secp256k1_musig_pubkey_combine(vrfy, scratch, &combined_pk, &pre_session, pk, 2) == 1);
     CHECK(secp256k1_musig_pubkey_combine(vrfy, scratch, &combined_pk, &pre_session, pk, 2) == 1);
 
+    /** Twerking */
```
apoelstra, line 173: :peach:

src/modules/musig/main_impl.h
```diff
@@ -181,22 +207,33 @@ int secp256k1_musig_session_init(const secp256k1_context* ctx, secp256k1_musig_s
         return 0;
     }
     secp256k1_musig_coefficient(&mu, session->pre_session.pk_hash, (uint32_t) my_index);
-    /* Compute the signers public key point and determine if the secret needs to
-     * be negated before signing. If the signer's pubkey has an odd Y coordinate
-     * XOR the MuSig-combined pubkey has an odd Y coordinate, the secret has to
-     * be negated. This can be seen by looking at the secret key belonging to
-     * `combined_pk`. Let's define
+    /* Compute the signers public key point and determine if the secret is
```
apoelstra, line 210: s/signers/signer's/

src/modules/musig/main_impl.h
```diff
@@ -502,6 +539,26 @@ int secp256k1_musig_partial_sig_combine(const secp256k1_context* ctx, const secp
         secp256k1_scalar_add(&s, &s, &term);
     }
 
+    /* If there is a tweak then add (or subtract) `msghash` times `tweak` to `s`.*/
+    if (session->pre_session.is_tweaked) {
+        unsigned char msghash[32];
+        secp256k1_scalar e, scalar_tweak;
```
apoelstra, line 545: I don't mind, but I think everywhere else in this library we avoid defining more than one variable on a line. This is a MISRA thing I think.

src/modules/musig/main_impl.h
```diff
@@ -548,10 +605,16 @@ int secp256k1_musig_partial_sig_verify(const secp256k1_context* ctx, const secp2
...
      * sign for the negation of their individual xonly public key such that the
-     * combined signature is valid for the MuSig aggregated xonly key. */
-    if (session->pre_session.pk_parity) {
+     * combined signature is valid for the MuSig aggregated xonly key. If the
+     * MuSig-combined point was tweaked then `e` is negated if the combined key
+     * has an odd Y coordinate XOR the internal key has an odd Y coordinate.*/
+    if (session->pre_session.pk_parity
+            + (session->pre_session.is_tweaked
+                && session->pre_session.internal_key_parity)
+            % 2 == 1) {
```
apoelstra, line 617: This made me laugh but it is correct.

It could maybe be more clearly written
```
if (session->pre_session.pk_parity != (session->pre_session.is_tweaked && session->pre_session.internal_key_parity))
```

apoelstra: Oh, I meant to comment on the earlier instance of `% 2 == 1`, where it makes more sense that you did this. Then this one makes sense for consistency's sake.

apoelstra (approved):
ack 7b4e7a46087000a95d542847c739e7019fe87c89

These all look great! I only have nits about comments and formatting.

src/modules/musig/tests_impl.h
real-or-random: Reminds me of https://github.com/ElementsProject/secp256k1-zkp/blob/fabc8f74e7e4d624d4d2d1aba621b8a75c729e5e/src/ecdsa_impl.h#L164 
 :wrench: :blue_car:  
I didn't touch this so far because it's too funny to fix.

src/modules/musig/main_impl.h
real-or-random: `!=` is certainly more idiomatic

src/modules/musig/main_impl.h
real-or-random: We don't have that convention, see
 https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/ecdsa_impl.h#L185
https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/ecdsa_impl.h#L209
for two random examples

apoelstra: Oh, ok, I guess I just imagined this and have been adhering to it for no reason.

Carry on.

include/secp256k1_musig.h

src/modules/musig/tests_impl.h
jonasnick: Was wondering if you'd find this :) Joke's over. I removed it.

src/modules/musig/main_impl.h

src/modules/musig/main_impl.h
jonasnick: even with consistency in mind `% 2 == 1` looks crappy. Switched to the `!=` version.

apoelstra (approved):
ack 3fb4d6db9ccb0546691e51c9a6e3d2e3fa7ba086

PR #106, opened 2020-11-16
reduce test rounds for rangeproof and surjectionproof

apoelstra:
This, along with setting the test count to 32 rather than 64, reduces the default test run from 1m40 on my laptop to 30s.

### DIFF
--- a/src/modules/rangeproof/tests_impl.h
+++ b/src/modules/rangeproof/tests_impl.h
@@ -498,7 +498,7 @@ static void test_rangeproof(void) {
         CHECK(maxv >= v);
     }
     memcpy(&commit2, &commit, sizeof(commit));
-    for (i = 0; i < (size_t) 2*count; i++) {
+    for (i = 0; i < (size_t) count; i++) {
         int exp;
         int min_bits;
         v = secp256k1_testrandi64(0, UINT64_MAX >> (secp256k1_testrand32()&63));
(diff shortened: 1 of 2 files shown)


apoelstra:
It appears [elements has disabled their unit tests in some builds](https://github.com/ElementsProject/elements/blob/88067ddbe4223e43a4c86912928b28b834bbf69e/.travis.yml#L110) due to our unit tests taking forever. I agree, I hate running them on this project. Most of the extra time is taken on the rangeproof and surjectionproof tests, which do a ton of iterations, and anyway this code has not changed in years.

Also reduces the global default count from 64 to 32, since we have a lot more modules than upstream does, and that's only going to grow.

real-or-random (approved):
ACK I inspected the diff, I didn't run the tests

jonasnick (commented):
Reducing the iterations in the rangeproof and surjectionproof tests is fine with me, but regarding the global test `count`, it's configurable by the caller, so elements can already use whatever they want - without changing the default (and fwiw 1.30 min shouldn't really matter in CI).

apoelstra:
elements call our own `make check` which does not expose a way to change the iteration count.

jonasnick:
We can easily add a way to change the iteration count with an environment variable, similar to how we use the `SECP256K1_BENCH_ITERS` variable to set iterations for benchmarks. It'd be a better choice conceptually, especially since `count=64` isn't a problem for -zkp CI (yet). But if you need this PR quickly it's also fine with me.

apoelstra:
No, there is definitely no time pressure. This is part of a post-Elements-rebase fixup of Elements CI

apoelstra:
I think we should PR upstream to use an env var.

apoelstra:
Removed the global round reduction. We should sync up with upstream now that they have an env var.

jonasnick (commented):
ACK 29f9a7dc62f5158c30366efbce073680314851dc

PR #110, opened 2020-11-25
Expose Borromean ring signature and de-anonymization functionality

### LINKED ISSUE #109: Add feature for the real signer to later reveal himself and to verify such claims with 100% probability
This work is epic, thanks to everyone involved.

In order to wide up its application areas, it would be very much useful to add functionality so that a signer in a group can later identify himself as the real signer with 0% false positive rate.

This opens the door for many use cases, for example assigning data to public ring signatures that is later granted to the signer that reveals himself as the real singer. This requires:
- functionality to create such a "link-able" ring-signature (with a special flag in the command of the request maybe, if the default is "unlikable")
- functionality to verify the claim of a signer that later reveals himself

From the old How to leak a secret original paper:
_"However, there may be cases in which the signer himself wants to have the option of later proving his authorship of the anonymized email (e.g., if he is successful in toppling the disgraced Prime Minister). Yet another possibility is that the signer A wants to initially use {A,B,C} as the list of possible signers, but later prove that C is not the real signer. There is a simple way to implement these options, by choosing the xi values for the nonsigners in a pseudorandom rather than truly random way. To show that C is not the author, A publishes the seed which pseudorandomly generated the part of the signature associated with C. To prove that A is the signer, A can reveal a single seed which was used to generate all the nonsigners’ parts of the signature.The signer A cannot misuse this technique to prove that he is not the signer since his part is computed rather than generated, and is extremely unlikely to have a corresponding seed. Note that these modified versions can guarantee only computational anonymity, since a powerful adversary can search for such proofs of non authorship and use them to expose the signer."_

apoelstra:
FIxes #109

### DIFF
--- a/include/secp256k1_rangeproof.h
+++ b/include/secp256k1_rangeproof.h
@@ -284,6 +284,99 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_rangeproof_info(
   size_t plen
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
+/** Returns the serialized size of a ring signature with the given number of keys
+ *  Args:   ctx: pointer to a context object (cannot be NULL)
+ *  In:  n_keys: number of keys
+ */
+SECP256K1_API size_t secp256k1_ring_signature_length(
+    const secp256k1_context* ctx,
+    size_t n_keys
+) SECP256K1_ARG_NONNULL(1);
+
+/** Returns the minimum scratch space size to sign or verify a ring signature with the given number of keys
+ *  Args:   ctx: pointer to a context object (cannot be NULL)
+ *  In:  n_keys: number of keys
+ */
+SECP256K1_API size_t secp256k1_ring_signature_scratch_space_size(
+    const secp256k1_context* ctx,
+    size_t n_keys
+) SECP256K1_ARG_NONNULL(1);
+
+/** Produces a linear-sized ring signature with a given set of keys, returning a "proof" which can be used to reveal the signer
+ *  Returns 1: Proof successfully created.
+ *          0: if sig_len was less than the `secp256k1_ring_signature_length` for this number of keys
+ *  Args:          ctx: pointer to a context object, initialized for both signing and verification (cannot be NULL)
+ *             scratch: scratch space with enough space available (to learn a minimum size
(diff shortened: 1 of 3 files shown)


jonasnick (commented):
Fwiw, this API worked just fine for [taproot-ringsig](https://github.com/jonasnick/taproot-ringsig).

apoelstra:
Nice :).

One design point though -- since we have to nail down the stream cipher to make the anonymity revocation work in the future, maybe we should switch to chacha rather than sha2_hmac? It'd be much much faster.

apoelstra:
Also worth considering whether we should expose an encyption API for this too.

real-or-random:
> Nice :).
> 
> One design point though -- since we have to nail down the stream cipher to make the anonymity revocation work in the future, maybe we should switch to chacha rather than sha2_hmac? It'd be much much faster.

I haven't looked at the PR so far but  we need to think about the properties we require from the function that maps a seed to the coefficients. The function still needs to be a PRG simply because we need random coefficients.

I think for security against "false claims" of not being the signer, we need only preimage resistance and AFAIU, the text quoted in #109 is essentially arguing that any function `{0,1}^s -> {0,1}^n` from a short seed of `s` bits to a much longer output of `n` bits is statistically preimage-resistant: Given a random element of `{0,1}^n`, it a preimage with probability at most `2^(s-n)`.


> Also worth considering whether we should expose an encyption API for this too.

Encryption API for what exactly?

apoelstra:
I didn't realize that chacha was biased and could not be used as a PRG. Will need to address this in the Bulletproofs PR as well.

The encryption API lets you hide data inside the rangeproof by xoring it with the PRG output.

real-or-random:
> I didn't realize that chacha was biased and could not be used as a PRG. Will need to address this in the Bulletproofs PR as well.

Wait, who said that ChaCha is biased?

> 
> The encryption API lets you hide data inside the rangeproof by xoring it with the PRG output.

Ah yes, that will be neat.

apoelstra:
>Wait, who said that ChaCha is biased?

I am quoting https://tools.ietf.org/html/rfc8439#page-20 which says

> Additionally, unlike HMAC, Poly1305 is biased, so using it for key derivation would reduce the security of the symmetric encryption.

apoelstra:
Oh, I'm an idiot, it says in the next sentence that chacha20 would be fine but that is not what some particular use case needs, so they don't specifiy it in the RFC.

How could something be a stream cipher but not a PRG?

jonasnick:
Sorry this was automaticaly closed. PR needs to be reopened against the `master` branch.

apoelstra:
All good. I will reopen. Looks like github won't let me retarget the same PR so we'll lose the comments, but given that they're mostly me being confused about chacha, that's not a big loss.

PR #111, opened 2020-11-26
Add ECDSA sign-to-contract module

apoelstra:
This is a backport and rebase of https://github.com/bitcoin-core/secp256k1/pull/669


src/secp256k1.c
```diff
@@ -769,6 +802,142 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+
+    secp256k1_sha256_initialize(&sha);
+    secp256k1_sha256_write(&sha, rbuf, rbuf_size);
+    secp256k1_sha256_write(&sha, data, data_size);
+    secp256k1_sha256_finalize(&sha, tweak32);
+    return 1;
+}
+
+/* Compute an ec commitment as pubkey + hash(pubkey, data)*G. */
+static int secp256k1_ec_commit(const secp256k1_context* ctx, secp256k1_pubkey *commitment, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
```
jonasnick, line 829: I like that we finally add that function but I didn't know what I was doing when I wrote it. Since this is an internal function it should operate on group elements and ec_commit_seckey should operate on scalars. This will get rid of a lot of `pubkey_save/load` calls and I think we can make the functions return `void` which simplifies sign-to-contract signing.

src/secp256k1.c
```diff
@@ -769,6 +802,142 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+    secp256k1_pubkey pubkey_tmp;
+
+    if (pubkey == NULL) {
+        /* Compute pubkey from seckey if not provided */
+        int overflow;
+        secp256k1_scalar x;
+        secp256k1_gej pj;
+        secp256k1_ge p;
+
+        secp256k1_scalar_set_b32(&x, seckey, &overflow);
```
jonasnick, line 852: should use `scalar_set_b32_seckey` here

src/secp256k1.c
```diff
@@ -769,6 +802,142 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
     return 1;
 }
 
+/* Compute an ec commitment tweak as hash(pubkey, data). */
+static int secp256k1_ec_commit_tweak(const secp256k1_context *ctx, unsigned char *tweak32, const secp256k1_pubkey *pubkey, const unsigned char *data, size_t data_size) {
```
jonasnick, line 806: How about we let this take an initialized sha object. That way we can easily add domain separation via tagged hashes. And ideally we'd add a tagged hash for "ECDSA sign-to-contract" or something.

.travis.yml
```diff
@@ -17,15 +17,15 @@ compiler:
   - gcc
 env:
   global:
-    - WIDEMUL=auto  BIGNUM=auto  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  WITH_VALGRIND=yes RUN_VALGRIND=no EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no SCHNORRSIG=no EXPERIMENTAL=no CTIMETEST=yes BENCH=yes ITERS=2 GENERATOR=no RANGEPROOF=no WHITELIST=no SCHNORRSIG=no MUSIG=no
+    - WIDEMUL=auto  BIGNUM=auto  STATICPRECOMPUTATION=yes  ECMULTGENPRECISION=auto  ASM=no  BUILD=check  WITH_VALGRIND=yes RUN_VALGRIND=no EXTRAFLAGS=  HOST=  ECDH=no  RECOVERY=no SCHNORRSIG=no ECDSA_SIGN_TO_CONTRACT=no EXPERIMENTAL=no CTIMETEST=yes BENCH=yes ITERS=2 GENERATOR=no RANGEPROOF=no WHITELIST=no SCHNORRSIG=no MUSIG=no
```
jonasnick, line 20: How about we rename `sign to contract` to `s2c` everywhere (including file names) except docs?

include/secp256k1.h
```diff
@@ -81,6 +82,29 @@ typedef struct {
...
+    /* magic is set during initialization */
+    uint64_t magic;
+    /* Public nonce before applying the sign-to-contract commitment */
+    secp256k1_pubkey original_pubnonce;
+    /* Byte indicating if signing algorithm negated the nonce. Alternatively when
+     * verifying we could compute the EC commitment of original_pubnonce and the
+     * data and negate if this would not be a valid nonce. But this would prevent
+     * batch verification of sign-to-contract commitments. */
+    int nonce_is_negated;
+} secp256k1_s2c_opening;
```
jonasnick, line 106: My initial idea was that schnorr and ecdsa could reuse the same s2c structure, but I think it would simplify things quite a bit if they wouldn't. ECDSA doesn't make use of the `nonce_is_negated` flag. So we could make the object opaque and identical to pubkeys, delegate serialize and parse to pubkey ser and parse and move the opening struct out of the secp256k1 include and into secp256k1_s2c.h

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,161 @@
...
+     * security as it is easier to find a collision with a point and its negation.
+     */
+    secp256k1_ecdsa_signature_load(ctx, &sigr, &sigs, sig);
+
+    if (!secp256k1_pubkey_load(ctx, &commitment_ge, &commitment)) {
+        return 0;
+    }
+    secp256k1_fe_normalize(&commitment_ge.x);
+    secp256k1_fe_get_b32(x_bytes, & commitment_ge.x);
+    secp256k1_scalar_set_b32(&x_scalar, x_bytes, NULL);
```
jonasnick, line 75: Perhaps add a comment that the signature wouldn't verify if this would overflow.

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,142 @@
...
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:              ctx: pointer to a context object (cannot be NULL)
+ *  Out: rand_commitment32: pointer to 32-byte array to store the returned commitment (cannot be NULL)
+ *  In:             rand32: the 32-byte randomness to commit to (cannot be NULL). It must come from
+ *                          a cryptographically secure RNG. As per the protocol, this value must not
+ *                          be revealed to the client until after the host has received the client
+ *                          commitment.
+ */
+SECP256K1_API int secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit(
```
jonasnick, line 61: This should be below the general ECDSA ANCCP doc.

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,161 @@
...
+    return secp256k1_scalar_eq(&sigr, &x_scalar);
+}
+int secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit(secp256k1_context *ctx, unsigned char *rand_commitment32, const unsigned char *rand32) {
```
jonasnick, line 78: This is already added in the sign-to-contract commit but should be added one commit later.

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,142 @@
...
+    secp256k1_context *ctx,
+    unsigned char *rand_commitment32,
+    const unsigned char *rand32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Compute commitment on the client as part of the ECDSA Anti Nonce Covert Channel Protocol.
```
jonasnick, line 67: As we discussed, we may want to call this ECDSA Anti Kleptography Protocol and correspondingly replace `ecdsa_s2c_anti_nonce_covert_channel` with `ecdsa_antiklepto`.

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,161 @@
...
+    while (1) {
+        int overflow = 0;
+        if (!secp256k1_nonce_function_default(nonce32, msg32, seckey32, NULL, rand_commitment32, count)) {
+            /* cannot happen with secp256k1_nonce_function_default */
+            memset(nonce32, 0, 32);
+            secp256k1_scalar_clear(&k);
+            return 0;
+        }
+
+        secp256k1_scalar_set_b32(&k, nonce32, &overflow);
```
jonasnick, line 113: This should be `scalar_set_b32_seckey`.

src/modules/ecdsa_sign_to_contract/tests_impl.h
```diff
@@ -0,0 +1,352 @@
...
+
+    /* Protocol step 1. */
+    CHECK(secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit(ctx, host_commitment, host_nonce_contribution) == 1);
+    /* Protocol step 2. */
+    CHECK(secp256k1_ecdsa_s2c_anti_nonce_covert_channel_client_commit(ctx, &client_commitment, host_msg, client_privkey, host_commitment) == 1);
+    /* Protocol step 3: host_nonce_contribution send to client to be used in step 4. */
+    /* Protocol step 4. */
+    CHECK(secp256k1_ecdsa_s2c_sign(ctx, &signature, &s2c_opening, host_msg, client_privkey, host_nonce_contribution) == 1);
+    /* Protocol step 5. */
+    CHECK(secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_verify(ctx, &signature, host_nonce_contribution, &s2c_opening, &client_commitment) == 1);
```
jonasnick, line 318: we should also verify the signature here.

src/modules/ecdsa_sign_to_contract/tests_impl.h
```diff
@@ -0,0 +1,352 @@
...
+            /* revert */
+            sigbytes[i] = (((int)sigbytes[i]) + 255) % 256;
+        }
+    }
+    { /* host_verify: client commitment != opening original pubnonce */
+
+        unsigned char tweak[32] = {1};
+        CHECK(secp256k1_ec_pubkey_tweak_add(ctx, &client_commitment, tweak) == 1);
+        CHECK(secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_verify(ctx, &signature, host_nonce_contribution, &s2c_opening, &client_commitment) == 0);
+    }
```
jonasnick, line 338: Can we add a test case that if the host sends randomness that doesn't match its commitment the result is that the original pubnonce after signing does not match the client commitment (i.e. the signer doesn't sign twice with the same nonce)?

jonasnick (commented):
Looks good in general. I think we can simplify a few things as mentioned in the comments. Also, we should  add `ecdsa_sign_s2c` and `ecdsa_s2c_anti_nonce_covert_channel_client_commit` to the ctime_tests.

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: I'm going to leave this one alone so that I don't need to edit existing commits.

apoelstra: Unless you think I should recreate everything since I'm doing to do a bunch of renames etc.

include/secp256k1.h
real-or-random: This sounds like a reasonable plan .

src/tests.c
```diff
@@ -4595,6 +4595,58 @@ void run_eckey_negate_test(void) {
...
+            0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00,
+            0x02
+    };
+    secp256k1_s2c_opening opening;
+    size_t ecount = 0;
+
+    secp256k1_context_set_illegal_callback(ctx, counting_illegal_callback_fn, &ecount);
+
+    /* Uninitialized opening can't be serialized. Actually testing that would be
+     * undefined behavior. Therefore we simulate it by setting the opening to 0. */
```
real-or-random, line 4616: nit: I'd just say "Opening with all bytes set to zero can't be serialized." 

(Whether the uninitialized opening can be serialized is UB, so this is indeed nothing we can test.)

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,55 @@
+#ifndef SECP256K1_ECDSA_SIGN_TO_CONTRACT_H
+#define SECP256K1_ECDSA_SIGN_TO_CONTRACT_H
+
+#include "secp256k1.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Same as secp256k1_ecdsa_sign, but s2c_data32 is committed to by adding `hash(k*G, s2c_data32)` to
```
real-or-random, line 10: nit: formally, "G" is not specified here... Feel free to ignore.

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,55 @@
...
+ *  the signing nonce `k`.
+ *  Returns: 1: signature created
+ *           0: the nonce generation function failed, or the private key was invalid.
+ *  Args:    ctx:  pointer to a context object, initialized for signing (cannot be NULL)
+ *  Out:     sig:  pointer to an array where the signature will be placed (cannot be NULL)
+ *   s2c_opening:  pointer to an secp256k1_s2c_opening structure which can be
+ *                 NULL but is required to be not NULL if this signature creates
+ *                 a sign-to-contract commitment (i.e. the `s2c_data` argument
```
real-or-random, line 18: s/i.e./i.e. if

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,55 @@
...
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** Same as secp256k1_ecdsa_sign, but s2c_data32 is committed to by adding `hash(k*G, s2c_data32)` to
+ *  the signing nonce `k`.
+ *  Returns: 1: signature created
+ *           0: the nonce generation function failed, or the private key was invalid.
+ *  Args:    ctx:  pointer to a context object, initialized for signing (cannot be NULL)
+ *  Out:     sig:  pointer to an array where the signature will be placed (cannot be NULL)
```
real-or-random, line 15: nit: s/structure/object
(C terms, we use "object" in other places in the API docs, I think)

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,55 @@
...
+ *   s2c_opening:  pointer to an secp256k1_s2c_opening structure which can be
+ *                 NULL but is required to be not NULL if this signature creates
+ *                 a sign-to-contract commitment (i.e. the `s2c_data` argument
+ *                 is not NULL).
+ *  In:
+ *         msg32: the 32-byte message hash being signed (cannot be NULL)
+ *        seckey: pointer to a 32-byte secret key (cannot be NULL)
+ *    s2c_data32: pointer to a 32-byte data to create an optional
+ *                sign-to-contract commitment to if not NULL (can be NULL).
```
real-or-random, line 24: strange grammar?

include/secp256k1_ecdsa_sign_to_contract.h
```diff
@@ -0,0 +1,55 @@
...
+    const unsigned char* s2c_data32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Verify a sign-to-contract commitment.
+ *
+ *  Returns: 1: the signature contains a commitment to data32
+ *           0: incorrect opening
+ *  Args:    ctx: a secp256k1 context object, initialized for verification.
+ *  In:      sig: the signature containing the sign-to-contract commitment (cannot be NULL)
+ *        data32: the 32-byte data that was committed to (cannot be NULL)
```
real-or-random, line 41: nit: maybe add "purportedly" or similar

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,93 @@
...
+    unsigned char ndata[32];
+    const unsigned char* noncedata = NULL;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(secp256k1_ecmult_gen_context_is_built(&ctx->ecmult_gen_ctx));
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(signature != NULL);
+    ARG_CHECK(seckey != NULL);
+
+    if(s2c_data32 != NULL) {
```
real-or-random, line 24: nit:spacing

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,93 @@
...
+        /* Provide s2c_data32 to the the nonce function as additional data to derive the nonce
```
real-or-random, line 25: s/the the/the

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,93 @@
...
+         * from. It's hashed because it should be possible to derive nonces even if only a SHA256
+         * commitment to the data is known.  This is for example important in the
+         * anti-nonce-covert-channel protocol.
+         */
+        secp256k1_sha256 sha;
```
real-or-random, line 30: This should be a tagged hash.

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,93 @@
...
+     */
+    secp256k1_ecdsa_signature_load(ctx, &sigr, &sigs, sig);
+
+    if (!secp256k1_pubkey_load(ctx, &commitment_ge, &commitment)) {
+        return 0;
+    }
+    secp256k1_fe_normalize(&commitment_ge.x);
+    secp256k1_fe_get_b32(x_bytes1, &commitment_ge.x);
+    secp256k1_scalar_get_b32(x_bytes2, &sigr);
+    return memcmp(x_bytes1, x_bytes2, 32) == 0;
```
real-or-random: secp_memcmp

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -0,0 +1,93 @@
...
+
+}
+int secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit(secp256k1_context *ctx, unsigned char *rand_commitment32, const unsigned char *rand32) {
+    secp256k1_sha256 sha;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(rand_commitment32 != NULL);
+    ARG_CHECK(rand32 != NULL);
+
+    secp256k1_sha256_initialize(&sha);
```
real-or-random, line 85: tagged hash

src/modules/ecdsa_sign_to_contract/main_impl.h
```diff
@@ -90,4 +90,74 @@ int secp256k1_ecdsa_s2c_anti_nonce_covert_channel_host_commit(secp256k1_context
...
+    ARG_CHECK(secp256k1_ecmult_gen_context_is_built(&ctx->ecmult_gen_ctx));
+    ARG_CHECK(client_commit != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(seckey32 != NULL);
+    ARG_CHECK(rand_commitment32 != NULL);
+
+    while (1) {
+        int overflow = 0;
+        if (!secp256k1_nonce_function_default(nonce32, msg32, seckey32, NULL, rand_commitment32, count)) {
+            /* cannot happen with secp256k1_nonce_function_default */
```
real-or-random, line 107: we could also call the error callback here,  not sure if it's better.

src/modules/ecdsa_sign_to_contract/tests_impl.h
```diff
@@ -201,12 +264,88 @@ static void test_ecdsa_s2c_sign_verify(void) {
     }
 }
 
+static void test_ecdsa_s2c_anti_nonce_covert_channel_client_commit(void) {
+    size_t i;
+    unsigned char privkey[32] = {
+        0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
+        0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55, 0x55,
```
real-or-random, line 271: TIL that trailing commas are allowed here. :D

real-or-random (commented):
Some of these functions need to be added to the constant-time tests.

src/secp256k1.c
```diff
@@ -769,6 +769,89 @@ int secp256k1_ec_pubkey_combine(const secp256k1_context* ctx, secp256k1_pubkey *
...
+    secp256k1_ge p;
+    unsigned char rbuf[33];
+    size_t rbuf_size = sizeof(rbuf);
+    secp256k1_sha256 sha;
+
+    if (data_size == 0) {
+        /* That's probably not what the caller wanted */
+        return 0;
```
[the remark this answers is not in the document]
jonasnick: + we can make this function void

src/modules/ecdsa_sign_to_contract/main_impl.h
jonasnick: it's replaced in a later commit :)

real-or-random: whoops, I missed that.

src/modules/ecdsa_sign_to_contract/main_impl.h
jonasnick: would need to be same tagged hash as `secdsa_s2c_sign` uses to create `noncedata` from the `s2c_data`.

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: I rebased on latest master without thinking, I guess I might as well fix this too :). Still going to add other changes as followup commits, mainly to preserve authorship.

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: Will be fixed when we change the internal API to use scalars/ges rather than byte arrays and pubkeys.

src/secp256k1.c
apoelstra: Will be fixed when we change the internal API to use scalars/ges rather than byte arrays and pubkeys.

src/modules/ecdsa_sign_to_contract/main_impl.h
real-or-random: No idea if it's the right tool for this specific case but this often helps me when editing commits of other people:

https://docs.github.com/en/free-pro-team@latest/github/committing-changes-to-your-project/creating-a-commit-with-multiple-authors#creating-co-authored-commits-on-the-command-line

include/secp256k1.h
apoelstra: Should we keep the magic field?

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: Ah, ok, I think I'll remake the history and just put Jonas and @benma as co-authors on every commit.

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: I wonder if we should also fail the ec commitment check in this case though.

src/modules/ecdsa_sign_to_contract/main_impl.h
apoelstra: Do we want two separate tagged hashes, one for the data commitment and the other for the P→P+eG commitment?

apoelstra:
Redid the whole API. I think I addressed all comments except to add ctime tests, which I will do in another commit.

src/modules/ecdsa_sign_to_contract/main_impl.h
real-or-random: I think yes. We should separate domains as much as possible.

include/secp256k1.h
jonasnick: No, calling `pubkey_load` is perfectly fine.

src/eccommit_impl.h
```diff
@@ -0,0 +1,78 @@
...
+    secp256k1_scalar_set_b32(&term, tweak, &overflow);
+    return !overflow && secp256k1_eckey_pubkey_tweak_add(ecmult_ctx, pubp, &term);
+}
+
+/* Compute an ec commitment tweak as hash(pubp, data). */
+static void secp256k1_ec_commit_tweak(unsigned char *tweak32, secp256k1_ge* pubp, secp256k1_sha256* sha, const unsigned char *data, size_t data_size)
+{
+    unsigned char rbuf[33];
+
+    /* secp256k1_eckey_pubkey_serialize is not constant-time */
```
jonasnick, line 35: Since in `ecdsa_sign_inner` R is going to be part of the signature anyway, we can declassify it before calling the `ec_commit_seckey` function. Then we can call `eckey_pubkey_serialize` here. I think it's more logical to declassify points when they are public after we created them instead of trying to work around.

src/modules/ecdsa_sign_to_contract/main_impl.h
jonasnick: I don't think this matters much because there are other ways to make `s2c_verify` succeed where `ecdsa_verify` fails - in particular by choosing a correct R and some random s. I didn't imagine actual attacks, but logically it makes most sense to consider `ecdsa_verify` part of sign-to-contract verification. If so, we need to either change `s2c_verify` to also verify the ecdsa signature or we need to clearly mention in the docs of `s2c_verify` (and antiklepto step 5) that one is supposed to also call `ecdsa_verify`.

src/secp256k1.c
```diff
@@ -552,7 +588,9 @@ int secp256k1_ecdsa_sign(const secp256k1_context* ctx, secp256k1_ecdsa_signature
     ARG_CHECK(signature != NULL);
     ARG_CHECK(seckey != NULL);
 
-    ret = secp256k1_ecdsa_sign_inner(ctx, &r, &s, NULL, msg32, seckey, noncefp, noncedata);
+    ret = secp256k1_ecdsa_sign_inner(ctx, &r, &s, NULL, NULL, NULL, NULL, msg32, seckey, noncefp, noncedata);
+    secp256k1_scalar_cmov(&r, &secp256k1_scalar_zero, !ret);
+    secp256k1_scalar_cmov(&s, &secp256k1_scalar_zero, !ret);
```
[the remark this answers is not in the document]
apoelstra: Hmm, it was part of the original PR, I assume as part of a rebase or something, and I didn't think much about it beyond "zeroing out memory, that's good practice, I'll leave it in". But I see that it's actually just zeroing out public data that never even reaches the user..

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,166 @@
...
+ *  2. The signing device computes a public nonce `R` using the host's commitment
+ *     as auxiliary randomness, using `secp256k1_ecdsa_anti_klepto_signer_commit`.
+ *     The signing device sends the resulting `R` to the host as a s2c_opening.
+ *  3. The host replies with `rho` generated in step 1.
+ *  4. The device signs with `secp256k1_ecdsa_s2c_sign`, using `rho` as `s2c_data`,
+ *     and sends the signature to the host. (It can drop the `s2c_opening` that this
+ *     outputs since it will match the public nonce from step 2.)
+ *  5. The host verifies that the signature's public nonce matches the opening from
+ *     step 2 and its original randomness `rho`, using `secp256k1_ecdsa_s2c_verify_commit`.
+ *     If this fails, the protocol has failed and may be restarted with a fresh `rho`.
```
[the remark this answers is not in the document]
apoelstra: Good point. I thought about this and wasn't sure how best to update the document.

src/secp256k1.c
```diff
@@ -633,17 +671,6 @@ int secp256k1_ec_pubkey_negate(const secp256k1_context* ctx, secp256k1_pubkey *p
 }
 
 
-static int secp256k1_ec_seckey_tweak_add_helper(secp256k1_scalar *sec, const unsigned char *tweak) {
```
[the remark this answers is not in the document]
apoelstra: Ah good point. The reason I moved this was just because it wasn't declared yet when I #included eccommit_impl.h. I'll change it to just add a forward declaration to ecccommit_impl.h.

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,166 @@
...
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_s2c_verify_commit(
+    const secp256k1_context* ctx,
+    const secp256k1_ecdsa_signature *sig,
+    const unsigned char *data32,
+    const secp256k1_ecdsa_s2c_opening *opening
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+
+/**** ECDSA Anti-Klepto Protocol ****
```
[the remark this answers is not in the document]
apoelstra: Is there a good way to test doccomments.

I wasn't really sure where this comment should even go, it's not attached to any particular function. I could attach it to `host_commit` similarly to what benma did originally.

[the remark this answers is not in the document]
jonasnick: I think it would be fine to just replace `/****` with `/**`.

src/secp256k1.c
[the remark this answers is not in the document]
apoelstra: Actually these lines appear at the end of `ecdsa_sign_inner` and are completely redundant. I think I may have introduced them as a rebasing mistake and then thought they were deliberate. I removed them.

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,180 @@
...
+ *     as auxiliary randomness, using `secp256k1_ecdsa_anti_klepto_signer_commit`.
+ *     The signing device sends the resulting `R` to the host as a s2c_opening.
+ *  3. The host replies with `rho` generated in step 1.
+ *  4. The device signs with `secp256k1_ecdsa_s2c_sign`, using `rho` as `s2c_data`,
+ *     and sends the signature to the host. (It can drop the `s2c_opening` that this
+ *     outputs since it will match the public nonce from step 2.)
+ *  5. The host verifies that the signature's public nonce matches the opening from
+ *     step 2 and its original randomness `rho`, using `secp256k1_ecdsa_s2c_verify_commit`.
+ *     If this fails, it indicates a potentially serious problem with the hardware device,
+ *     and we strongly discourage its continued use.
```
jonasnick, line 126: Perhaps `s/fails/fails or if the signing device aborts after step 3/` and `s/use./use (see Rationale below)./` With rationale

- If the hardware device is not discarded after wrongly or not at all responding after step 3, the hardware device can successfully exfiltrate the signing key _even if_ the host only lets signatures pass after verifying the commitment. This is because after step 3, the signing device knows the final nonce and can simply abort if it does not match the desired bias.

Hm, I'm really skeptical about implementers getting this right.

include/secp256k1_ecdsa_s2c.h
```diff
@@ -88,13 +96,18 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_s2c_verify_commit
...
 
-/**** ECDSA Anti-Klepto Protocol ****
+/** ECDSA Anti-Klepto Protocol
  *
  *  The ecdsa_anti_klepto_* functions can be used to prevent a signing device from
  *  exfiltrating the secret signing keys through biased signature nonces. The general
  *  idea is that a host provides additional randomness to the signing device client
  *  and the client commits to the randomness in the nonce using sign-to-contract.
  *
+ *  The following scheme is described by Stepan Snirigiv here:
```
apoelstra, line 106: Oops, I really butchered Stepan's name here.

include/secp256k1_ecdsa_s2c.h
real-or-random: Hm yeah... Idk, throwing away the device may not be a solution either if it's the only thing that stores your key.  It's just hard to give advice in that case.

We should consistently call it "signing device" or "hardware device".

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,180 @@
...
+ *  exfiltrating the secret signing keys through biased signature nonces. The general
+ *  idea is that a host provides additional randomness to the signing device client
+ *  and the client commits to the randomness in the nonce using sign-to-contract.
+ *
+ *  The following scheme is described by Stepan Snirigiv here:
+ *    https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2020-February/017655.html
+ *  and by Pieter Wuille (as "Scheme 6") here:
+ *    https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2020-March/017667.html
+ *
+ *  In order to ensure the host cannot trick the signing device into revealing its
```
real-or-random, line 111: It worries me that this is a mix of protocol description and usage instructions. 

I think it will be easier for the user to have clear instructions with just what functions must be called in which order and on what device and what messages are transferred. And we can still have a protocol description.

Host:
 * Call..  
 * Initiate signing session by sending...
 * .. 

Singing device:
 * Listen for messages  ... 
 * Call... 
 * ...  

Now that I wrote that, this will ideally be example code.  But we could do this in another PR instead if that's easier.

real-or-random:
Note (no need to do this in this PR, I just want to write it down):

I think we could add synthetic nonces to this scheme. Sure this will break its determinism, and then the HW needs to keep state. But I think there’s no need for secure storage. HW could simply draw some aux_rand, add it to the nonce derivation and store it. Then in the second round, look up aux_rand and recompute the nonce.

include/secp256k1_ecdsa_s2c.h
apoelstra: Let's do it in another PR. I think it's reasonable to have a protocol description which includes which API functions correspond to what steps.

include/secp256k1_ecdsa_s2c.h
benma: Throwing away the signer cause of one failed session might be too much, especially since there might be transient errors (e.g. a communication error after step 3). It might be good to distinguish between a stopped process after step 3, and an actually failed commitment check.

Maybe something like 'If the signing process fails repeatedly'? To produce even one signature with a desired bias, multiple attempts by the signing device might be needed, and multiple signatures with a correct bias are needed to leak a full key. 

Of course, it might fail 'repeatedly' over many days or months and for many different transactions, which is harder to recognize for a user :shrug:

src/modules/ecdsa_sign_to_contract/main_impl.h
jonasnick: Okay, seems like one idea that has support is to keep `s2c_verify` as is, clearly document that it doesn't do ecdsa_verify and for the antiklepto protocol add a host_verify function that does both `s2c_verify` and `ecdsa_verify`.

apoelstra:
remaining todo 

- [ ] update docs for what to do in case of failure
- [ ] make anti_klepto functions for all steps, even if they are just wrappers (or almost-wrappers) around s2c
- [ ] add example code

include/secp256k1_ecdsa_s2c.h
apoelstra: Well, failing every few weeks or months is not a big deal. In principle you can exfiltrate a nonce that way but in practice you cannot. Even if you have a whole bit of bias (and failing every time is likely to make a user throw away your device even if they don't know any crypto..) you still need hundreds of signatures with the same key to exfiltrate a nonce this way. And a single bit is on the edge of what's computationally feasible to work with.

It's probably fine to say "if this fails a noticeable number of times, over the lifetime of the device, you should throw it away".

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,220 @@
...
+ *     The signing device sends the resulting `R` to the host as a s2c_opening.
+ *  3. The host replies with `rho` generated in step 1.
+ *  4. The device signs with `secp256k1_anti_klepto_sign`, using `rho` as `host_data32`,
+ *     and sends the signature to the host.
+ *  5. The host verifies that the signature's public nonce matches the opening from
+ *     step 2 and its original randomness `rho`, using `secp256k1_anti_klepto_host_verify`.
+ *     End users should be advised that if this fails a noticeable proportion of the time
+ *     (say, more than a few times in the device's lifetime, absent obvious connectivity
+ *     problems), it indicates a serious problem with the device, which should probably
+ *     be securely destroyed.
```
jonasnick, line 127: This should reflect that not only is it a problem if `host_verify` fails but also if the signing device doesn't answer after step 3.

jonasnick (commented):
859d73ef829dce9f7436a2a42a86fa6d933edbb6 lgtm mod nit. Feel free to squash.

src/secp256k1.c
```diff
@@ -502,6 +515,10 @@ static int secp256k1_ecdsa_sign_inner(const secp256k1_context* ctx, secp256k1_sc
     if (noncefp == NULL) {
         noncefp = secp256k1_nonce_function_default;
     }
+    /* sign-to-contract commitments only work with the default nonce function,
+     * because we need to ensure that s2c_data is actually hashed into the nonce and
+     * not just ignored. */
+    VERIFY_CHECK(s2c_data32 == NULL || noncefp == secp256k1_nonce_function_default);
```
real-or-random, line 521: I believe I'm missing something obvious but don't understand this. Isn't the nonce derivation function (of the inner nonce) unrelated to the s2c commitment?

real-or-random (commented):
I think the API is pretty nice now! 
I haven't checked the tests in detail. Sorry that this is my first in-depth review and that I have many late comments now....

src/secp256k1.c
apoelstra: No, because if the inner nonce is the same for two differing s2c commitments (in two signatures which have the same key/message), this will leak the secret nonce.

real-or-random: Ok, I see. But maybe add to the comment that this is the reason that `s2c_data` should be hashed. This is different than the usual "sign two different messages with the same nonce" thing.

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,220 @@
...
+ *  Returns: 1: signature created
+ *           0: the nonce generation function failed, or the private key was invalid.
+ *  Args:    ctx:  pointer to a context object, initialized for signing (cannot be NULL)
+ *  Out:     sig:  pointer to an array where the signature will be placed (cannot be NULL)
+ *   s2c_opening:  if non-NULL, pointer to an secp256k1_ecdsa_s2c_opening structure to populate
+ *  In:    msg32: the 32-byte message hash being signed (cannot be NULL)
+ *        seckey: pointer to a 32-byte secret key (cannot be NULL)
+ *    s2c_data32: pointer to a 32-byte data to commit to in the nonce (cannot be NULL)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_ecdsa_s2c_sign(
```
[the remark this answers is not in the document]
apoelstra: Lol, it should.

src/eccommit.h
```diff
@@ -0,0 +1,24 @@
+/**********************************************************************
+ * Copyright (c) 2020 The libsecp256k1 Developers                     *
```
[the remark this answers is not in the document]
apoelstra: Heh, I suppose ... this came from a libsecp PR, and we're all the same people on both sides..

src/modules/ecdsa_s2c/main_impl.h
```diff
@@ -0,0 +1,199 @@
...
+     * derive the nonce It is first hashed because it should be possible
+     * to derive nonces even if only a SHA256 commitment to the data is
+     * known.  This is important in the ECDSA anti-klepto protocol. */
+    secp256k1_s2c_ecdsa_data_sha256_tagged(&s2c_sha);
+    secp256k1_sha256_write(&s2c_sha, s2c_data32, 32);
+    secp256k1_sha256_finalize(&s2c_sha, ndata);
+    noncedata = ndata;
+
+    secp256k1_s2c_ecdsa_point_sha256_tagged(&s2c_sha);
+    ret = secp256k1_ecdsa_sign_inner(ctx, &r, &s, NULL, &s2c_sha, s2c_opening, s2c_data32, msg32, seckey, NULL, noncedata);
```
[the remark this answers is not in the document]
apoelstra: No, there just used to be a condiitional assignment here that I removed.

src/modules/ecdsa_s2c/main_impl.h
```diff
@@ -0,0 +1,199 @@
...
+    ARG_CHECK(secp256k1_ecmult_gen_context_is_built(&ctx->ecmult_gen_ctx));
+    ARG_CHECK(opening != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(seckey32 != NULL);
+    ARG_CHECK(rand_commitment32 != NULL);
+
+    memset(nonce32, 0, 32);
+    while (!is_nonce_valid) {
+        /* cast to void* removes const qualifier, but secp256k1_nonce_function_default does not modify it */
+        CHECK(secp256k1_nonce_function_default(nonce32, msg32, seckey32, NULL, (void*)rand_commitment32, count));
```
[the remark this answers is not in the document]
apoelstra: We can't, signing repeats with a new nonce but we are forced by the original nonce and the commitment.

We could return a failure here, but this is cryptographically unreachable.

src/modules/ecdsa_s2c/main_impl.h
```diff
@@ -0,0 +1,199 @@
...
+
+    /* Check that sig_r == commitment_x (mod n)
+     * sig_r is the x coordinate of R represented by a scalar.
+     * commitment_x is the x coordinate of the commitment (field element).
+     *
+     * It is sufficient to only compare the x coordinates as it is as difficult to find an client
+     * commitment to the negation of the point as to any other point. There is a small reduction in
+     * security as it is easier to find a collision with a point and its negation.
+     */
+    secp256k1_ecdsa_signature_load(ctx, &sigr, &sigs, sig);
```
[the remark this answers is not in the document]
apoelstra: What is the additional code? I need to verify the ECDSA siganture to get the full R point don't I?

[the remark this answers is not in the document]
real-or-random: Oh ok, I understand now. I was on the wrong here track, and I believe the comment led me to onto this track.

> It is sufficient to only compare the x coordinates as it is as difficult to find an client commitment to the negation of the point as to any other point. There is a small reduction in security as it is easier to find a collision with a point and its negation.

The commitment comprises only the x-coordinate, so you can of course only verify the x-coordinate. And I would not call this a reduction in security. If you don't include y-bit, you can't expect that it provides any security.

Having said this, it's true that if the signature is valid, it implies a particular y-bit, so for *valid* signatures, you could compare the entire point. But since we separated verification of the commitment and the signature, this is not relevant for us.   

Now when I read the commitment again, it makes sense in the context of the antiklepto protocol. (Note also the term "client commitment".) But this should not bother us either since we also separated s2c and antiklepto.

include/secp256k1_ecdsa_s2c.h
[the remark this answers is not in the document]
real-or-random: Hehe, no idea. I suspect it's like this on purpose. The unused result is catastrophic for verification but typically not for signing. If you ask me, I'd prefer to keep it consistent but it doesn't really matter in the end.

src/modules/ecdsa_s2c/main_impl.h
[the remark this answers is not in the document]
real-or-random: The implicit convention is to use CHECK only in the tests, and call the error callback if you want to abort in the code. We violate this rule exactly once in the current code case (in the gmp inversion code but we don't want to pass down the context down to the field code, and hopefully we'll get rid of this exception soon.)

I see your point that returning failure just makes life of the user harder. So I think the error callback is the right thing here.

src/secp256k1.c
```diff
@@ -517,7 +517,8 @@ static int secp256k1_ecdsa_sign_inner(const secp256k1_context* ctx, secp256k1_sc
     }
     /* sign-to-contract commitments only work with the default nonce function,
      * because we need to ensure that s2c_data is actually hashed into the nonce and
-     * not just ignored. */
+     * not just ignored. Otherwise an attacker can exfiltrate the secret key by
+     * signing the same message twice with different commitments. */
```
real-or-random, line 521: twice? ;)

real-or-random (commented):
The fixes look good, feel free to squash. :+1:

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,222 @@
...
+ *        seckey: pointer to a 32-byte secret key (cannot be NULL)
+ *   host_data32: pointer to 32-byte host-provided randomness (cannot be NULL)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_anti_klepto_sign(
+    const secp256k1_context* ctx,
+    secp256k1_ecdsa_signature* sig,
+    const unsigned char* msg32,
+    const unsigned char* seckey,
+    const unsigned char* host_data32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
```
benma, line 196: Can we add `int* recid` here and populate it if it is not NULL? It is only a matter of passing the argument down to `secp256k1_ecdsa_sign_inner()`.

src/secp256k1.c
apoelstra: Lol, fine, do we agree that you can do it with three sigs at least?

include/secp256k1_ecdsa_s2c.h
apoelstra: I really don't want to support recoverable signatures any more than we absolutely have to.

benma: In the BitBox02 [we make use of it](https://github.com/digitalbitbox/bitbox02-firmware/blob/e5b6782c7568370f0f6022feef2f33964498b77d/src/keystore.h#L194). I plan on rolling out the anti-klepto protocol there, so it would definitely make my life a bit easier. Otherwise I will likely maintain a small commit on top which adds this here.

apoelstra: I think I'm gonna make you maintain a patch, sorry. It's only a couple lines for you to expose this variable but for us to do it properly we'd need to add another function gated on the `SECP256K1_RECOVERY` feature flag.

benma: Okay, but for my information: why gate it behind `SECP256K1_RECOVERY`? I would have just exposed without that. It is not in the recovery module after all, all the functionality is present in secp256k1.c.

benma:
tACK 374d944. I used this branch to add support for the anti klepto protocol when signing a transaction in the BitBox02, including successful host verification in the bitbox02 client libraries. Thanks again for picking up my original PR.

include/secp256k1_ecdsa_s2c.h
apoelstra: Because `SECP256K1_RECOVERY` is the (semi-deprecated) feature flag to get recoverable signatures :). We don't want to support back-door ways to get this functionality.

apoelstra:
Okay, I think we're actually good to go now.

include/secp256k1_ecdsa_s2c.h
```diff
@@ -0,0 +1,234 @@
...
+ *     as auxiliary randomness, using `secp256k1_ecdsa_anti_klepto_signer_commit`.
+ *     The signing device sends the resulting `R` to the host as a s2c_opening.
+ *
+ *     If, at any point from this step onward, the hardware device fails, it is
+ *     okay to restart the protocol using **exactly the same `rho`** and checking
+ *     that the hardware device proposes **exactly the same** `R`. Otherwise, the
+ *     hardware device may be selectively aborting and thereby biasing the set of
+ *     nonces that are used in actual signatures.
+ *
+ *     It takes many (>100) such aborts before there is a plausible attack, given
```
jonasnick, line 127: Micro-nit: I'd say it takes about a 100 such aborts not more. To communicate 200 bits you need no more than 100 aborts on average and the remaining 56 bits are somewhat possible to brute-force.

src/eccommit.h
```diff
@@ -0,0 +1,24 @@
...
+#ifndef SECP256K1_ECCOMMIT_H
+#define SECP256K1_ECCOMMIT_H
+
+/** Helper function to add a 32-byte value to a scalar */
+static int secp256k1_ec_seckey_tweak_add_helper(secp256k1_scalar *sec, const unsigned char *tweak);
+/** Helper function to add a 32-byte value, times G, to an EC point */
+static int secp256k1_ec_pubkey_tweak_add_helper(const secp256k1_ecmult_context* ecmult_ctx, secp256k1_ge *p, const unsigned char *tweak);
+
+/** Compute an ec commitment tweak as hash(pubkey, data). */
+static void secp256k1_ec_commit_tweak(unsigned char *tweak32, secp256k1_ge* pubp, secp256k1_sha256* sha, const unsigned char *data, size_t data_size);
```
jonasnick, line 16: We should document in (all) ec_commit functions that pubp must not be the point at infinity. Or alternatively, correctly deal with that case and declassify only the infinity bit in `ecdsa_sign_inner`. I implemented that here: https://github.com/jonasnick/secp256k1-zkp/commits/ecdsa-s2c-inf Feel free to cherry-pick & squash.

real-or-random:
Fixes look good.

I think this one is still open:
https://github.com/ElementsProject/secp256k1-zkp/pull/111#discussion_r543138109 (change the comment)

And not sure if you overlooked that one or delibaretly choose to keep the current version:
https://github.com/ElementsProject/secp256k1-zkp/pull/111#discussion_r543140968

apoelstra:
@jonasnick sure, cherry-picked your commit. I have a mild preference for just documenting that the function won't work, since all of our usage avoids the point at infinity, but since you took the time to implement this and add a test and fix the ctime tests, I might as well take your complete work :).

@real-or-random ok changed the comment. It was unclear to me from your earlier comment that you actually wanted me to make a change there.

jonasnick:
Thought that the ec_commit functions will be useful in other contexts where it isn't as clear that there is no point at infinity. PR looks good. Are you going to squash this?

apoelstra:
Yep, give me 10 minutes to squash.

apoelstra:
Heh, took more than ten minutes. But there you go.

jonasnick (commented):
Nice! :+1: 

ACK 47efb5e39a1bf6330bd3bf6bc4b4416c5ca11878

benma:
@apoelstra is this PR ready to be merged? I have a bit of downtime to work on the downstream integration.

apoelstra:
I think so. @real-or-random  @jonasnick can you merge this?

real-or-random (approved):
ACK 47efb5e39a1bf6330bd3bf6bc4b4416c5ca11878

benma:
Awesome. Thanks everyone!

include/secp256k1_ecdsa_s2c.h
benma: @jonasnick 

> If the hardware device is not discarded after wrongly or not at all responding after step 3, the hardware device can successfully exfiltrate the signing key even if the host only lets signatures pass after verifying the commitment. This is because after step 3, the signing device knows the final nonce and can simply abort if it does not match the desired bias.

It only knows the final nonce if the host nonce is the same in a restart, right? If the host picks a new random nonce contribution every time, a restart after step 2 should be safe.

So the comment in the code:

> If, at any point from this step onward, the hardware device fails, it is
> okay to restart the protocol using **exactly the same `rho`** and checking
> that the hardware device proposes **exactly the same** `R`. Otherwise, the
> hardware device may be selectively aborting and thereby biasing the set of
> nonces that are used in actual signatures.

is not complete I think. We could add that the protocol can also be restarted with a fresh `rho`. 

Am I missing something?

jonasnick: Not sure I understand
1. Restart after step 2 is the same as restart after step 3, because the host must provide the same `rho` in step 3.
2. Restarting with a fresh `rho` means restarting from step 1 which allows the hardware wallet to bias the final nonce.

benma: > Restart after step 2 is the same as restart after step 3, because the host must provide the same rho in step 3.

How so? After step 2 (before step 3), the signer does not know the final nonce yet.

I get that after step 3, the signer knows `rho` and with that the final nonce, so it can introduce bias by aborting the process. 

But if there is a restart before that (e.g. after the signer commitment is delivered, before sending the `rho`), restarting with a fresh `rho` should be fine? 

Not that it matters much in practice, it was more for my information and to maybe move the comment in the code to step 3.

jonasnick: Oh, I think I misunderstood.

> a restart after step 2 should be safe.

You mean you can restart the whole protocol if step 3 hasn't happened yet ( I thought you mean that after restart the protocol can continue after step 2). I think in that case it can also be run with the same rho.The host can just re-request step 2 if the HWW failst to provide a step 2 messge. Yeah I think it would make sense to move the "If, at any point from this step onward, the hardware device fails, it is [...]" section to step 3 but on the other hand it shouldn't make a difference in practice - at least I don't really see how the protocol can fail at step 3.

benma: Thanks for confirming.

> at least I don't really see how the protocol can fail at step 3.

When the host sends `rho`, the signer could just not respond, or respond with an error message, or respond with invalid data.

src/eccommit_impl.h
```diff
@@ -0,0 +1,73 @@
...
+    return 1;
+}
+
+/* Compute an ec commitment as pubp + hash(pubp, data)*G. */
+static int secp256k1_ec_commit(const secp256k1_ecmult_context* ecmult_ctx, secp256k1_ge* commitp, const secp256k1_ge* pubp, secp256k1_sha256* sha, const unsigned char *data, size_t data_size) {
+    unsigned char tweak[32];
+
+    *commitp = *pubp;
+    return secp256k1_ec_commit_tweak(tweak, commitp, sha, data, data_size)
+           && secp256k1_ec_pubkey_tweak_add_helper(ecmult_ctx, commitp, tweak);
```
roconnor-blockstream, line 47: `secp256k1_ec_pubkey_tweak_add_helper` (and `secp256k1_ec_seckey_tweak_add_helper` below) implement the unusual BIP32 semantics of failing (and moving to the next code) when the tweak value exceeds the group order.  No other modern protocol does this. BIP-340's default signing takes k' = int(rand) mod n; [Jonas's proposed key aggregation](https://github.com/jonasnick/secp256k1-zkp/blob/62f8a04e4da8329d36eb1296f0d8d06d1c5a45db/src/modules/musig/musig-spec.mediawiki#Key_Aggregation)'s 'ComputeCoefficient' is taken modulo n.  And when you write up the ecdsa_s2c protocol documentation, I expect you will also want to take the this tweak value modulo n.

I don't think we want to fold in BIP32's behavior into the s2c protocol specification, just because it is expedient from a libsecp256k1 implementation perspective.

roconnor-blockstream: Jonas points out that the tweaking in BIP-341 isn't taken modulo n.

So I give up.  tweaks are not taken modulo n for some reason, but everything else is.

include/secp256k1_ecdsa_s2c.h
roconnor-blockstream: > It's probably fine to say "if this fails a noticeable number of times, over the lifetime of the device, you should throw it away".

Throwing it away isn't enough.  You should act like your master secret has been compromised.

PR #117, opened 2020-12-30
Add ECDSA adaptor signatures module

jesseposner:
### Overview

This PR adds support for ECDSA adaptor signatures. It is based on the work of @jonasnick in https://github.com/jonasnick/secp256k1/pull/14.

### ECDSA Adaptor Signatures

This implementation conforms to the [DLC specification](https://github.com/discreetlogcontracts/dlcspecs/pull/114) and includes test vectors from the specification.

The specification is based on the [paper](https://github.com/LLFourn/one-time-VES/blob/master/main.pdf) "One-Time Verifiably Encrypted Signatures A.K.A. Adaptor Signatures" by @LLFourn. The idea was first proposed in a lightning-dev [post](https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002316.html).

### Python

I've [implemented](https://github.com/jesseposner/dlc/blob/master/ecdsa_adaptor.py) the specification in Python as well, which can be useful for testing purposes. [needs to be updated to match the latest spec revision]

### DIFF
--- a/include/secp256k1_ecdsa_adaptor.h
+++ b/include/secp256k1_ecdsa_adaptor.h
@@ -1,0 +1,19 @@
+#ifndef SECP256K1_ECDSA_ADAPTOR_H
+#define SECP256K1_ECDSA_ADAPTOR_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/** This module implements single signer ECDSA adaptor signatures following
+ *  "One-Time Verifiably Encrypted Signatures A.K.A. Adaptor Signatures" by
+ *  Lloyd Fournier
+ *  (https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002316.html
+ *  and https://github.com/LLFourn/one-time-VES/blob/master/main.pdf).
+*/
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* SECP256K1_ECDSA_ADAPTOR_H */
--- a/include/secp256k1_ecdsa_adaptor.h
+++ b/include/secp256k1_ecdsa_adaptor.h
@@ -12,6 +12,40 @@ extern "C" {
  *  and https://github.com/LLFourn/one-time-VES/blob/master/main.pdf).
 */
 
+/** A pointer to a function to deterministically generate a nonce.
+ *
+ *  Same as secp256k1_nonce_function_hardened with the exception of using the
+ *  compressed 33-byte encoding for the pubkey argument.
+ *
+ *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
+ *           return an error.
+ *  Out:     nonce32:   pointer to a 32-byte array to be filled by the function
+ *  In:        msg32:   the 32-byte message hash being verified
+ *             key32:   pointer to a 32-byte secret key
+ *              pk33:   the 33-byte serialized pubkey corresponding to key32
+ *              algo:   pointer to an array describing the signature algorithm
+ *           algolen:   the length of the algo array
+ *              data:   arbitrary data pointer that is passed through
+ *
+ *  Except for test cases, this function should compute some cryptographic hash of
+ *  the message, the key, the pubkey, the algorithm description, and data.
+ */
+typedef int (*secp256k1_nonce_function_hardened_ecdsa_adaptor)(
+    unsigned char *nonce32,
+    const unsigned char *msg32,
+    const unsigned char *key32,
+    const unsigned char *pk33,
+    const unsigned char *algo,
+    size_t algolen,
+    void *data
+);
+
+/** A modified BIP-340 nonce generation function. If a data pointer is passed, it is
+ *  assumed to be a pointer to 32 bytes of auxiliary random data as defined in BIP-340.
+ *  The hash will be tagged with algo after removing all terminating null bytes.
+ */
+SECP256K1_API extern const secp256k1_nonce_function_hardened_ecdsa_adaptor secp256k1_nonce_function_ecdsa_adaptor;
+
 #ifdef __cplusplus
 }
 #endif
--- a/include/secp256k1_ecdsa_adaptor.h
+++ b/include/secp256k1_ecdsa_adaptor.h
@@ -10,7 +10,22 @@ extern "C" {
  *  Lloyd Fournier
  *  (https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002316.html
  *  and https://github.com/LLFourn/one-time-VES/blob/master/main.pdf).
-*/
+ *
+ *  WARNING! DANGER AHEAD!
+ *  As mentioned in Lloyd Fournier's paper, the adaptor signature leaks the
+ *  Elliptic-curve Diffie–Hellman (ECDH) key between the signing key and the
+ *  encryption key. This is not a problem for ECDSA adaptor signatures
+ *  themselves, but may result in a complete loss of security when they are
+ *  composed with other schemes. More specifically, let us refer to the
+ *  signer's public key as X = x*G, and to the encryption key as Y = y*G.
+ *  Given X, Y and the adaptor signature, it is trivial to compute Y^x = X^y.
+ *
+ *  A defense is to not reuse the signing key of ECDSA adaptor signatures in
+ *  protocols that rely on the hardness of the CDH problem, e.g., Diffie-Hellman
+ *  key exchange and ElGamal encryption. In general, it is a well-established
+ *  cryptographic practice to seperate keys for different purposes whenever
+ *  possible.
+ */
 
 /** A pointer to a function to deterministically generate a nonce.
  *
@@ -45,6 +60,100 @@ typedef int (*secp256k1_nonce_function_hardened_ecdsa_adaptor)(
  *  The hash will be tagged with algo after removing all terminating null bytes.
  */
 SECP256K1_API extern const secp256k1_nonce_function_hardened_ecdsa_adaptor secp256k1_nonce_function_ecdsa_adaptor;
+
+/** Encrypted Signing
+ *
+ *  Creates an adaptor signature, which includes a proof to verify the adaptor
+ *  signature.
+ *  WARNING: Make sure you have read and understood the WARNING at the top of
+ *  this file and applied the suggested countermeasures.
+ *
+ *  Returns: 1 on success, 0 on failure
+ *  Args:             ctx: a secp256k1 context object, initialized for signing
+ *  Out:   adaptor_sig162: pointer to 162 byte to store the returned signature
+ *  In:          seckey32: pointer to 32 byte secret key that will be used for
+ *                         signing
+ *                 enckey: pointer to the encryption public key
+ *                  msg32: pointer to the 32-byte message hash to sign
+ *                noncefp: pointer to a nonce generation function. If NULL,
+ *                         secp256k1_nonce_function_ecdsa_adaptor is used
+ *                  ndata: pointer to arbitrary data used by the nonce generation
+ *                         function (can be NULL). If it is non-NULL and
+ *                         secp256k1_nonce_function_ecdsa_adaptor is used, then
+ *                         ndata must be a pointer to 32-byte auxiliary randomness
+ *                         as per BIP-340.
+ */
+SECP256K1_API int secp256k1_ecdsa_adaptor_encrypt(
+    const secp256k1_context* ctx,
+    unsigned char *adaptor_sig162,
+    unsigned char *seckey32,
+    const secp256k1_pubkey *enckey,
+    const unsigned char *msg32,
+    secp256k1_nonce_function_hardened_ecdsa_adaptor noncefp,
+    void *ndata
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Encryption Verification
+ *
+ *  Verifies that the adaptor decryption key can be extracted from the adaptor signature
+ *  and the completed ECDSA signature.
+ *
+ *  Returns: 1 on success, 0 on failure
+ *  Args:            ctx: a secp256k1 context object, initialized for verification
+ *  In:   adaptor_sig162: pointer to 162-byte signature to verify
+ *                pubkey: pointer to the public key corresponding to the secret key
+ *                        used for signing
+ *                 msg32: pointer to the 32-byte message hash being verified
+ *                enckey: pointer to the adaptor encryption public key
+ */
+SECP256K1_API int secp256k1_ecdsa_adaptor_verify(
+    const secp256k1_context* ctx,
+    const unsigned char *adaptor_sig162,
+    const secp256k1_pubkey *pubkey,
+    const unsigned char *msg32,
+    const secp256k1_pubkey *enckey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Signature Decryption
+ *
+ *  Derives an ECDSA signature from an adaptor signature and an adaptor decryption key.
+ *
+ *  Returns: 1 on success, 0 on failure
+ *  Args:              ctx: a secp256k1 context object
+ *  Out:               sig: pointer to the ECDSA signature to create
+ *  In:           deckey32: pointer to 32-byte decryption secret key for the adaptor
+ *                          encryption public key
+ *          adaptor_sig162: pointer to 162-byte adaptor sig
+ */
+SECP256K1_API int secp256k1_ecdsa_adaptor_decrypt(
+    const secp256k1_context* ctx,
+    secp256k1_ecdsa_signature *sig,
+    const unsigned char *deckey32,
+    const unsigned char *adaptor_sig162
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Decryption Key Recovery
+ *
+ *  Extracts the adaptor decryption key from the complete signature and the adaptor
+ *  signature.
+ *
+ *  Returns: 1 on success, 0 on failure
+ *  Args:             ctx: a secp256k1 context object, initialized for signing
+ *  Out:         deckey32: pointer to 32-byte adaptor decryption key for the adaptor
+ *                         encryption public key
+ *  In:               sig: pointer to ECDSA signature to recover the adaptor decryption
+ *                         key from
+ *         adaptor_sig162: pointer to adaptor signature to recover the adaptor
+ *                         decryption key from
+ *                 enckey: pointer to the adaptor encryption public key
+ */
+SECP256K1_API int secp256k1_ecdsa_adaptor_recover(
+    const secp256k1_context* ctx,
+    unsigned char *deckey32,
+    const secp256k1_ecdsa_signature *sig,
+    const unsigned char *adaptor_sig162,
+    const secp256k1_pubkey *enckey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
 #ifdef __cplusplus
 }
--- a/src/modules/ecdsa_adaptor/dleq_impl.h
+++ b/src/modules/ecdsa_adaptor/dleq_impl.h
@@ -1,0 +1,158 @@
+#ifndef SECP256K1_DLEQ_IMPL_H
+#define SECP256K1_DLEQ_IMPL_H
+
+/* Initializes SHA256 with fixed midstate. This midstate was computed by applying
+ * SHA256 to SHA256("DLEQ")||SHA256("DLEQ"). */
+static void secp256k1_nonce_function_dleq_sha256_tagged(secp256k1_sha256 *sha) {
+    secp256k1_sha256_initialize(sha);
+    sha->s[0] = 0x8cc4beacul;
+    sha->s[1] = 0x2e011f3ful;
+    sha->s[2] = 0x355c75fbul;
+    sha->s[3] = 0x3ba6a2c5ul;
+    sha->s[4] = 0xe96f3aeful;
+    sha->s[5] = 0x180530fdul;
+    sha->s[6] = 0x94582499ul;
+    sha->s[7] = 0x577fd564ul;
+
+    sha->bytes = 64;
+}
+
+/* algo argument for nonce_function_ecdsa_adaptor to derive the nonce using a tagged hash function. */
+static const unsigned char dleq_algo[4] = "DLEQ";
+
+static int secp256k1_dleq_hash_point(secp256k1_sha256 *sha, secp256k1_ge *p) {
+    unsigned char buf[33];
+    size_t size = 33;
+
+    if (!secp256k1_eckey_pubkey_serialize(p, buf, &size, 1)) {
+        return 0;
+    }
+
+    secp256k1_sha256_write(sha, buf, size);
+    return 1;
+}
+
+static int secp256k1_dleq_nonce(secp256k1_scalar *k, const unsigned char *sk32, const unsigned char *gen2_33, const unsigned char *p1_33, const unsigned char *p2_33, secp256k1_nonce_function_hardened_ecdsa_adaptor noncefp, void *ndata) {
+    secp256k1_sha256 sha;
+    unsigned char buf[32];
+    unsigned char nonce[32];
+    size_t size = 33;
+
+    if (noncefp == NULL) {
+        noncefp = secp256k1_nonce_function_ecdsa_adaptor;
+    }
+
+    secp256k1_sha256_initialize(&sha);
+    secp256k1_sha256_write(&sha, p1_33, size);
+    secp256k1_sha256_write(&sha, p2_33, size);
+    secp256k1_sha256_finalize(&sha, buf);
+
+    if (!noncefp(nonce, buf, sk32, gen2_33, dleq_algo, sizeof(dleq_algo), ndata)) {
+        return 0;
+    }
+    secp256k1_scalar_set_b32(k, nonce, NULL);
+    if (secp256k1_scalar_is_zero(k)) {
+        return 0;
+    }
+
+    return 1;
+}
+
+/* Generates a challenge as defined in the DLC Specification at
+ * https://github.com/discreetlogcontracts/dlcspecs */
+static void secp256k1_dleq_challenge(secp256k1_scalar *e, secp256k1_ge *gen2, secp256k1_ge *r1, secp256k1_ge *r2, secp256k1_ge *p1, secp256k1_ge *p2) {
+    unsigned char buf[32];
+    secp256k1_sha256 sha;
+
+    secp256k1_nonce_function_dleq_sha256_tagged(&sha);
+    secp256k1_dleq_hash_point(&sha, p1);
+    secp256k1_dleq_hash_point(&sha, gen2);
+    secp256k1_dleq_hash_point(&sha, p2);
+    secp256k1_dleq_hash_point(&sha, r1);
+    secp256k1_dleq_hash_point(&sha, r2);
+    secp256k1_sha256_finalize(&sha, buf);
+
+    secp256k1_scalar_set_b32(e, buf, NULL);
+}
+
+/* P1 = x*G, P2 = x*Y */
+static void secp256k1_dleq_pair(const secp256k1_ecmult_gen_context *ecmult_gen_ctx, secp256k1_ge *p1, secp256k1_ge *p2, const secp256k1_scalar *sk, const secp256k1_ge *gen2) {
+    secp256k1_gej p1j, p2j;
+
+    secp256k1_ecmult_gen(ecmult_gen_ctx, &p1j, sk);
+    secp256k1_ge_set_gej(p1, &p1j);
+    secp256k1_ecmult_const(&p2j, gen2, sk, 256);
+    secp256k1_ge_set_gej(p2, &p2j);
+}
+
+/* Generates a proof that the discrete logarithm of P1 to the secp256k1 base G is the
+ * same as the discrete logarithm of P2 to the base Y */
+static int secp256k1_dleq_prove(const secp256k1_context* ctx, secp256k1_scalar *s, secp256k1_scalar *e, const secp256k1_scalar *sk, secp256k1_ge *gen2, secp256k1_ge *p1, secp256k1_ge *p2, secp256k1_nonce_function_hardened_ecdsa_adaptor noncefp, void *ndata) {
+    secp256k1_ge r1, r2;
+    secp256k1_scalar k = { 0 };
+    unsigned char sk32[32];
+    unsigned char gen2_33[33];
+    unsigned char p1_33[33];
+    unsigned char p2_33[33];
+    int ret = 1;
+    size_t pubkey_size = 33;
+
+    secp256k1_scalar_get_b32(sk32, sk);
+    if (!secp256k1_eckey_pubkey_serialize(gen2, gen2_33, &pubkey_size, 1)) {
+        return 0;
+    }
+    if (!secp256k1_eckey_pubkey_serialize(p1, p1_33, &pubkey_size, 1)) {
+        return 0;
+    }
+    if (!secp256k1_eckey_pubkey_serialize(p2, p2_33, &pubkey_size, 1)) {
+        return 0;
+    }
+
+    ret &= secp256k1_dleq_nonce(&k, sk32, gen2_33, p1_33, p2_33, noncefp, ndata);
+    /* R1 = k*G, R2 = k*Y */
+    secp256k1_dleq_pair(&ctx->ecmult_gen_ctx, &r1, &r2, &k, gen2);
+    /* We declassify the non-secret values r1 and r2 to allow using them as
+     * branch points. */
+    secp256k1_declassify(ctx, &r1, sizeof(r1));
+    secp256k1_declassify(ctx, &r2, sizeof(r2));
+
+    /* e = tagged hash(p1, gen2, p2, r1, r2) */
+    /* s = k + e * sk */
+    secp256k1_dleq_challenge(e, gen2, &r1, &r2, p1, p2);
+    secp256k1_scalar_mul(s, e, sk);
+    secp256k1_scalar_add(s, s, &k);
+
+    secp256k1_scalar_clear(&k);
+    return ret;
+}
+
+static int secp256k1_dleq_verify(const secp256k1_ecmult_context *ecmult_ctx, const secp256k1_scalar *s, const secp256k1_scalar *e, secp256k1_ge *p1, secp256k1_ge *gen2, secp256k1_ge *p2) {
+    secp256k1_scalar e_neg;
+    secp256k1_scalar e_expected;
+    secp256k1_gej gen2j;
+    secp256k1_gej p1j, p2j;
+    secp256k1_gej r1j, r2j;
+    secp256k1_ge r1, r2;
+    secp256k1_gej tmpj;
+
+    secp256k1_gej_set_ge(&p1j, p1);
+    secp256k1_gej_set_ge(&p2j, p2);
+
+    secp256k1_scalar_negate(&e_neg, e);
+    /* R1 = s*G  - e*P1 */
+    secp256k1_ecmult(ecmult_ctx, &r1j, &p1j, &e_neg, s);
+    /* R2 = s*gen2 - e*P2 */
+    secp256k1_ecmult(ecmult_ctx, &tmpj, &p2j, &e_neg, &secp256k1_scalar_zero);
+    secp256k1_gej_set_ge(&gen2j, gen2);
+    secp256k1_ecmult(ecmult_ctx, &r2j, &gen2j, s, &secp256k1_scalar_zero);
+    secp256k1_gej_add_var(&r2j, &r2j, &tmpj, NULL);
+
+    secp256k1_ge_set_gej(&r1, &r1j);
+    secp256k1_ge_set_gej(&r2, &r2j);
+    secp256k1_dleq_challenge(&e_expected, gen2, &r1, &r2, p1, p2);
+
+    secp256k1_scalar_add(&e_expected, &e_expected, &e_neg);
+    return secp256k1_scalar_is_zero(&e_expected);
+}
+
+#endif
--- a/src/modules/ecdsa_adaptor/main_impl.h
+++ b/src/modules/ecdsa_adaptor/main_impl.h
@@ -1,0 +1,12 @@
+/**********************************************************************
+ * Copyright (c) 2020-2021 Jonas Nick, Jesse Posner                   *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef SECP256K1_MODULE_ECDSA_ADAPTOR_MAIN_H
+#define SECP256K1_MODULE_ECDSA_ADAPTOR_MAIN_H
+
+#include "include/secp256k1_ecdsa_adaptor.h"
+
+#endif
--- a/src/modules/ecdsa_adaptor/main_impl.h
+++ b/src/modules/ecdsa_adaptor/main_impl.h
@@ -9,4 +9,82 @@
 
 #include "include/secp256k1_ecdsa_adaptor.h"
 
+/* Initializes SHA256 with fixed midstate. This midstate was computed by applying
+ * SHA256 to SHA256("ECDSAadaptor/non")||SHA256("ECDSAadaptor/non"). */
+static void secp256k1_nonce_function_ecdsa_adaptor_sha256_tagged(secp256k1_sha256 *sha) {
+    secp256k1_sha256_initialize(sha);
+    sha->s[0] = 0x791dae43ul;
+    sha->s[1] = 0xe52d3b44ul;
+    sha->s[2] = 0x37f9edeaul;
+    sha->s[3] = 0x9bfd2ab1ul;
+    sha->s[4] = 0xcfb0f44dul;
+    sha->s[5] = 0xccf1d880ul;
+    sha->s[6] = 0xd18f2c13ul;
+    sha->s[7] = 0xa37b9024ul;
+
+    sha->bytes = 64;
+}
+
+/* Initializes SHA256 with fixed midstate. This midstate was computed by applying
+ * SHA256 to SHA256("ECDSAadaptor/aux")||SHA256("ECDSAadaptor/aux"). */
+static void secp256k1_nonce_function_ecdsa_adaptor_sha256_tagged_aux(secp256k1_sha256 *sha) {
+    secp256k1_sha256_initialize(sha);
+    sha->s[0] = 0xd14c7bd9ul;
+    sha->s[1] = 0x095d35e6ul;
+    sha->s[2] = 0xb8490a88ul;
+    sha->s[3] = 0xfb00ef74ul;
+    sha->s[4] = 0x0baa488ful;
+    sha->s[5] = 0x69366693ul;
+    sha->s[6] = 0x1c81c5baul;
+    sha->s[7] = 0xc33b296aul;
+
+    sha->bytes = 64;
+}
+
+/* algo argument for nonce_function_ecdsa_adaptor to derive the nonce using a tagged hash function. */
+static const unsigned char ecdsa_adaptor_algo[16] = "ECDSAadaptor/non";
+
+/* Modified BIP-340 nonce function */
+static int nonce_function_ecdsa_adaptor(unsigned char *nonce32, const unsigned char *msg32, const unsigned char *key32, const unsigned char *pk33, const unsigned char *algo, size_t algolen, void *data) {
+    secp256k1_sha256 sha;
+    unsigned char masked_key[32];
+    int i;
+
+    if (algo == NULL) {
+        return 0;
+    }
+
+    if (data != NULL) {
+        secp256k1_nonce_function_ecdsa_adaptor_sha256_tagged_aux(&sha);
+        secp256k1_sha256_write(&sha, data, 32);
+        secp256k1_sha256_finalize(&sha, masked_key);
+        for (i = 0; i < 32; i++) {
+            masked_key[i] ^= key32[i];
+        }
+    }
+
+    /* Tag the hash with algo which is important to avoid nonce reuse across
+     * algorithims. An optimized tagging implementation is used if the default
+     * tag is provided. */
+    if (algolen == sizeof(ecdsa_adaptor_algo)
+            && secp256k1_memcmp_var(algo, ecdsa_adaptor_algo, algolen) == 0) {
+        secp256k1_nonce_function_ecdsa_adaptor_sha256_tagged(&sha);
+    } else {
+        secp256k1_sha256_initialize_tagged(&sha, algo, algolen);
+    }
+
+    /* Hash (masked-)key||pk||msg using the tagged hash as per BIP-340 */
+    if (data != NULL) {
+        secp256k1_sha256_write(&sha, masked_key, 32);
+    } else {
+        secp256k1_sha256_write(&sha, key32, 32);
+    }
+    secp256k1_sha256_write(&sha, pk33, 33);
+    secp256k1_sha256_write(&sha, msg32, 32);
+    secp256k1_sha256_finalize(&sha, nonce32);
+    return 1;
+}
+
+const secp256k1_nonce_function_hardened_ecdsa_adaptor secp256k1_nonce_function_ecdsa_adaptor = nonce_function_ecdsa_adaptor;
+
 #endif
--- a/src/modules/ecdsa_adaptor/main_impl.h
+++ b/src/modules/ecdsa_adaptor/main_impl.h
@@ -10,6 +10,61 @@
 #include "include/secp256k1_ecdsa_adaptor.h"
 #include "modules/ecdsa_adaptor/dleq_impl.h"
 
+/* (R, R', s', dleq_proof) */
+static int secp256k1_ecdsa_adaptor_sig_serialize(unsigned char *adaptor_sig162, secp256k1_ge *r, secp256k1_ge *rp, const secp256k1_scalar *sp, const secp256k1_scalar *dleq_proof_e, const secp256k1_scalar *dleq_proof_s) {
+    size_t size = 33;
+
+    if (!secp256k1_eckey_pubkey_serialize(r, adaptor_sig162, &size, 1)) {
+        return 0;
+    }
+    if (!secp256k1_eckey_pubkey_serialize(rp, &adaptor_sig162[33], &size, 1)) {
+        return 0;
+    }
+    secp256k1_scalar_get_b32(&adaptor_sig162[66], sp);
+    secp256k1_scalar_get_b32(&adaptor_sig162[98], dleq_proof_e);
+    secp256k1_scalar_get_b32(&adaptor_sig162[130], dleq_proof_s);
+
+    return 1;
+}
+
+static int secp256k1_ecdsa_adaptor_sig_deserialize(secp256k1_ge *r, secp256k1_scalar *sigr, secp256k1_ge *rp, secp256k1_scalar *sp, secp256k1_scalar *dleq_proof_e, secp256k1_scalar *dleq_proof_s, const unsigned char *adaptor_sig162) {
+    /* If r is deserialized, require that a sigr is provided to receive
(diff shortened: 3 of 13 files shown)


stevenans9859 (approved):

stevenans9859 (approved):

stevenans9859 (approved):

jonasnick:
Sorry this was automatically closed. PR needs to be reopened against the `master` branch.

jonasnick:
fixed it should be fine now

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+        }
+    }
+    if (rp != NULL) {
+        if (!secp256k1_eckey_pubkey_parse(rp, &adaptor_sig162[33], 33)) {
+            return 0;
+        }
+    }
+    if (sp != NULL) {
+        int overflow;
+        secp256k1_scalar_set_b32(sp, &adaptor_sig162[66], &overflow);
```
LLFourn, line 47: ❌ Must not be zero. I think the reason for this is that since ECDSA s can't be zero then there's no way a valid `sp` can be zero either.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+    return ret;
+}
+
+
+/* Modified secp256k1_ecdsa_sig_verify */
+static int secp256k1_ecdsa_adaptor_sig_verify(const secp256k1_context* ctx, secp256k1_gej *pr, secp256k1_scalar *sigr, secp256k1_scalar *sigs, const secp256k1_ge *pubkey, const secp256k1_scalar *message) {
+    secp256k1_scalar sn, u1, u2;
+    secp256k1_gej pubkeyj;
+
+    if (secp256k1_scalar_is_zero(sigr) || secp256k1_scalar_is_zero(sigs)) {
```
LLFourn, line 246: Ok I see now that you are checking this here. I guess that's a fine idea.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+        return 0;
+    }
+
+    /* return R' == s'⁻¹(m * G + R.x * X) */
+    if (!secp256k1_ecdsa_adaptor_sig_verify(ctx, &derived_rp, &sigr, &sp, &pubkey_ge, &msg)) {
+        return 0;
+    }
+    secp256k1_gej_neg(&derived_rp, &derived_rp);
+    secp256k1_gej_add_ge_var(&derived_rp, &derived_rp, &rp, NULL);
+    return secp256k1_gej_is_infinity(&derived_rp);
```
LLFourn, line 299: 🔧 It's confusing to me to have this crucial bit just dangling around here. Can we do this in  `ecdsa_adaptor_sig_verify` instead where the rest of the calculation of `derived_rp` is done. Or maybe just inline the above function here. I don't think it's making anything much clearer for me being separate.

LLFourn (changes requested):
Nice work @jesseposner. Very easy to review. I think there is some things to fixup wrt to serialization which were not specified.

include/secp256k1_ecdsa_adaptor.h
```diff
@@ -0,0 +1,156 @@
...
+ *                  msg32: pointer to the 32-byte message to sign (cannot be NULL)
+ *                noncefp: pointer to a nonce generation function. If NULL,
+ *                         secp256k1_nonce_function_ecdsa_adaptor is used
+ *                  ndata: pointer to arbitrary data used by the nonce generation
+ *                         function (can be NULL). If it is non-NULL and
+ *                         secp256k1_nonce_function_ecdsa_adaptor is used, then
+ *                         ndata must be a pointer to 32-byte auxiliary randomness
+ *                         as per BIP-340.
+ */
+SECP256K1_API int secp256k1_ecdsa_adaptor_encrypt(
```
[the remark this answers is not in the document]
jesseposner: Great point. I didn't realize there is already some adaptor signature code in the MuSig module. I will update the terminology to make it consistent with how adaptor signatures are currently described in the code base.

src/modules/ecdsa_adaptor/main_impl.h
jesseposner: The idea behind `secp256k1_ecdsa_adaptor_sig_verify` is that it encapsulates the shared code from [secp256k1_ecdsa_sig_verify](https://github.com/ElementsProject/secp256k1-zkp/blob/cfac088e1b2cd3e4a9c06cca3f37ee21bbfd3ab9/src/ecdsa_impl.h#L207), to make it easier to compare to non-adaptor ECDSA verification. This is also the case for `secp256k1_ecdsa_adaptor_sig_sign` which corresponds to [secp256k1_ecdsa_sig_sign](https://github.com/ElementsProject/secp256k1-zkp/blob/cfac088e1b2cd3e4a9c06cca3f37ee21bbfd3ab9/src/ecdsa_impl.h#L277).

I'm not sure how useful that is though, and perhaps it's better to inline and denote the shared code with a comment instead of pulling it out into a function (or just inline it without any comment about the shared code)?

jesseposner:
Thanks for the review @LLFourn, much appreciated!

src/modules/ecdsa_adaptor/main_impl.h
LLFourn: > I'm not sure how useful that is though, and perhaps it's better to inline and denote the shared code with a comment instead of pulling it out into a function

Yes I think this would be better :)

include/secp256k1_ecdsa_adaptor.h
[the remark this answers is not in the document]
jonasnick: Fwiw, I'd be fine with leaving it as is. If you rename it, then users have to remap between spec and code which seems like the more immediate problem. If you do rename, perhaps best to mention in the function doc to which algorithm in spec it corresponds to.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+    secp256k1_ge_set_gej(&enckey_expected_ge, &enckey_expected_gej);
+    /* We declassify non-secret enckey_expected_ge to allow using it as a
+     * branch point. */
+    secp256k1_declassify(ctx, &enckey_expected_ge, sizeof(enckey_expected_ge));
+    secp256k1_pubkey_save(&enckey_expected, &enckey_expected_ge);
+    /* try Y_implied == Y */
+    if (memcmp(&enckey_expected, enckey, sizeof(enckey_expected)) != 0) {
+        /* try Y_implied == -Y */
+        secp256k1_scalar_negate(&deckey, &deckey);
+        secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &enckey_expected_gej, &deckey);
```
jonasnick, line 374: I think you can just negate the existing enckey_expected_ge instead of recomputing it which would be way more efficient.

src/modules/ecdsa_adaptor/tests_impl.h
```diff
@@ -0,0 +1,736 @@
...
+    /* Check that a bitflip in an argument results in different nonces. */
+    args[0] = msg;
+    args[1] = key;
+    args[2] = pk;
+    args[3] = algo16;
+    args[4] = aux_rand;
+    for (i = 0; i < count; i++) {
+        nonce_function_ecdsa_adaptor_bitflip(args, 0, 32);
+        nonce_function_ecdsa_adaptor_bitflip(args, 1, 32);
+        nonce_function_ecdsa_adaptor_bitflip(args, 2, 32);
```
jonasnick, line 501: Need to switch to 33 here too (or just `sizeof(pk)`)

src/modules/ecdsa_adaptor/tests_impl.h
```diff
@@ -0,0 +1,736 @@
...
+ * into the hash functions. */
+void ecdsa_adaptor_test_sha256_eq(const secp256k1_sha256 *sha1, const secp256k1_sha256 *sha2) {
+    /* Is buffer fully consumed? */
+    CHECK((sha1->bytes & 0x3F) == 0);
+
+    CHECK(sha1->bytes == sha2->bytes);
+    CHECK(secp256k1_memcmp_var(sha1->s, sha2->s, sizeof(sha1->s)) == 0);
+}
+
+void run_nonce_function_ecdsa_adaptor_tests(void) {
```
jonasnick, line 451: Ideally, there would be similar tests for the `dleq_nonce` function

src/valgrind_ctime_test.c
```diff
@@ -181,6 +185,46 @@ int main(void) {
...
+        unsigned char expected_deckey[32];
+        secp256k1_pubkey enckey;
+
+        for (i = 0; i < 32; i++) {
+            deckey[i] = i + 2;
+        }
+
+        ret = secp256k1_ec_pubkey_create(ctx, &enckey, deckey);
+        VALGRIND_MAKE_MEM_DEFINED(&enckey, sizeof(secp256k1_pubkey));
+        VALGRIND_MAKE_MEM_DEFINED(&ret, sizeof(ret));
```
jonasnick, line 201: Both of these should be unnecessary. The first one is not, in fact due to a bug that [we're fixing upstream](https://github.com/bitcoin-core/secp256k1/pull/894). If you don't want to wait for the upstream PR to land in secp-zkp, you can just cherry-pick my commit there.

jonasnick (commented):
Very impressed how you figured out the peculiarities of this lib (including ctime tests, api tests, etc). PR is easy to follow and in a good state.

I didn't check whether the code exactly matches the spec and I only skimmed the tests.

As you can see [here](https://nickler.ninja/files/coverage.src_modules_ecdsa_adaptor_dleq_impl.h.html) and [here](https://nickler.ninja/files/coverage.src_modules_ecdsa_adaptor_main_impl.h.html), the tests don't achieve full coverage yet. Generally we try to do that, perhaps you want to have another look. You can generate these coverage htmls with

```
./configure --enable-module-ecdsa-adaptor --enable-experimental --enable-coverage --without-asm
# rebuild
make -B
./tests
gcov src/tests-tests.gcda
gcovr -r . --html --html-details -o coverage.html
```

jesseposner:
Thanks @jonasnick, this is super helpful! I will start updating the PR next week to respond to the comments.

include/secp256k1_ecdsa_adaptor.h
```diff
@@ -0,0 +1,156 @@
...
+ *
+ *  Returns: 1 if a nonce was successfully generated. 0 will cause signing to
+ *           return an error.
+ *  Out:     nonce32:   pointer to a 32-byte array to be filled by the function.
+ *  In:        msg32:   the 32-byte message hash being verified (will not be null)
+ *             key32:   pointer to a 32-byte secret key (will not be null)
+ *              pk33:   the 33-byte serialized pubkey corresponding to key32
+ *                      (will not be null)
+ *            algo16:   pointer to a 16-byte array describing the signature
+ *                      algorithm (will not be null).
```
jonasnick, line 33: The [schnorrsig API overhaul PR](https://github.com/bitcoin-core/secp256k1/pull/844) adds an `algolen` argument to the bip340 nonce function which gets rid of the NUL byte weirdness (initially we didn't add arguments because we wanted to have compatibility with the original nonce function). I think it would be best to do the same thing here.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(sig != NULL);
+    ARG_CHECK(deckey32 != NULL);
+    ARG_CHECK(adaptor_sig162 != NULL);
+
+    secp256k1_scalar_set_b32(&deckey, deckey32, &overflow);
+    ret &= !overflow;
+    ret &= secp256k1_ecdsa_adaptor_sig_deserialize(NULL, &sigr, NULL, &sp, NULL, NULL, adaptor_sig162);
+    secp256k1_scalar_inverse(&s, &deckey);
```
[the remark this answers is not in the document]
LLFourn: In theory, decryption key is your secret so it was either generated by you or you recovered it from another adaptor signature so it won't be zero.
In practice, I guess so.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+    }
+    secp256k1_ecdsa_signature_load(ctx, &r, &s, sig);
+    /* We declassify non-secret r to allow using it as a branch point. */
+    secp256k1_declassify(ctx, &r, sizeof(r));
+    /* Check that we're not looking at some unrelated signature */
+    if (!secp256k1_scalar_eq(&adaptor_sigr, &r)) {
+        return 0;
+    }
+    /* y = s⁻¹ * s' */
+    secp256k1_scalar_inverse(&deckey, &s);
```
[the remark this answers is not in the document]
LLFourn: I actually would have expected the deserialization of a zero `s` ECDSA signature to fail but if it doesn't then yes.

LLFourn:
@jesseposner It looks like a few of the code coverage failures are in the serialization -- I think I will add some serialization test vectors to the spec to hit them.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+
+#include "include/secp256k1_ecdsa_adaptor.h"
+#include "modules/ecdsa_adaptor/dleq_impl.h"
+
+/* (R, R', s', dleq_proof) */
+static void secp256k1_ecdsa_adaptor_sig_serialize(unsigned char *adaptor_sig162, secp256k1_ge *r, secp256k1_ge *rp, const secp256k1_scalar *sp, const secp256k1_scalar *dleq_proof_e, const secp256k1_scalar *dleq_proof_s) {
+    size_t size = 33;
+
+    secp256k1_eckey_pubkey_serialize(r, adaptor_sig162, &size, 1);
+    secp256k1_eckey_pubkey_serialize(rp, &adaptor_sig162[33], &size, 1);
```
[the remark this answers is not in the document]
jesseposner: Nope; the return values should be checked.

src/modules/ecdsa_adaptor/main_impl.h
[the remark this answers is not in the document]
jesseposner: The deserialization of `s` (via `secp256k1_ecdsa_signature_load`) does not check whether it is zero.

LLFourn:
See new tests: https://github.com/discreetlogcontracts/dlcspecs/pull/114/commits/2de65349ef8777dfc13488d511cc8a80f866d03b

jesseposner:
I've added commits that address most of the comments, but still working on https://github.com/ElementsProject/secp256k1-zkp/pull/117#discussion_r571847280 and https://github.com/ElementsProject/secp256k1-zkp/pull/117#discussion_r571242768.

jesseposner:
I'm also working on additional test coverage (the coverage report is showing ~97%).

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,391 @@
...
+    secp256k1_scalar_set_b32(&sk, seckey32, &overflow);
+    ret &= !overflow;
+    ret &= !secp256k1_scalar_is_zero(&sk);
+    secp256k1_scalar_cmov(&sk, &secp256k1_scalar_one, !ret);
+    secp256k1_scalar_set_b32(&msg, msg32, NULL);
+    /* s' = k⁻¹(m + R.x * x) */
+    ret &= secp256k1_ecdsa_adaptor_sig_sign(&sp, &msg, &k, &r, &sk);
+
+    /* return (R, R', s', dleq_proof) */
+    secp256k1_ecdsa_adaptor_sig_serialize(adaptor_sig162, &r, &rp, &sp, &dleq_proof_e, &dleq_proof_s);
```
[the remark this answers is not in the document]
jesseposner: Good catch, those should be initialized, otherwise the serialize function will read from uninitialized memory if `dleq_prove` fails to initialize them.

jesseposner:
I've added some [multi-hop lock tests](https://github.com/ElementsProject/secp256k1-zkp/pull/117/commits/5a11d5563529a7a53e49f54605596874c8ff64bb) and I believe all the comments have now been addressed.

@jonasnick @LLFourn This is now ready for another review when y'all get a chance. 🙏

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,360 @@
...
+    ARG_CHECK(deckey32 != NULL);
+    ARG_CHECK(adaptor_sig162 != NULL);
+
+    secp256k1_scalar_set_b32(&deckey, deckey32, &overflow);
+    ret &= !overflow;
+    ret &= secp256k1_ecdsa_adaptor_sig_deserialize(NULL, &sigr, NULL, &sp, NULL, NULL, adaptor_sig162);
+    ret &= !secp256k1_scalar_is_zero(&deckey);
+    secp256k1_scalar_inverse(&s, &deckey);
+    /* s = s' * y⁻¹ */
+    secp256k1_scalar_mul(&s, &s, &sp);
```
jonasnick, line 294: My gcc (v 10.2) warns me that sp may be uninitialized here, which may happen if `sigr` is `0` and `sig_deserialize` fails.

jonasnick (commented):
Ah multi-hop locks test... nice! Looks good to me besides that uninitialized var. Feel free to squash.

LLFourn (approved):
utACK fc14f40c87548d0dbbadfd5182257a2175ffd4cc. 

In so far as I can review c-code it looks really good. Specifically I've just looked at:

- Core cyrpto and serialization is same as my implementation I used to produce spec.
- Tests data are all as in spec.
- Test assertions cover everything intended.

I'm not sure how to easily confirm the tests are running in CI though. If they are then let's merge!

jesseposner:
I (1) removed the "work in progress" warning and (2) [initialized sp](https://github.com/ElementsProject/secp256k1-zkp/pull/117/files#diff-f4614e8f5280013c04ae54758be90e65dbab41f8441bfdd9bdddc5d31a6d0bf5R288) ([diff](https://github.com/ElementsProject/secp256k1-zkp/compare/fc14f40c87548d0dbbadfd5182257a2175ffd4cc..54985c2b87f9e361b951f929fad6f5b89f126d1b)). I also squashed the commits.

jesseposner:
d241ad0 is cherry-picked from https://github.com/bitcoin-core/secp256k1/pull/894/commits/7d3497cdc4c747bdd51db70f42fe218622c3169f

jonasnick (commented):
sorry, final nit: can you prefix the commit messages with something like "ecdsa_adaptor:"? That way it's immediately obvious to which module a commit belongs to (as opposed to just "Initialize project").

apoelstra:
This looks great! I haven't yet run the tests but I reviewed all the commits. The only nit I have is that in 41df7700b05f5f498b698119f297bf408333ecba in `secp256k1_ecdsa_adaptor_recover` you use `memcmp` on `secp256k1_ge` structs in a couple places. I am worried that these may fail supriously due to padding.

The protocol is really neat. I'm afraid I let my anti-ECDSA bias get in the way of studying this earlier. I didn't realize that it could be done without any 2p ECDSA. To check my understanding, is this roughly how an atomic swap would work between two parties Alice and Bob?:

1. The two parties put their respective coins in 2-of-2 `CHECKMULTISIG` outputs, sign timelocked backout transactions, etc
2. Alice chooses a lock keypair `(r, R)` and gives `R` to Bob, as well as an adaptor signature encrypting `r` on a transaction giving Bob his coins.
3. Bob then creates an adaptor signature encrypting `r` to Alice, on a tx that gives her her coins. **Bob does not know `r` at this point, nor does he know his own final signature** but this is fine.
4. Alice, using `r` which she knows, decrypts Bob's signature, adds her own sig, and publishes a transaction taking her own coins.
5. Bob learns `r` by reading his own signature, which he previously didn't know.

Then to extend to the multi-lock case we replace the lock `R` with a "left lock" and "right lock" which differ by some secret value known to Bob, so he can effectively reblind the lock when connecting his incoming and outgoing payment channels.

Unless I'm confused, this is amazing!

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,361 @@
...
+
+    /* Deal with ECDSA malleability */
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &enckey_expected_gej, &deckey);
+    secp256k1_ge_set_gej(&enckey_expected_ge, &enckey_expected_gej);
+    /* We declassify non-secret enckey_expected_ge to allow using it as a
+     * branch point. */
+    secp256k1_declassify(ctx, &enckey_expected_ge, sizeof(enckey_expected_ge));
+    secp256k1_pubkey_save(&enckey_expected, &enckey_expected_ge);
+    /* try Y_implied == Y */
+    if (memcmp(&enckey_expected, enckey, sizeof(enckey_expected)) != 0) {
```
real-or-random, line 343: This should be `secp256k1_memcmp_var`

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -0,0 +1,361 @@
...
+        /* try Y_implied == -Y */
+        secp256k1_ge_neg(&enckey_expected_ge, &enckey_expected_ge);
+        secp256k1_pubkey_save(&enckey_expected, &enckey_expected_ge);
+        if (memcmp(&enckey_expected, enckey, sizeof(enckey_expected)) != 0) {
```
real-or-random, line 347: same here

src/modules/ecdsa_adaptor/dleq_impl.h
```diff
@@ -0,0 +1,158 @@
+#ifndef _SECP256K1_DLEQ_IMPL_H_
```
real-or-random, line 1:
```suggestion
#ifndef SECP256K1_DLEQ_IMPL_H
```
(identifiers starting with `_` and an uppercase letter are reserved in C)

real-or-random (commented):
(I didn't do a proper review but @apoelstra's mention of `memcmp` caught my attention.)

jesseposner:
Thanks for the feedback @apoelstra and @real-or-random! I will push some fixes later today.

@apoelstra Your description of the single and multi-hop atomic-swap looks correct to me. The multi-hop protocol with proof of payment described [here](https://github.com/ElementsProject/scriptless-scripts/blob/master/md/multi-hop-locks.md) works as well, and there's a [test case](https://github.com/ElementsProject/secp256k1-zkp/blob/d0ba0e21a677465b114c81c1a7472304adb392f8/src/modules/ecdsa_adaptor/tests_impl.h#L1086) that implements it.

I was also quite surprised to see how nicely this protocol works with ECDSA once you drop the requirement for scriptless 2p!

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -91,4 +146,221 @@ static int nonce_function_ecdsa_adaptor(unsigned char *nonce32, const unsigned c
...
+    secp256k1_scalar_inverse(&deckey, &s);
+    secp256k1_scalar_mul(&deckey, &deckey, &sp);
+
+    /* Deal with ECDSA malleability */
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &enckey_expected_gej, &deckey);
+    secp256k1_ge_set_gej(&enckey_expected_ge, &enckey_expected_gej);
+    /* We declassify non-secret enckey_expected_ge to allow using it as a
+     * branch point. */
+    secp256k1_declassify(ctx, &enckey_expected_ge, sizeof(enckey_expected_ge));
+    if (!secp256k1_eckey_pubkey_serialize(&enckey_expected_ge, enckey_expected33, &size, 1)) {
```
jesseposner, line 343: The calls to `secp256k1_eckey_pubkey_serialize` on 343 and `secp256k1_ec_pubkey_serialize` on 346 were just [added](https://github.com/ElementsProject/secp256k1-zkp/compare/d0ba0e21a677465b114c81c1a7472304adb392f8..431b602b81693974d2e3a424681fe62262b7a700). Test coverage for the error cases will be added soon.

jesseposner:
I've made 4 changes in response to the latest comments. Tracking these changes is a little tricky due to rebasing, so I'm listing them here:

- [(1) Remove underscore prefix and suffix from identifier and (2) serialize `enckey` structs for comparison](https://github.com/ElementsProject/secp256k1-zkp/compare/d0ba0e21a677465b114c81c1a7472304adb392f8..431b602b81693974d2e3a424681fe62262b7a700)
- [Optimize `enckey` serialization comparison](https://github.com/ElementsProject/secp256k1-zkp/compare/431b602b81693974d2e3a424681fe62262b7a700..4d10284a1100ff510b377db09b1c98f6454e4fe0)
- [(1) Make `target` `const` and (2) add test coverage for `enckey` serialization](https://github.com/ElementsProject/secp256k1-zkp/compare/4d10284a1100ff510b377db09b1c98f6454e4fe0..2ae65d55d8acbeefb359bbede42cac6bd5966e60)
- [(1) Use `SECP256K1_EC_COMPRESSED` instead of `1`, (2) add API tests for `secp256k1_ecdsa_adaptor_decrypt` and `secp256k1_ecdsa_adaptor_recover`, and (3) refactor `enckey` serialization tests](https://github.com/ElementsProject/secp256k1-zkp/compare/2ae65d55d8acbeefb359bbede42cac6bd5966e60..5fbed0ac5c97cf24a737dfb23906aac2cf996c3e)

apoelstra:
ack 2f6c8bb13fb87602af6e7be2138f4c2e62c0d7d4

I'm comfortable merging this into -zkp, but I'm not really convinced by the security argument for this scheme being EUF-CMA[VES]. My expectation is that the scheme will need to be tweaked somehow to get provable security.

jonasnick (commented):
Sorry @jesseposner, we synced upstream and you have to rebase. This will allow removing the ctime_test commit (because we just have it from upstream now). And it includes the switch to cirrus, which means that you have to apply pretty much the same changes to cirrus.yml as you did to travis.yml before.

> Optimize enckey serialization comparison

As far as I can tell the previous version was fine because you were comparing pubkey structs, not group elements. Now that you serialize the point it would be simpler to just compare them like this:

```
    if (secp256k1_memcmp_var(&enckey_expected33[1], &enckey33[1], 32) != 0) {
        return 0;
    }
    if (enckey_expected33[0] != enckey33[0]) {
        /* try Y_implied == -Y */
        secp256k1_scalar_negate(&deckey, &deckey);
    }
```

Otherwise your changes look good and I can merge this after the rebase.

apoelstra:
@LLFourn on further reflection, I don't believe your "it's impossible to prove this EUF-CMA[VES] secure" argument :). You describe a CDH attacker that receives a CDH challenge (X, Y) then uses X as a public key when interacting with a EUF-CMA[VES] challenger ... but this is invalid, because the challenger chooses the public key!

LLFourn:
> 
> 
> @LLFourn on further reflection, I don't believe your "it's impossible to prove this EUF-CMA[VES] secure" argument :). You describe a CDH attacker that receives a CDH challenge (X, Y) then uses X as a public key when interacting with a EUF-CMA[VES] challenger ... but this is invalid, because the challenger chooses the public key!

@apoelstra This proof is bit awkward. I actually removed it in an unfinished revision I was working on[1]. Here's the explanation anyway: In this proof the EUF-CMA[VES] challenger is a DL attacker just like the EUF-CMA challenger is a DL attacker in a Schnorr EUF-CMA reduction. Just as in the Schnorr EUF-CMA the DL attacker chooses the public key. But we show this reduction cannot exist from DL to EUF-CMA[VES] since the necessary simulator algorithm for the signature encryption oracle doesn't exist if CDH is hard (by interacting with such a simulator we can solve arbitrary CDH so it can't exist). 

The reason I don't like this proof is that it makes the flaw seem like a theoretical point. It isn't. **you can literally recover the DH key from any ECDSA adaptor signature** (the formula is in the secitoin but it is the same as the last line of the proof). It might even be worth adding as a function to the library so that it's super obvious that is has this flaw! There is no way this can fit EUF-CMA[VES] because of this observation.

> I'm comfortable merging this into -zkp, but I'm not really convinced by the security argument for this scheme being EUF-CMA[VES]. My expectation is that the scheme will need to be tweaked somehow to get provable security.

Thanks for bringing up the security of this scheme. In terms of provable security my "CDH easy" proof for ECDSA EUF-CMA[VES] in the master version (appendix A) is wrong (there is a note there). It turns out you can't make CDH easy without making DL itself easy! In [1- Appendix A] I fixed that by offering a more restricted "static" CDH oracle and re-working the reduction. The proof is at least satisfying for me. Fortunately there is another take on the security of the ECDSA scheme in [2] which shows EUF-CMA[VES] security (under the name of aEUF–CMA) if a proof of knowledge is given of the decryption key. This is obviously consistent with my result since if you have knowledge of the decryption key then you can already generate the DH key between the public signing key and encryption key.

In any case this scheme is dangerous and should be applied very carefully because of the DH key problem.  One solution is to do as in [1] and enforce a PoK on any encryption key which renders it unconditionally secure under DL. Unfortunately this would break our use in DLCs where it would be inefficient to have an explicit PoK on each encryption key because each encryption key is derived from a public key, nonce and an outcome index given by the oracle (we can easily argue the oracle "knows" each of the derived decryption keys though).

I suggest adding a healthy warning about the flaw and implementing the function to recover the DH key to fully expose the problem. It is useful to point out as well the only published security claim relies on there being a proof of knowledge on the encryption key from the party receiving the encrypted signature (which is consistent with how we are using it in DLCs).

[1] https://github.com/LLFourn/one-time-VES/blob/dev/main.pdf
[2] https://eprint.iacr.org/2020/476.pdf

apoelstra:
> Just as in the Schnorr EUF-CMA reduction the DL attacker chooses the public key

Yes but the DL attacker is not the Schnorr attacker :)

> The reason I don't like this proof is that it makes the flaw seem like a theoretical point. It isn't. you can literally recover the DH key from any ECDSA adaptor signature (the formula is in the secitoin but it is the same as the last line of the proof). It might even be worth adding as a function to the library so that it's super obvious that is has this flaw!

You can only recover the DH key if you know one of the two secret keys involved. And there is already a function in libsecp for this :). It is not a flaw.

LLFourn:
> 
> 
> > Just as in the Schnorr EUF-CMA reduction the DL attacker chooses the public key
> 
> Yes but the DL attacker is not the Schnorr attacker :)

I do not claim that it is (I think).

The DL attacker uses the Schnorr attacker to solve DL. We have a specific example of such a DL attacker that simulates the EUF-CMA signature oracle by programming their random oracle (and another in GGM).
My claim is that there cannot exist a DL attacker that uses a ECDSA EUF-CMA[VES] attacker to do the same since a simulator for the encrypted signature oracle cannot exist if the CDH problem is hard (in any model).

> 
> > The reason I don't like this proof is that it makes the flaw seem like a theoretical point. It isn't. you can literally recover the DH key from any ECDSA adaptor signature (the formula is in the secitoin but it is the same as the last line of the proof). It might even be worth adding as a function to the library so that it's super obvious that is has this flaw!
> 
> You can only recover the DH key if you know one of the two secret keys involved. And there is already a function in libsecp for this :). It is not a flaw.

You are missing my point. You can recover the DH key between X and Y from `(X,Y, <ECDSA encrypted signature signed by X encrypted by Y>)`.  This is only public information (no secret keys needed) and is certainly a flaw. The function to do so is easy to implement but not in this library. This is the formula using notation from the one-time VES paper:

image

This is why ECDSA encrypted signatures cannot be simulated regardless of the model if CDH is hard.

apoelstra:
Ah, I did misunderstand which ECDH pair we were talking about. I agree that this is surprising and means that users should not reuse an ECDH key as a signing key in an adaptor-signature-based protocol. (Of course, it is generally good advice to avoid reusing ECDH keys with anything else!)

However I still disagree that this prevents proving EUF-CMA[VES] for ECDSA adaptor sigs in a hard-CDH setting.

You have three parties: an CDH challenger, a EUF-CMA[VES] challenger, and a forger. The CDH challenger provides a pair of points (A,B). The EUF-CMA[VES] challenger provides a signing key X. The forger would like to get the DH point of (A,B), but as near as I can tell, he is only able to obtain (X,A) or (X,B) ... or more generally, (X,P) for anything else. And this is no problem at all.

jonasnick (commented):
The rebased state of the PR looks good to me. I pushed a [commit to my branch](https://github.com/jonasnick/secp256k1-zkp/commits/ecdsa-adaptor-sigs-jn) that adds a WARNING about the DH key leak to the include file. @jesseposner feel free to cherry-pick. @llfourn is it accurate? Adding a function to extract the DH key would be fun but I don't think it would help much.

real-or-random:
@LLFourn 

I was trying to understand Lemma 4.3 but I couldn't follow. Here are some specific questions:
 - When you say "a reduction for the ECDSA one-time VES scheme", what is it supposed to reduce to? To the EUF-CMA security of ECDSA? If yes, EUF-CMA security of ECDSA in which model? (Is H an algorithm a random oracle or a (bidirectional) random oracle?)
 - What's a "key-preserving" reduction?
 - I don't understand why you're allowed to assume the existence of S. Is this related to the two points above? So I get that the lemma supposes that some reduction exists and it must implement the E-oracle. But I don't see why this must be done via a *standalone* algorithm S. In general, couldn't S be stateful and interact with the reduction arbitrarily? Also in which model is the reduction supposed to work? (For example, is H an algorithm or is H an oracle that the reduction/S may program?)

LLFourn:
@real-or-random 

> * When you say "a reduction for the ECDSA one-time VES scheme", what is it supposed to reduce to? To the EUF-CMA security of ECDSA? If yes, EUF-CMA security of ECDSA in which model? (Is H an algorithm a random oracle or a (bidirectional) random oracle?)
>

Oops looks like I forgot to specify that. It is a reduction from DL to EUF-CMA[VES] security of ECDSA adaptor sig scheme (EUF-CMA is not involved). My claim is that it can't exist in any model -- this might be a bit ambitious (see below). But I suppose that it could have equivalently be for EUF-CMA of ECDSA to EUF-CMA[VES].  As long as it's a game where you can challenge with a key and the reduction is "key-preserving" (maybe that's why I left the particular problem out). Looking back it seems that proving things with respect to DL instead of to key-only or EUF-CMA is a mistake I repeated in this paper several times. I can't really recall why I did things this way.


> * What's a "key-preserving" reduction?
>

In context of signature security reductions it means that the reduction challenges the forger with key it was challenged. E.g. a DL -> key-only Schnorr reduction passes through the group element it is challenged with to the Schnorr forger.

> * I don't understand why you're allowed to assume the existence of S. Is this related to the two points above? So I get that the lemma supposes that some reduction exists and it must implement the E-oracle. But I don't see why this must be done via a _standalone_ algorithm S. In general, couldn't S be stateful and interact with the reduction arbitrarily? Also in which model is the reduction supposed to work? (For example, is H an algorithm or is H an oracle that the reduction/S may program?)

This is an interesting point and one I thought about a lot. Whether a simulator correctly simulates something depends on the other queries (e.g. like ROM queries) being answered in a way that is consistent with them. If you are careful I think they can be treated as standalone algorithms that are called by the reduction and return something to the reduction. I don't believe this point is necessary to explore for the particular lemma in question though.

Let me try and salvage the lemma somewhat in a way that will draw out more clearly any disagreement.

Here's why the bijective random oracle (BiROM) model reduction _cannot_ exist from DL to EUF-CMA[VES] security of ECDSA one-times VES/adaptor signatures.

1. Let's call **A** the set of efficient algorithms that when activated by an instance of the DL problem respond to EUF-CMA[VES] E oracle queries and BiROM queries in a consistent way (the E responses are valid with respect to the BiROM responses). These algorithms need not return an DL solution -- they just have to answer the oracle queries (in a key-preserving way).
2. Clearly any reduction from DL to EUF-CMA[VES] security of the ECDSA scheme can be made part of the set by ignoring the reduction's final output.
3. **A** must be empty since any algorithm in **A** can be used to solve CDH(X,Y).  You just activate the algorithm with X and then query Y as the encryption key for an E query and use the DH key extraction algorithm.
4. Since **A** is empty if CDH is hard the set of reductions from DL to EUF-CMA[VES] in the BiROM is empty too.

Hopefully that makes sense. This should hold true in any model (BiROM oracles can be replaced with any set of oracles) . Although now I think about it, it would be very awkward to define in the generic group model where the reduction controls the group representation. So I think I agree with you that I should have stated which model in particular (the usefulness of the lemma is not really reduced by proving only for BiROM).

@apoelstra 
> Ah, I did misunderstand which ECDH pair we were talking about. I agree that this is surprising and means that users should not reuse an ECDH key as a signing key in an adaptor-signature-based protocol. (Of course, it is generally good advice to avoid reusing ECDH keys with anything else!)

On this topic I believe it could be useful to use node public keys (which are used in noise ECDH) in a funding output's actual script pubkey. The idea is that you can find your channels with any peer without a backup just by doing ECDH and deterministically deriving them. See the idea here:

https://lists.linuxfoundation.org/pipermail/lightning-dev/2020-December/002907.html

I believe this is likely to be included in the updated lightning funding protocol proposal so register your objections before it's too late!

> However I still disagree that this prevents proving EUF-CMA[VES] for ECDSA adaptor sigs in a hard-CDH setting.
> 
> You have three parties: an CDH challenger, a EUF-CMA[VES] challenger, and a forger. The CDH challenger provides a pair of points (A,B). The EUF-CMA[VES] challenger provides a signing key X. The forger would like to get the DH point of (A,B), but as near as I can tell, he is only able to obtain (X,A) or (X,B) ... or more generally, (X,P) for anything else. And this is no problem at all.

Hopefully the above also answers most of this. The claim follows a "meta-reduction" like claim where we extract the DH key by playing both the challenger to the reduction and the forger the reduction is using.

Your point is good though. From a encrypted signature oracle you can only get the DH keys for (X,A), (X,B) etc. Realizing the importance of this was the reason that I started working on the revision where in the actual proof of security I give the reduction a "static" CDH oracle with a fixed X (in the original they were able to query for any CDH solution). I stopped working on it because I didn't think anyone would ever care about this and that [1] would be sufficient (yet here we are!).

> And this is no problem at all.

It turns out this **is a problem**! If you can query the DH keys against X you can solve discrete log for X faster than without such an oracle. I even did concrete estimates of how much faster in [2] for secp256k1 (see Appendix B). So in the end ECDSA adaptor signatures are less secure than ECDSA itself even if you don't insecurely compose it with a scheme using Diffie-Hellman!

[1] https://eprint.iacr.org/2004/306
[2] https://github.com/LLFourn/one-time-VES/blob/dev/main.pdf

@jonasnick 
> @LLFourn is it accurate? 

Yes LGTM.

include/secp256k1_ecdsa_adaptor.h
```diff
@@ -10,7 +10,24 @@ extern "C" {
  *  Lloyd Fournier
  *  (https://lists.linuxfoundation.org/pipermail/lightning-dev/2019-November/002316.html
  *  and https://github.com/LLFourn/one-time-VES/blob/master/main.pdf).
-*/
+ *
+ *  WARNING! DANGER AHEAD!
+ *  As mentioned in Lloyd Fournier's paper, the adaptor signature leaks the
+ *  Diffie-Hellmann key between the signing key and the encryption key. This is
```
real-or-random, line 16:
```suggestion
 *  Diffie-Hellman key between the signing key and the encryption key. This is
```

include/secp256k1_ecdsa_adaptor.h
```diff
@@ -10,7 +10,24 @@ extern "C" {
...
+ *  not a problem for ECDSA adaptor signatures themselves, but may result in a
+ *  complete loss of security when they are composed with other schemes. More
+ *  specifically, let us refer to the signer's public key as X = x*G, and to the
+ *  encryption key as Y = y*G. Given X, Y and the adaptor signature, it is
+ *  trivial to compute Y^x = X^y.
+ *  To completely avoid this problem, the signer can require a proof of
+ *  knowledge of the decryption key y before creating an adaptor signature. This
```
real-or-random, line 23: Maybe explain how this proof of knowledge can be implemented.

On a second thought, I'm not sure how. It's formally not known to be a proof of knowledge but is an ECDSA signature enough here? Maybe recommend a Schnorr signature to be on the safe side? We have it in the library now, so why not.

real-or-random:
@LLFourn Ah okay, I think I understand now! So you give a meta-reduction even against any approach simulating the EUF-CMA[VES] game towards the adversary (without caring about solving DL in the end). 

Here's some further thoughts (really not relevant to this PR, feel free to ignore): 
I think the restriction to "key-preserving" reductions makes this somewhat limited. For example, a reduction that receives challenge xG and passes (x+1)G as public key is not captured by the lemma. Here's a vague intuition on how to do this differently: I think what one would like to capture on an intuitive level is "reductions that don't know the DL of the public key". One way to formalize this to write a meta-reduction on DL which plays the adversary and extracts the DL of the public key from the reduction. So the meta-reduction would (ab)use the reduction to solve the DL problem. But that seems hard because it seems that the meta-reduction can only extract CDH solutions and not DLs. What could work instead is to restrict the lemma to algebraic reductions: So the reduction itself needs to interact with the group like in the AGM, i.e., whenever it outputs a group element it must send along a representation of this group element in the base that consists of all group elements it has received thus far. Since CDH and DL are equivalent in the AGM (see for example https://eprint.iacr.org/2018/1220.pdf), the meta-reduction could extract DLs. The resulting lemma  would cover a much larger and very natural class of reductions.

LLFourn:
@real-or-random Good insight. I agree an algebraic reduction would probably work here instead and be a more compelling claim. It seems like I would be able to extract the CDH solution regardless of the reduction being key-preserving in that case.

apoelstra:
Yeah, this description (and @real-or-random's summary) was very helpful for me.

include/secp256k1_ecdsa_adaptor.h
jesseposner: As far as I can tell, there are at least a few pitfalls with a Schnorr signature proof of knowledge. Firstly, a naive construction is replayable. For example, most of the schemes I've seen essentially just sign the generator (`c = H(g, y, t)` from https://en.wikipedia.org/wiki/Fiat%E2%80%93Shamir_heuristic and `c = H(g, h, u)` from https://crypto.stanford.edu/cs355/19sp/lec5.pdf). This can perhaps be resolved by requiring the verifier to send some random data to the prover to be used as the message. However, this opens a second potential attack, which is to trick the prover into signing a message, for example by sending "random data" that is in fact a transaction or some other non-random value. This second issue could be potentially resolved by having the prover hash the random data (and the prover should also probably include the nonce pubkey in the hash so that there's an input to the hash function which is not controlled by the verifier). So I think a scheme where (1) the verifier sends the prover some random data and (2) the prover hashes the random data along with the nonce pubkey and signs the hash, could be okay, but curious to hear what other folks think of the scheme.

Also, maybe suggesting a proof of knowledge scheme here is more trouble than it's worth because if the recipient of the adaptor signature knows the decryption key, then they can simply decrypt the adaptor signature when they receive it which seems like it would obviate the usefulness of encrypting the signature.

cc @jonasnick @real-or-random @LLFourn

jonasnick: Hm, I hadn't considered that the recipient doesn't actually know the decryption key. Perhaps it would be okay in practice if before creating an adaptor signature for `enckey`, the signer would require a a signature for `enckey` and a message like "this is an adaptor sig enckey". Thus, whoever creates `enckey`s must also provide such a signature and thereby communicates that they don't care about the DH key between the signing and the `enckey`. The idea is that keys whose DH key with the signing key must not be leaked (such as an ephemeral ElGamal key) will not be used to create such signatures. However, this approach does not work for multi-hop-locks because there is no party that on its own knows the DL of the enckey when exchanging adaptor sigs to establish the route.

real-or-random: @jesseposner I think the canonical way would be a signature on `H(m)` where `m` is a random challenge message  and `H` is a tagged hash with some unique tag such as `ECDSA adaptor PoK`. 

But I agree, the proof of knowledge seems not useful in this setting. 

Note that I also didn't notice this because my comments here are pretty superficial. You guys have a better understanding of the entire thing. 

Maybe warning like this is just fine: "A weaker defense is to simply not compose ECDSA adaptor signatures with other protocols."

Could we make that more specific?  "A defense is to not reuse the signing key of ECDSA adaptor signatures in other protocols." or "in the protocols that rely on the hardness of the CDH problem, e.g., Diffie-Hellman key exchange."

include/secp256k1_ecdsa_adaptor.h
real-or-random: Mabye also mention the acronym ECDH here to make sure everyone gets this (because that's the name of the corresponding library module).

include/secp256k1_ecdsa_adaptor.h
jesseposner: I updated the [warning description](https://github.com/ElementsProject/secp256k1-zkp/pull/117/files#diff-0bc5e1a03ce026e8fea9bfb91a5334cc545fbd7ba78ad83ae5489b52e4e48856R14).

LLFourn: > Also, maybe suggesting a proof of knowledge scheme here is more trouble than it's worth because if the recipient of the adaptor signature knows the decryption key, then they can simply decrypt the adaptor signature when they receive it which seems like it would obviate the usefulness of encrypting the signature.

Note the point is that the encryption is "one-time" in that it leaks the decryption key. The point of giving an adaptor signature to someone who knows the decryption key is precisely because you hope they will leak it by decrypting and publishing the signature. DLCs are an exception to this.

There was another attempt at improving terminology here by generalizing this to "lockable" signatures (still in my list of things to read):  https://eprint.iacr.org/2020/1613.pdf. Perhaps a "locked" signature is more intuitive than an "encrypted" signature. We already use the term "lock" in the sense of multi-hop locks so this might make things easier for people. Impressively they claim to be able to do it with BLS which I had thought was impossible to do efficiently.

> However, this approach does not work for multi-hop-locks because there is no party that on its own knows the DL of the enckey when exchanging adaptor sigs to establish the route.

Yep. Nothing much you can do here other than avoid CDH or direct people to using Schnorr on signet until it arrives on mainnet later this year 🤞 .

jonasnick (commented):
The updated WARNING section looks good to me.

ACK b0ffa923199c45f717adf3fb003bcef796259032

include/secp256k1_ecdsa_adaptor.h
jesseposner: >Note the point is that the encryption is "one-time" in that it leaks the decryption key. The point of giving an adaptor signature to someone who knows the decryption key is precisely because you hope they will leak it by decrypting and publishing the signature. DLCs are an exception to this.

Great point, for example, the last hop in a multi-hop payment will know the decryption key for the adaptor signature it receives. However, each hop before it will not know the decryption key when they receive their adaptor signatures (but they will of course learn the respective decryption keys during the settlement phase of the protocol).

>Perhaps a "locked" signature is more intuitive than an "encrypted" signature.

"Lock" is nice particularly because, as you point out, it makes the phrase "multi-hop locks" more clear, but then you need to refer to a "locking key" and an "unlocking key" and the "one-time" analogy to the OTP is lost, as is the scheme's verifiability. "One-time verifiably encrypted signatures" conveys more information to the reader about the scheme than "lockable signatures."

LLFourn:
ACK b0ffa923199c45f717adf3fb003bcef796259032 I've added a small warning to the spec too.

GeneFerneau:
@jonasnick @LLFourn I recently did an implementation of ECDSA adaptor signatures before being made aware of this work.

I did a quick review of your code, and didn't see a proof for `Y = y*g` as specified in [Generalized Bitcoin-Compatible Channels](https://eprint.iacr.org/2020/476.pdf). Aumayr et al. claim the proof is necessary for the scheme to be secure under the Universal Composability framework.

What are your thoughts on adding [such a proof](https://github.com/geneferneau/fischlin), similar to what I've done [here](https://github.com/geneferneau/adaptor/blob/main/src/ecdsa.rs)?

jonasnick:
Hi Gene, thanks for your review. Do you mean requiring a proof of knowledge for the adaptor before signing? If so, then we've spent half of the thread discussing this issue :) The PoK would prevent some applications like multi-hop locks which is why we added a warning to the top of the include file that states that the scheme is not universally composable.

GeneFerneau:
> Do you mean requiring a proof of knowledge for the adaptor before signing? If so, then we've spent half of the thread discussing this issue :)

I skimmed the conversation, the proof in the paper I linked does not prove `Y^x != X^y`. It proves that `Y = y*G` using a Schnorr identification scheme, and Fischlin proof to make the scheme non-interactive. The purpose of the proof is to prevent the pre-signature/verifier from accepting a `Y` which has no known private key (e.g. NUMS generators).

Being supplied such a `Y` would enable a malicious prover to pass pre-signature verification, and only be found out at the decrypt/recover (adapt/extract) phase. The Schnorr-Fischlin proof does not prevent multi-hop locks, as the paper "Generalized Bitcoin-Compatible Channels" is specifically targeted at such a use-case.

I do somewhat understand how proving `Y^x != X^y` would make multihop locks impossible when the prover does not have access to both private keys (needed for generating the proof). However, the proof I'm suggesting is not that, and the paper I am working from includes security proofs under the Universal Composability framework.

LLFourn:
> 
> 
> > Do you mean requiring a proof of knowledge for the adaptor before signing? If so, then we've spent half of the thread discussing this issue :)
> 
> I skimmed the conversation, the proof in the paper I linked does not prove `Y^x != X^y`. It proves that `Y = G^y` using a Schnorr identification scheme, and Fischlin proof to make the scheme non-interactive. The purpose of the proof is to prevent the pre-signature/verifier from accepting a `Y` which has no known private key (e.g. NUMS generators).

You cannot prove that Y^x != X^y since it is always true given that Y = G^y and X = G^x :)

I think you miunderstood the discussion above. The problem are aware of with this scheme is that from each adaptor signature you may extract X^y. If the party receiving the adaptor signature knows y then then this isn't an issue. That's why to get the proof of security to work in a simulation paradigm then you will need a proof of knowledge so you can simulate the adaptor signature in the proof.
> 
> Being supplied such a `Y` would enable a malicious prover to pass pre-signature verification, and only be found out at the decrypt/recover (adapt/extract) phase.

I don't understand how. Can you point out the algorithm to do this?

> The Schnorr-Fischlin proof does not prevent multi-hop locks, as the paper "Generalized Bitcoin-Compatible Channels" is specifically targeted at such a use-case.
> 
> I do somewhat understand how proving `Y^x != X^y` would make multihop locks impossible when the prover does not have access to both private keys (needed for generating the proof). However, the proof I'm suggesting is not that, and the paper I am working from includes security proofs under the Universal Composability framework.

From memory there is no discussion of doing multi-hop locks with this scheme in their paper (the multi-hop locks paper uses 2pECDSA). The proof of knowledge for Y to the base G is difficult do since the receiver of the adaptor signature cannot produce that on their own.

GeneFerneau:
> You cannot prove that Y^x != X^y since it is always true given that Y = G^y and X = G^x :)
>
> I think you miunderstood the discussion above.

Yeah, after thinking about it more, I understand now. I was confused because of the below warning:

```
Given X, Y and the adaptor signature, it is
 *  trivial to compute Y^x = X^y.
```

thinking the point was to avoid `Y^x = X^y`.

> I don't understand how. Can you point out the algorithm to do this?

If a malicious prover supplies a NUMS generator as `Y`, the pre-signing and verification will pass, since `y` (private key) is not involved in those steps. When the prover goes to decrypt (adapt) the pre-signature, they will be unable to produce a valid signature that passes as an ECDSA signature, and allows the other party to recover (extract) the witness `y`.

> From memory there is no discussion of doing multi-hop locks with this scheme in their paper

The paper directly cites "Anonymous Multi-Hop Locks for Blockchain Scalability and Interoperability", your work "One-Time Verifiably Encrypted Signatures A.K.A. Adaptor Signatures", and mentions that their ECDSA adaptor signature scheme is a formalization (under UC framework) of the adaptor signatures used by Malavolta et al. and yourself:

> Given  the  utility  of  AS,  there  have  been  some  attempts  to formally use them. For instance, Malavolta et al. [22] use AS as 
> building  block  to  define  and  realize  multi-hop  payments in  PCNs.  However,  they  do  not  define  AS  as  a  stand  alone
> primitive  that  can  be  then  used  in  other  works.  Concurrent to  our  work,  Fournier  [13]  attempts  to  formalize  AS  as 
> an instance  of  one-time  verifiable  encrypted  signatures.  Yet,  in their  definition  the  adversary  is  not  given  a  pre-signature 
> on the challenge message in the unforgeability and extractability games.  However,  in  applications  including  our
> generalized channels, the adversary learns a pre-signature on the message for which it wishes to forge a signature.

So, the 2P-ECDSA is used along with ECDSA adaptor signatures by Malavolta et al.

> The proof of knowledge for Y to the base G is difficult do since the receiver of the adaptor signature cannot produce that on their own.

The prover / pre-signature signer computes the proof of knowledge, and provides it to the verifier along with the pre-signature. I don't think I understand your point. If the verifier could produce the proof on their own, how would it be a zero-knowledge proof of knowledge?

LLFourn:
@GeneFerneau From my perspective the "generalized" channel work and my (admittedly flawed and unfinished) work own have the same conclusion about ECDSA AS and AS security in general. I simply have a stronger claim that the ECDSA construction is fine as long as you don't care about leaking DH keys (in which case you don't need the PoK). I think you can use this code for multi-hop locks etc as long as you are very careful about not composing it with some scheme that relies on DH. In other words, creating ECDSA AS oracle is like giving a DH oracle on the signing key.

FWIW I never got to the bottom of those comments from the Aumayr et al. on my definitions. To me the definitions especially related to "unforgeability" are pretty much the same.

PR #118, opened 2021-01-05
rangeproof: clarify rewind outlen argument

jonasnick:
Found this note on this in @thomaseizinger's rust-secp256k1-zkp:
```
// TODO: File bug with upstream: message length is not set correctly
```

The message length is okay but this is a doc bug.


apoelstra (approved):
ack 41d6963bc1c99081e1a0cebbd1e97a77b6d02445

apoelstra:
It's a bit of a confusing API, but I think it would be breaking to try to change it.

PR #119, opened 2021-01-06
Remove repeated schnorr flag from travis config

jesseposner:
The `SCHNORRSIG=no` flag is set twice for `global` in `travis.yml`. This PR removes the duplicated flag.


real-or-random (approved):
Thanks!

I'm sure this happened when syncing with upstream.

ACK 96c83a83dcf742bc175c079188b06dfd0622406c

PR #120, opened 2021-01-12
Add MuSig Key Aggregation spec

jonasnick:
This PR is for discussing the MuSig (Key Agg) specification. I started this by documenting what the code does currently.
In particular, the open design questions include:

- Should hash index or should hash key?
  - how to encode the index (~~right now it's least significant first~~ (EDIT: changed that) but it's most significant for every other integer)?
- Should the first MuSig coefficient be the constant 1? What would be the security argument?
- Should sort pubkeys in KeyAgg algorithm?


jonasnick:
Just realized that this does not reflect the currently implemented key aggregation because I was looking at the wrong branch (master instead of secp256k1-zkp...).

jonasnick:
Fixed

jonasnick:
We may also want to consider how this relates to signature aggregation (H/T @roconnor-blockstream)

roconnor-blockstream:
My specific consideration was whether we want this key aggregation function to be use both for key aggregation and signature aggregation.  If we want it to be usable for signature aggregation, I believe that will place some constraints on the answers to Jonas's question.

roconnor-blockstream:
If we want our Key Aggregation function to be compatible with multi-message signing (i.e. cross-input aggregation), then we want to allow the same public key to be able to aggregate their signatures on multiple (different) messages.  This all but precludes hashing by key because each different signed message for the same public key will need a different musig coefficient. 

There needs to be an association between the unique musig coefficient and the message being signed.  One way of making this association would be to put the signed message directly into the musig coefficient.  However we would like to be able to compute the Key Aggregation before the messages are known.  Therefore the association between the message and the musig coefficient needs be indirect.  Using indexes is one natural way of building this indirection.  Another way would be to create a guid and add guid × message pairs to the aggregate message, but that appears to have no advantages over using indexes.

While hashing by key essentially forces unordered keys, hashing by index allows for the possibility of using an ordered set of pubkeys.  Protocols wanting an unordered set of pubkeys, such as Bitcoin's multisig descriptor, can easily build upon an ordered api by sorting their keys. Indeed this is what the descriptor currently does.

For the specific application of cross-input aggregation, if the ordered api is available, what we'd likely do is order the pubkey × message  pairs by input and then by execution order within the input.  If only an unordered api is available, then we would collect those pairs and sort them by pubkey, and then within each repeated set of pubkeys, sort the messages when created the aggregate message.  It strikes me that being forced to use the unordered api is slightly more complex than going with the ordered api.   OTOH, if the unsorted API handles all the work, maybe it isn't so bad.  (Sorry, I've failed to come to a conclusion on the sorting question.) (self-aside: Because Simplicity doesn't have a consensus critical execution order, so I would probably end up sorting the pubkey × message pairs emitted by a Simplicity execution.)

I haven't yet given any consideration to nested multi-message signing.

I understand the jonas has an MBDL security proof sketch for the 1 optimization and tim has a security proof sketch for the OMDL + ROM.

P.S. I hate little endian formats.  Also isn't 32-bytes too much?

P.P.S. indexes should be 0 based per Dijkstra.

jonasnick:
> P.S. I hate little endian formats. Also isn't 32-bytes too much?

My mistake, should have been 4 bytes. Changed spec and C code to encode index with MSB first.

roconnor-blockstream:
@jonasnick has explained to me in private that signature aggregation does not work the way I had imagined it to work.  signature aggregation requires a (modified) Bellare-Neven signature, which is not the same thing as a Schnorr signature.  (In particular jonas pointed out that the scheme I had in mind is broken.)

Thus we can disregard my comments about signature aggregation.

This puts hash-by-key back on the table as an option.  IMHO it is an option we should seriously consider.

jonasnick:
What would be the advantage of hash-by-key? My understanding is that we need to serialize the whole multiset of public keys before hashing them to a MuSig coefficient and therefore have to assign an order anyway.

jonasnick:
As @roconnor-blockstream mentioned to me in a separate channel, there isn't really an advantage of hash-by-key besides perhaps being a little bit simpler.

OTOH there isn't a disadvantage either. Since the input public keys are 32 byte arrays ("x-only"), hash-by-key wouldn't require additional SHA rounds.

real-or-random:
Here's my take:

### Data structure that contains the public keys: *ordered list*

This could be a set, a multiset (as in the paper), or an ordered list. I think it should be an ordered list.

#### Duplicate public keys allowed: *yes*
I don't think there's any use case that strictly needs duplicate public keys when all involved peers are honest. But making duplicate public keys illegal is difficult in higher-level protocols, where an attacker simply copies the key: Say you run MuSig with random strangers of which you don't know their public keys in a higher-level protocol (e.g., in a CoinJoin-like protocol). Moreover say there's a honest peer with public key A, and there's a malicious peer who also claims to have public key A. If key aggregation with [A, A] is illegal, then the higher-level protocol must fail. In this case, you'd at least like to know who is to blame for the failure but this at least requires additional complexity to make it possible to tell apart which of the peers is malicious. You could also canonicalize the key aggregation to [A] then, effectively removing duplicates. But this again adds complexity. 

So we should allow for duplicate public keys. However, we don't really need to optimize for this case, I think.

#### Canonical ordering/sorted: *no*
Users of the spec know best in what representation they want to deal with the public keys, so I believe we should not put any further restrictions. In particular, we shouldn't require that public keys are in some particular canonical ordering. I don't see how anyone would strictly *need* order-independence. If users really need this, they can always enforce this by sorting. 
But we should not force the users to implement sorting, nor should we make libraries forcing sorting upon their users.  

I believe there are many cases where higher-level protocol already have some indexed list of peers, so it should possible to simply take that list and use it.  

### Optimize first coefficient to be 1: *yes* 
We should do this, simply because it's more efficient. Removing a scalar multiplication is just not a micro-optimization, it can be a real difference for embedded devices, in particular when the number n of peers is small.

Of course, we need to double-check that the security proof still works out. 

### Hashing the index instead of the public key: *probably yes*
I don't think this makes much of a difference in terms of complexity. It makes a difference for duplicate public keys but I don't think we should optimize for this case (see above). So if a saves a compression, we should probably hash the index it. But this depends on how the hash is constructed exactly

### Additional argument for the hash *yes?*
Should we hash an additional byte string that can be chosen by the user? This would make it possible for instance to have a series of aggregate keys ~X that look unrelated to each other if you don't know the underlying list of public keys (by using a counter as the additional argument). Is this useful to avoid key reuse? I believe yes but I'm not entirely sure. 

Here we should also take into account possible extensions to threshold signing, where key setup is more difficult and there may be good reasons to avoid doing setup over and user.

jonasnick:
> Users of the spec know best in what representation they want to deal with the public keys,

My thinking is that ideally, they wouldn't have to care about the representation at all. @apoelstra mentioned that wallet devs are in favor of sorting the public keys in a MuSig descriptor. Then why not sort in MuSig key aggregation directly? The argument would be that that's not flexible enough, has additional computational cost for callers who really don't need them signed or has a high penalty in terms of usability.

In the meantime I have  been working on an [implementation](https://github.com/jonasnick/secp256k1-zkp/commits/musig-spec-sorted) of sorting inside key aggregation to explore these questions. It required a few refactorings but the meat of it happens at the beginning of `musig_combine_pubkeys`:

```
    /* Fast sorting algorithms have non-linear best case complexity. Therefore,
     * we only sort if the pubkeys are not already sorted */
    if (!is_sorted(pubkeys, n_pubkeys)) {
        if (signers != NULL) {
            secp256k1_musig_signers_init(signers, n_pubkeys, 1);
            sort_pubkeys(pubkeys, n_pubkeys, signers);
        } else {
            qsort(pubkeys, n_pubkeys, sizeof(*pubkeys), cmppubkey);
        }
    } else if (signers != NULL) {
        secp256k1_musig_signers_init(signers, n_pubkeys, 0);
    }
```

If the `musig_combine_pubkeys` call isn't the start of a signing session, we just qsort the pubkeys. Otherwise, we call `sort_pubkeys` which 1) sorts the signers based on their corresponding pubkey 2) assigns indices to the signers referring to the position of the pubkey in the sorted list 3) puts the pubkeys in the right position 4) signers are put in the same position (not really) they were before. As a result, the pubkeys list is sorted but `signer[i]` still refers to the `i`-th pubkey in the original list (Just noticing the nested loop now, so the algorithm is actually in O(n^2) :[).

I haven't fully repaired the tests yet, because the main problem is that the initial pubkey array can not be reused for other signers, because its order changed after the `musig_pubkey_combine` call. Another issue is that instead of calling `partial_sig_verify(pubkeys[i], ...)`, one has to call `partial_sig_verify(pubkeys[signers[i].index], ...)` but that may be fixable.

real-or-random:
> > Users of the spec know best in what representation they want to deal with the public keys,
> 
> My thinking is that ideally, they wouldn't have to care about the representation at all. @apoelstra mentioned that wallet devs are in favor of sorting the public keys in a MuSig descriptor. Then why not sort in MuSig key aggregation directly? The argument would be that that's not flexible enough, has additional computational cost for callers who really don't need them signed or has a high penalty in terms of usability.

Indeed, that's my argument. I don't want to stop anyone from sorting the keys but as I as a spec writer I wouldn't want to force the decision upon users of the spec who may not need it.


> ```
>     /* Fast sorting algorithms have non-linear best case complexity. Therefore,
>      * we only sort if the pubkeys are not already sorted */

Are you aware of https://en.wikipedia.org/wiki/Adaptive_sort ?

src/modules/musig/main_impl.h
```diff
@@ -65,10 +65,10 @@ static void secp256k1_musig_coefficient(secp256k1_scalar *r, const unsigned char
      * equivalent to hashing the public key. Because the public key can be
      * identified by the index given the ordered list of public keys (included in
      * ell), the index is just a different encoding of the public key.*/
```
real-or-random, line 67: This seems to be an argument in favor of hashing indices. What are the subtle issues with protocols building on threshold sigs?

real-or-random:
I think that the fact that we anyway have indexed arrays in our API makes sorting somewhat pointless: The user sometimes needs to pass an index to identify a specific public key. In that sense, the user anyway needs to look at this as an indexed data structure. Since the indices are there anyway, sorting the keys internally won't make anything significantly simpler from the perspective of the user. (Even if we'd require the user to pass public keys instead of indices, the user probably needs to look these public keys up in an array, so that doesn't really help.)

src/modules/musig/main_impl.h
jonasnick: Hm, I don't know and I don't remember writing that paragraph. Perhaps @apoelstra knows.

jonasnick:
> Additional argument for the hash: yes?

Alternatively, users can p2c-tweak their keys. The current API only supports doing that once which prevents adding an additional taproot commitment.

apoelstra:
After talking with @jonasnick I agree with @real-or-random that we should not try to sort the keys internally. It seemed like the simpler option but actually it makes the API a fair bit more complex, as well as (potentially) imposing a performance hit on users who don't want it.

src/modules/musig/main_impl.h
apoelstra: I believe I had some hypothetical variant of Russell's attack in mind here -- but since all participants are signing the same message, it doesn't apply. I was just confused.

jonasnick:
> I haven't fully repaired the tests yet, because the main problem is [...] Another issue is that instead of calling [...] but that may be fixable.

One way to fix this would be to sort the pubkey array back into the original order, but that would only work by requiring additional space even if no signin session is initiated. Instead, I pushed the following changes

1. The MuSig coefficient is a hash of the public key, not the index. This also allows to completely remove the `signers` array in MuSig2.
2. A new `sort_pubkeys` function is added that can be used before `combine_pubkey`. Since this is now separate, in-place sorting of pubkeys doesn't create a mess in the tests.

jonasnick:
I pushed a commit to add an optimization where the second unique key in the pubkey list gets MuSig coefficient 1 as per the latest (unpublished) revision of the MuSig2 paper.
The actual spec is still a very rough draft, but it can be improved over time. I think this PR is ready for review and I reordered the commits to make that easier.

To quickly summarize the changes:
- A function for sorting public keys allows to aggregate keys independent of the (initial) order.
- The MuSig coefficient is computed by hashing the key instead of key index because. Otherwise, if the pubkey list gets sorted, the signer needs to translate between key indices pre- and post-sorting. It also means that with MuSig2 we can completely remove the `secp256k1_musig_session_signer_data` struct.
- The second unique key in the pubkey list gets the constant MuSig coefficient 1 which saves an exponentiation.

It doesn't yet accept an additional argument like a counter to allow generating distinct keys from the same list of pubkeys as @real-or-random suggested.
I think this is an interesting idea, but maybe we can discuss this on a separate PR.

real-or-random:
This sounds good. What caught my eye on a first glimpse is that there are a few places where `secp256k1_memcmp_var` should be used instead of `memcmp`. I'll have a closer look soon.

jonasnick:
I replaced the added `memcmp`s with `secp256k1_memcmp_var`.

> Additional argument for the hash: yes?

Perhaps worth noting that a similar effect can be achieved with key tweaking.

real-or-random:
Maybe add a random sorting test too. (Random array, assert it's sorted after calling sort).

ACK otherwise

jonasnick:
Done. @roconnor-blockstream @apoelstra concept ACK? See https://github.com/ElementsProject/secp256k1-zkp/pull/120#issuecomment-800677101 for a summary of the changes or the spec draft [here](https://github.com/ElementsProject/secp256k1-zkp/pull/120/files#diff-cc77294f093b6c3f65a8ec4d96a1246988a88ddf845d9f6a784fd99343945004R23).

real-or-random (dismissed):
ACK 4000700249f5a1e8b2a5c93121b5d97e35346aa3

roconnor-blockstream:
I’m a little bit nervous that `secp256k1_musig_pubkey_combine_callback` both takes an index argument yet is required to be called sequentially.  I know that currently the implementation of `secp256k1_ecmult_multi_var` does invoke the callback sequentially, but I'm unconvinced that this sequential calling pattern is mandated by of the `secp256k1_ecmult_multi_var` API and may just happen to be an detail of the current implementation.  (If sequential access is supposed to be guaranteed by the API, why does it then bother adding an index parameter to the call back which could instead be computed and stored by the callback itself?)

src/modules/musig/tests_impl.h
```diff
@@ -950,6 +961,191 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+        CHECK(secp256k1_musig_sort_pubkeys(ctx, pk, n_pks));
+    }
+    CHECK(secp256k1_musig_pubkey_combine(ctx, NULL, &combined_pk, &pre_session, pk, n_pks) == 1);
+    CHECK(secp256k1_fe_set_b32(&second_pk_x, pre_session.second_pk));
+    CHECK(secp256k1_fe_is_zero(&second_pk_x) == !has_second_pk);
+    if (!secp256k1_fe_is_zero(&second_pk_x)) {
+        CHECK(secp256k1_memcmp_var(&pk_ser[second_pk_idx], &pre_session.second_pk, sizeof(pk_ser[second_pk_idx])) == 0);
+    }
+    CHECK(secp256k1_xonly_pubkey_serialize(ctx, combined_pk_ser, &combined_pk));
+    /* TODO: remove */
```
ysangkok, line 1061: what's the point of keeping this commented code around?

src/modules/musig/main_impl.h
```diff
@@ -12,6 +12,27 @@
...
 
+int secp256k1_musig_sort_pubkeys(const secp256k1_context* ctx, secp256k1_xonly_pubkey *pubkeys, size_t n_pubkeys) {
+    size_t i, j;
+    secp256k1_xonly_pubkey key;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(pubkeys != NULL);
+
+    /* insertion sort */
+    /* TODO: replace with more efficient algo. May require scratch space. */
```
ysangkok, line 23: I am curious as to why the ANSI C qsort function can't be used?

ysangkok: is it because it could have "non-linear best case complexity"? would be nice to have that mentioned in the comment.

jonasnick:
@roconnor-blockstream Doing it outside of ecmult looks less clean, because we have an additional loop that calls pubkey_load. I convinced myself that non-sequential callback calling is not a problem because a test vector would immediately catch it. But that's of course not true if the behavior would be changed to something non-deterministic. So I pushed a fixup commit to move it outside the callback (and while doing that I noticed that we were missing test vector which I added in a another commit).

src/modules/musig/tests_impl.h
jonasnick: The point is that it's easy to print test vectors if we change them. I updated the comment to make that more clear.

src/modules/musig/main_impl.h
```diff
@@ -130,6 +127,16 @@ int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scrat
     ecmult_data.pks = pubkeys;
     /* No point on the curve has an X coordinate equal to 0 */
     secp256k1_fe_set_int(&ecmult_data.second_pk_x, 0);
+    for (i = 0; i < n_pubkeys; i++) {
```
roconnor-blockstream, line 130: Not 100% sure, but I think it is safe to start at `i = 1`.

src/modules/musig/main_impl.h
jonasnick: It's because a previous iteration of this PR (hash-by-index) required a more flexible sorting API than qsort (like qsort_r, but that is not available everywhere). I forgot that this could be removed now. Thanks! Pushed fixup commit to replace by qsort.

src/modules/musig/main_impl.h
```diff
@@ -12,24 +12,15 @@
 #include "include/secp256k1_musig.h"
 #include "hash.h"
 
-int secp256k1_musig_sort_pubkeys(const secp256k1_context* ctx, secp256k1_xonly_pubkey *pubkeys, size_t n_pubkeys) {
-    size_t i, j;
-    secp256k1_xonly_pubkey key;
+static int secp256k1_musig_sort_pubkeys_cmp (const void *p1, const void *p2) {
+  return secp256k1_memcmp_var(p1, p2, sizeof(secp256k1_xonly_pubkey));
```
roconnor-blockstream, line 16: from `secp256k1_xonly_pubkey`:

    *  The exact representation of data inside is implementation defined and not
    *  guaranteed to be portable between different platforms or versions. It is
    *  however guaranteed to be 64 bytes in size, and can be safely copied/moved.
    *  If you need to convert to a format suitable for storage, transmission, or
    *  comparison, use secp256k1_xonly_pubkey_serialize and
    *  secp256k1_xonly_pubkey_parse.

Not only is sorting this way an API violation, I believe that you will literally get different orders on architectures of different endianness.

real-or-random: Indeed, and the tests can be improved then. Maybe we should sync this with https://github.com/bitcoin-core/secp256k1/pull/850

roconnor-blockstream: Just be aware that `secp256k1_xonly_pubkey_cmp` cannot be used with `qsort` due to its (nearly useless) `ctx` argument.  I'm not sure if this comment belongs here or there.

roconnor-blockstream: I suppose here it could be wrapped up with a `secp256k1_context_no_precomp` context for the `qsort` call.

jonasnick: Ugh, that's what you get when you use memcmp for equality testing too often. CI caught it.

All functions operating on pubkeys need a context, because its required for `pubkey_load`. If I understand the `no_precomp` suggestion correctly, it has the downside that if `pubkey_load` fails, the illegal callback of that temporary context will be called instead of whatever the caller of `sort_pubkeys` may have set. So I see the two options

1. Call `pubkey_load` for all pubkeys with the correct context before qsorting to make sure that it won't fail when called in the comparator.
2. Revert back to insertion sort (temporary). My preferred solution because it doesn't make fragile assumptions about the behavior of the contexts.

roconnor-blockstream: I figure you would do 1.

real-or-random: 1b. Define a `secp256k1_pubkey_load_no_callback` that simply returns 0 but doesn't need a context to access a callback.

These dynamic callbacks are annoying in a few other places, e.g. you can't call for them for illegal arguments when *creating* a context (chicken-and-egg problem). I'll probably make a suggestion to get rid of the callbacks in the contexts and only have compile-time overrides for the default callbacks. I don't think anyone needs the ability to change the callbacks at runtime.

real-or-random: By the way, once keys are loaded, we  can just use `secp256k1_fe_cmp_var`.

jonasnick: @roconnor-blockstream why?

roconnor-blockstream: @real-or-random (a) `secp256k1_pubkey_load_no_callback` would mean not using `secp256k1_xonly_pubkey_cmp` or involve changing the definition in that PR (though these options are doable I guess).

(b) There is no place to load all the pubkeys into in order to sort by `secp256k1_fe_cmp_var`.

@jonasnick Because you get you use `qsort`[1] instead of insertion sort, and I'm dubious of your claim that it would make fragile assumptions of the behavior of contexts.  Maybe I don't understand what you mean by "fragile assumptions", but if `secp256k1_xonly_pubkey_cmp` cannot compare xonly pubkeys that can all be successfully loaded, then something is wrong with the specification of `secp256k1_xonly_pubkey_cmp`.

But hey, if you want to use insertion sort, I'm not going to get in your way.

[1] not that qsort is that great to begin with: https://travisdowns.github.io/blog/2019/05/22/sorting.html

roconnor-blockstream: Oh i see, your concern is that you want to be open to the possibility that loading a pubkey requires a precomputation table or something.

    /** A simple secp256k1 context object with no precomputed tables. These are useful for
     *  type serialization/parsing functions which require a context object to maintain
     *  API consistency, but currently do not require expensive precomputations or dynamic
     *  allocations.
     */
    SECP256K1_API extern const secp256k1_context *secp256k1_context_no_precomp; 

Loading an x-only pubkey seems like it qualifies as "type serialization/parsing functions which require a context object to maintain API consistency, but currently do not require expensive precomputations or dynamic allocations."  If you consider this use as fragile then you might as well open a PR to eliminate `secp256k1_context_no_precomp` because I cannot imagine a safer place to use it than here.

jonasnick: But as mentioned above, just using `secp256k1_context_no_precomp` isn't safe at all in this case. When encountering an illegal pubkey, the default callback aborts the program instead of calling the user's callback. The assumption inside `pubkey_load` is that the context is the user's not some temporary context. We can fix this for now by trying to load all pubkeys with the user's context object before calling qsort just to ensure that there's no illegal pubkey and the default illegal callback isn't called on the temporary context.

The main issue is that qsort doesn't have a powerful enough interface, so I don't see why we'd want to hack around this instead of just using a function with the right interface.

roconnor-blockstream: @jonas as you yourself mention above you can "Call pubkey_load for all pubkeys with the correct context before qsorting to make sure that it won't fail when called in the comparator." Subsequent calls to pubkey_load can then use `secp256k1_context_no_precomp` safely.

Anyhow if you think "[deliberately quadratic](https://accidentallyquadratic.tumblr.com/)] is okay for this code, then I won't stop you.

real-or-random: > (b) There is no place to load all the pubkeys into in order to sort by `secp256k1_fe_cmp_var`.

Indeed but I didn't want to suggest this. My suggestion was to always perform two loads inside `secp256k1_xonly_pubkey_cmp` and then to call `secp256k1_fe_cmp_var`. That's not great for performance but at least it's the same complexity as `qsort`.

jonasnick: Yes, this hack works in this specific case, but it still breaks the assumption that pubkey_load is called on the users context object. The quadratic sort function was certainly only meant to be temporary for the purpose of implementing the spec. I can see the point that until it is replaced this can be viewed as a DoS vector. Do you know of a drop-in sorting implementation that we could use?

roconnor-blockstream: @real-or-random How do you "perform a load" inside `secp256k1_xonly_pubkey_cmp` without a context object?

roconnor-blockstream: I don't see what assumption break.  the preprocessing loop *does* call pubkey_load on the user's context object.  The only "assumption" I'm making is that `secp256k1_xonly_pubkey_cmp` doesn't call anything in the context object when the two keys being compared are loadable.

roconnor-blockstream: Finding an off the shelf quasi-linear, in-place, no-allocation, "reentrant" sorting function with a suitable license might be a challenge.  glibc's quicksort has a GPL license.  FreeBSD's `qsort_r` [might not be quasi-linear](http://calmerthanyouare.org/2014/06/11/algorithmic-complexity-attacks-and-libc-qsort.html).  Though presumably it is fixed by now.

jonasnick: My specific worry was that the context could be used to store additional things, like the number of `pubkey_load`s which would be wrong after `pubkey_sort`. Seeing that the context is const almost everywhere this is likely a non-issue.

> Do you know of a drop-in sorting implementation that we could use?

If the answer is no, I can implement the qsort hack for now with a TODO comment to fix this at some point so we have at least that quadratic DoS issue fixed.

real-or-random: > @real-or-random How do you "perform a load" inside `secp256k1_xonly_pubkey_cmp` without a context object?

Using `secp256k1_pubkey_load_no_callback`. Sorry I should be verbose.

roconnor-blockstream: For the record, I floated an idea on IRC that `secp256k1_xonly_pubkey_cmp` and friends be amended to explicitly allow a `NULL` context to be passed, and won't invoke the callback upon an error.

jonasnick: That's not unreasonable, but it's not clear to me how it helps here. If we'd call `secp256k1_xonly_pubkey_cmp` with ctx=NULL, we'd still want to separately try loading with the user's context. Otherwise, there are API inconsistencies where an illegal pubkey provided to `pubkey_serialize` invokes the callback, but when provided to `pubkey_sort` the function just returns 0.

roconnor-blockstream: I suppose that is a question for @real-or-random, but what is the API inconsistency exactly?

Assuming we went with this modified real-or-random-1b plan, we simply observe that  pubkey_serialize zeros the output buffer when it returns 0.  Thus all invalid keys would compare as less than every valid pubkey, and all invalid keys would compare as equal, and get sorted in an implementation defined order.  This behavour can be clearly specified in `secp256k1_xonly_pubkey_cmp` and `pubkey_sort`.  Indeed this is already what happens if/when the `illegal_callback` returns.  After sorting with erroneous keys, if you try to musig assemble the pubkeys, you will fail then anyways.

But if you have erroneous blobs of memory in your xonly-pubkey array, then every proposal so far dictates garbage-in garbage-out.  As it stands already nothing in `secp256k1_xonly_pubkey_load` checks that the memory blob is on curve or has an even y-coordinate.  The `secp256k1_xonly_pubkey_load` function only fails when the blob of memory has a 0 x-coordinate, which is a scenario that should only happen if someone has failed to check the return code of some operation.  In which case I don't see what's wrong with just specifying that all these invalid keys get sorted to the beginning of the array.  They will still be just as invalid when they are used later.  There seems to be no harm in sorting them.

roconnor-blockstream: If we do choose to go wtihout the preprocessing step of trying to load every key before sorting, we would simply want to remove the ctx argument to make it clear to everyone that no callback is going to be invoked.

jonasnick: The inconsistency would be as follows
```
secp256k1_xonly_pubkey pk;
memset(pk, 0, sizeof(pk));
secp256k1_xonly_serialize(ctx, buf, pk); /* operates on illegal pubkey, calls illegal callback, returns 0 */
secp256k1_pubkey_sort(ctx, buf, pk); /* operates on illegal pubkey, does not call illegal callback, doesn't return 0 */
```

It seems that you're saying that this wouldn't be a problem because the callback would be invoked after pubkey_sort, in key aggregation or serialization. 

> remove the ctx argument 

Remove it where?

roconnor-blockstream: Yes, I'm saying that sorting illegal pubkeys could be considered okay because the only thing you can do with a sorted array with illegal pubkeys is to do something else with them and whatever that is will trigger the callback.

The modified-1b proposal would end up looking something like

    static int secp256k1_musig_sort_pubkeys_cmp (const void *p1, const void *p2) {
        return secp256k1_xonly_pubkey_cmp(NULL, p1, p2);
    }
    
    void secp256k1_musig_sort_pubkeys(secp256k1_xonly_pubkey *pubkeys, size_t n_pubkeys) {
        VERIFY_CHECK(pubkeys != NULL);

        qsort(pubkeys, n_pubkeys, sizeof(*pubkeys), secp256k1_musig_sort_pubkeys_cmp);
    }

roconnor-blockstream: Possible problem:  C doesn't make any complexity guarantees about `qsort`'s complexity.  In particular [glibc's `_quicksort`](https://elixir.bootlin.com/glibc/latest/source/stdlib/qsort.c#L74), which `qsort` falls back to when malloc fails, doesn't implement the median of medians pivot which is usually required to bring down the worst case complexity to quasi-linear.

Would you like me to write you an in-place heap sort?

jonasnick: @roconnor-blockstream good catch. Yes that would be great. Standard in-place and iterative heap sort should be easy to review.

jonasnick:
Rebased, cherry-picked latest version of https://github.com/bitcoin-core/secp256k1/pull/850 and squashed commits.

src/modules/musig/main_impl.h
jonasnick: Replaced this again with a heapsort implementation by roconnor implementation, because qsort can be slow on malicious input (as you also mentioned). Also added comment.

jonasnick:
Added and squashed a few minor tweaks.

src/modules/extrakeys/tests_impl.h
```diff
@@ -571,6 +571,43 @@ void test_keypair_add(void) {
...
+}
+
+static int test_hsort_cmp(const void *i1, const void *i2, void *counter) {
+  *(size_t*)counter += 1;
+  return *(int*)i1 - *(int*)i2;
+}
+
+#define NUM 64
+void test_hsort(void) {
+    int ints[NUM] = { 0 };
```
real-or-random, line 588: These integer tests never hit the `stride > 64` path in the swaps. It may be a good idea to add a test case for wider types (e.g., scalars because we have a cmp function in the code already). 

edit: ah ok, we're doing this later anyway with the pubkeys... So feel free to ignore.

include/secp256k1_musig.h
```diff
@@ -137,7 +137,13 @@ typedef struct {
...
 
 /** Computes a combined public key and the hash of the given public keys.
- *  Different orders of `pubkeys` result in different `combined_pk`s.
+ *
+ *  Different orders of `pubkeys` result in different `combined_pk`s. The
+ *  pubkeys can be sorted before combining with `secp256k1_xonly_sort` which
+ *  ensures the same resulting `combined_pk` for the same multiset of pubkeys.
+ *
+ *  This is useful to do before pubkey_combine, such that the order of pubkeys
+ *  does not affect the combined public key.
```
real-or-random, line 147: Nit: It would be more natural to break the paragraph after "different `combined_pk`s" instead of where it is broken now.

src/modules/extrakeys/main_impl.h
```diff
@@ -155,6 +155,25 @@ int secp256k1_xonly_pubkey_tweak_add_check(const secp256k1_context* ctx, const u
             && secp256k1_fe_is_odd(&pk.y) == tweaked_pk_parity;
 }
 
+typedef struct {
+    const secp256k1_context *ctx;
+} secp256k1_xonly_sort_cmp_data;
+
+static int secp256k1_xonly_sort_cmp(const void* pk1, const void* pk2, void *cmp_data_tmp) {
+    secp256k1_xonly_sort_cmp_data* cmp_data = cmp_data_tmp;
+    return secp256k1_xonly_pubkey_cmp(cmp_data->ctx, *(secp256k1_xonly_pubkey **)pk1, *(secp256k1_xonly_pubkey **)pk2);
```
real-or-random, line 164: nit: I think it would be more readable to simply cast the pointer (but I guess the compiler does not really do an assignment anyway).

Since the context pointer is a pointer, we could also save the `secp256k1_xonly_sort_cmp_data` type entirely and simply pass the context pointer a third argument to `secp256k1_xonly_sort_cmp`. Not sure if this is better, but it's less boilerplate at least.

real-or-random (commented):
I didn't go through the entire PR again but I looked at the heapsort stuff, which is new.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,100 @@
...
+ */
+
+/* Precondition: 2*i <= SIZE_MAX - 1 */
+static SECP256K1_INLINE size_t child1(size_t i) { return 2*i + 1; }
+
+/* Precondition: 2*i <= SIZE_MAX - 2 */
+static SECP256K1_INLINE size_t child2(size_t i) { return child1(i)+1; }
+
+/* Precondition: 0 < i */
+static SECP256K1_INLINE size_t parent(size_t i) { return (i-1)/2; }
```
roconnor-blockstream, line 23: I don't know if this is a problem or not, but I did notice that `parent` seems to be only used in comments.

src/modules/extrakeys/hsort.h
```diff
@@ -0,0 +1,19 @@
...
+ * file COPYING or https://www.opensource.org/licenses/mit-license.php.*
+ ***********************************************************************/
+
+#ifndef SECP256K1_HSORT_H_
+#define SECP256K1_HSORT_H_
+
+#include <stddef.h>
+#include <string.h>
+
+/* In-place, iterative heapsort with an interface matching glibc's qsort_r. This
```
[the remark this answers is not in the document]
roconnor-blockstream: Isn't that interface spelled out below in hsort's signature?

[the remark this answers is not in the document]
real-or-random: Ok yes, I meant the semantics. You can make an educated guess but it's not entirely obvious what `cmp_data` is, what the difference between `count` and `size` is, `ptr` doesn't make entirely clear that we're sorting an array, etc.

src/modules/extrakeys/hsort_impl.h
real-or-random: Indeed, I think then we should remove it or better replace it by a comment.

src/modules/extrakeys/main_impl.h
jonasnick: > I think it would be more readable

Hm, I found this slightly less readable, but either way is fine for me.

> we could also save the secp256k1_xonly_sort_cmp_data type

The reason the context is wrapped is that the `cmp_data` argument expects a non-const pointer and this seems slightly cleaner despite the boilerplate than having to cast the context's constness away.

src/modules/extrakeys/hsort_impl.h
jonasnick: done

src/modules/extrakeys/main_impl.h
jonasnick: Added a comment why the struct exists.

include/secp256k1_musig.h
jonasnick: done

src/modules/extrakeys/tests_impl.h
jonasnick: ignored :)

src/modules/extrakeys/tests_impl.h
```diff
@@ -571,6 +571,43 @@ void test_keypair_add(void) {
...
+    size_t counter;
```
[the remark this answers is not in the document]
jonasnick: good idea

src/modules/extrakeys/hsort.h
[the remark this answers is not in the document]
jonasnick: No need to make an educated guess, just run `man qsort`. But anyway, I copied and adjusted the manpage text.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,100 @@
...
+    swap64(a + (stride - 64), i, j, 64);
+    stride -= 64;
+  }
+  swap64(a, i, j, stride);
+}
+
+static SECP256K1_INLINE void heap_down(char *a, size_t i, size_t heap_size, size_t stride,
+                            int (*cmp)(const void *, const void *, void *), void *cmp_data) {
+  while (i < heap_size/2) {
+    /* (a) child(1) < heap_size:
```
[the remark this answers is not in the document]
jonasnick: I think so too. Fixed and added VERIFY_CHECKs.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,100 @@
...
+
+    /* (b) 2*i <= SIZE_MAX - 2 :
+     * i < heap_size/2
+     * i + 1 <= heap_size/2
+     * 2*i + 2 <= heap_size <= SIZE_MAX
+     * 2*i <= SIZE_MAX - 2
+     */
+
+    /* If any child of i is larger than i, swap the i with that child.
+     * If both children of i are larger than i, swap i with the larger child to ensure the new parent is larger than both childeren.
```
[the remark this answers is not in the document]
jonasnick: fixed

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,100 @@
...
+
+/* Precondition: 2*i <= SIZE_MAX - 2 */
+static SECP256K1_INLINE size_t child2(size_t i) { return child1(i)+1; }
+
+/* Precondition: 0 < i */
+static SECP256K1_INLINE size_t parent(size_t i) { return (i-1)/2; }
+
+/* Precondition: stride <= 64 */
+static SECP256K1_INLINE void swap64(char *a, size_t i, size_t j, size_t stride) {
+  char tmp[64];
```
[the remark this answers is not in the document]
jonasnick: good idea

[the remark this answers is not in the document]
jonasnick: good catch

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,100 @@
...
+#define SECP256K1_HSORT_IMPL_H_
+
+#include "hsort.h"
+
+/* An array is a heap when, for all non-zero indexes i,
+ * the element at index i compares as less than or equal to the element at index parent(i).
+ */
+
+/* Precondition: 2*i <= SIZE_MAX - 1 */
+static SECP256K1_INLINE size_t child1(size_t i) { return 2*i + 1; }
```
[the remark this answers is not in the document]
jonasnick: done

configure.ac
```diff
@@ -561,7 +562,8 @@ fi
 if test x"$enable_module_surjectionproof" = x"yes"; then
   AC_DEFINE(ENABLE_MODULE_SURJECTIONPROOF, 1, [Define this symbol to enable the surjection proof module])
 fi
-
+# Test if extrakeys is set _afte_r the MuSig module to allow the MuSig
```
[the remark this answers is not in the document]
jonasnick: fixed

src/modules/extrakeys/hsort_impl.h
```diff
@@ -9,92 +9,96 @@
 
 #include "hsort.h"
 
-/* An array is a heap when, for all non-zero indexes i,
- * the element at index i compares as less than or equal to the element at index parent(i).
+/* An array is a heap when, for all non-zero indexes i, the element at index
```
real-or-random, line 12: s/index/index i/

src/modules/extrakeys/main_impl.h
real-or-random: Oh sure, I missed the constness. In that case, the current version is cleaner for sure.

src/modules/extrakeys/hsort_impl.h
jonasnick: fixed

src/modules/extrakeys/hsort.h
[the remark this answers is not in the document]
jonasnick: :(

I removed the description and instead refer to the manpage.

src/modules/extrakeys/hsort_impl.h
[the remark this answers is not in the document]
roconnor-blockstream: Maybe it makes sense to reverse the order of these two VERIFY_CHECKs, as the second check is the precondition of the `child1(i)` which is called by the first VERIFY_CHECK.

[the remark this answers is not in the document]
roconnor-blockstream: My claim above is slightly incorrect.  The second check is the precondition for `child2(i)`, which in turn implies the precondition for `child1(i)`.  I still think the order should be reversed though.  Of course, the `child1` function does have its own VERIFY_CHECK statement, so if you do want to let this go, I'm okay with that.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,104 @@
...
+         * i + 1 <= heap_size/2
+         * 2*i + 2 <= heap_size <= SIZE_MAX
+         * 2*i <= SIZE_MAX - 2
+         */
+
+        /* If any child of i is larger than i, swap the i with that child. If
+         * both children of i are larger than i, swap i with the larger child to
+         * ensure the new parent is larger than both children.
+         */
+        if (child2(i) < heap_size && 0 < cmp(a + child2(i)*stride, a + child1(i)*stride, cmp_data)
```
roconnor-blockstream, line 66: I don't know if this comment will be is going too far, but maybe this comparison between child2 and child1 should be non-strict.  Why? Because if child1 and child2 are equal (or equivalent) then we should prefer to swap the parent with child2 over child1 because the heap on the child2 side is smaller or equal in size compared to the size of the heap on the child1 side.  And thus if we recurse into the child2 heap, we might terminate one iteration faster.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,104 @@
...
+                              && 0 < cmp(a + child2(i)*stride, a +         i*stride, cmp_data)) {
```
roconnor-blockstream, line 67: It might be slightly better to break the comparison of child2 with i into its own if statement, because if this comparison fails (after having checked that child2 is greater than (or equal to) child1) then it must be the case that, by transitivity, that i is also greater than or equal to child1, and thus the next comparison on the line below is going to fail, and we can skip it.

Thus my suggestion is to replace this branch with:

    if (0 < cmp(a + child2(i)*stride, a +         i*stride, cmp_data)) {
        swap(a, i, child2(i), stride);
        i = child2(i);
    } else {
        return;
    }

src/modules/extrakeys/hsort_impl.h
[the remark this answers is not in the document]
jonasnick: Ok, I reversed the VERIFY_CHECKs

src/modules/extrakeys/hsort_impl.h
jonasnick: Done

src/modules/extrakeys/hsort_impl.h
jonasnick: Done, also added comment explaining this.

src/modules/extrakeys/hsort_impl.h
```diff
@@ -0,0 +1,110 @@
...
+         * ensure the new parent is larger than both children.
+         */
+        if (child2(i) < heap_size
+                && 0 <= cmp(a + child2(i)*stride, a + child1(i)*stride, cmp_data)) {
+            if (0 < cmp(a + child2(i)*stride, a +         i*stride, cmp_data)) {
+                swap(a, i, child2(i), stride);
+                i = child2(i);
+            } else {
+               /* Since child2(i) <= child1(i) and child2(i) > i, we have
+                * child1(i) > i. Thus, the next comparison can be skipped. */
```
roconnor-blockstream, line 73: The comparison are all backwards/in error/misleadingly looking like the indexes are compared rather than the contents at the index in this comment.

At this point we have `[child2(i)] >= [child1(i)]` and we have `[child2(i)] <= [i]`, and thus `[child1(i)] <= [i]`, where `[x]` indicating the contents at `a[x*stride]`.

jonasnick: > backwards

Ugh. Interesting how my brain put the comparison in the correct order (non-constant first) but wrongly...

This notation should already be used then in the comment above ("If any child of i is larger than i").

src/modules/extrakeys/hsort_impl.h
```diff
@@ -59,18 +59,22 @@ static SECP256K1_INLINE void heap_down(unsigned char *a, size_t i, size_t heap_s
          * child1(i) < heap_size
          */
 
-        /* If any child of i is larger than i, swap the i with that child. If
-         * both children of i are larger than i, swap i with the larger child to
+        /* Let [x] be notation for the contents at a[x*stride].
+         *
+         * If [child1(i)] > [i] and [child2(i)] > [i], swap [i] with the larger child to
          * ensure the new parent is larger than both children.
```
roconnor-blockstream, line 65: ... preferring to swap with child2 in the case when [child1(i)] == [child2(i)].

src/modules/extrakeys/hsort_impl.h
```diff
@@ -59,18 +59,22 @@ static SECP256K1_INLINE void heap_down(unsigned char *a, size_t i, size_t heap_s
...
+         * Else if [child1(i)] > [i], swap [i] with [child1(i)].
+         * Else if [child2(i)] >= [i], swap [i] with [child2(i)].
```
roconnor-blockstream, line 67: `[child2(i)] > [i]`

Swaps only happen when the child is strictly greater than the parent.

src/modules/extrakeys/hsort_impl.h
jonasnick: fixed

src/modules/extrakeys/hsort_impl.h
jonasnick: fixed

jonasnick:
Squashed fixups and added commit that renames the MuSig coefficient to KeyAgg coefficient (which changes the test vectors due to tagged hashing).

jonasnick:
rebased

real-or-random:
Asan is complaining here

src/modules/musig/main_impl.h
```diff
@@ -45,23 +45,36 @@ static void secp256k1_musig_sha256_init_tagged(secp256k1_sha256 *sha) {
...
+        secp256k1_musig_sha256_init_tagged(&sha);
+        secp256k1_sha256_write(&sha, ell, 32);
+        secp256k1_fe_get_b32(buf, x);
+        secp256k1_sha256_write(&sha, buf, 32);
+        secp256k1_sha256_finalize(&sha, buf);
+        secp256k1_scalar_set_b32(r, buf, NULL);
+    }
+}
+
+static void secp256k1_musig_coefficient(secp256k1_scalar *r, const secp256k1_musig_pre_session *pre_session, secp256k1_fe *x) {
```
real-or-random: Also add a comment here like "Assumes both field elements x and second_pk_x are normalized."? Or move the existing one here.

include/secp256k1_musig.h
```diff
@@ -25,6 +25,8 @@ extern "C" {
  *        magic: Set during initialization in `pubkey_combine` to allow
  *               detecting an uninitialized object.
  *      pk_hash: The 32-byte hash of the original public keys
+ *    second_pk: Serialized x-coordinate of the second public key in the list. Is 0
```
real-or-random, line 28: nit: "0" is not 100% clear for this type

src/modules/musig/musig-spec.mediawiki
```diff
@@ -0,0 +1,102 @@
...
+
+=== Motivation ===
+
+== Description ==
+
+=== Design ===
+
+* A function for sorting public keys allows to aggregate keys independent of the (initial) order.
+* The KeyAgg coefficient is computed by hashing the key instead of key index. Otherwise, if the pubkey list gets sorted, the signer needs to translate between key indices pre- and post-sorting.
+* The second unique key in the pubkey list gets the constant KeyAgg coefficient 1 which saves an exponentiation.
```
real-or-random, line 27: nit: maybe refer to the Appendix in the paper

(ok, yes, we need to expand this anyway...)

src/modules/musig/main_impl.h
```diff
@@ -129,6 +131,7 @@ int secp256k1_musig_pubkey_combine(const secp256k1_context* ctx, secp256k1_scrat
         return 0;
     }
     if (!secp256k1_ecmult_multi_var(&ctx->error_callback, &ctx->ecmult_ctx, scratch, &pkj, NULL, secp256k1_musig_pubkey_combine_callback, (void *) &ecmult_data, n_pubkeys)) {
+        /* The current implementation of ecmult_multi_var makes this code unreachable with tests. */
```
real-or-random, line 134: If you want, you could try playing around with https://gcovr.com/en/stable/guide.html#exclusion-markers . (But this can be done later)

real-or-random (approved):
ACK mod nits

real-or-random:
Wait, the sorting function has disappeared. Is this intentional?

jonasnick:
@real-or-random oops...messed up the rebase

src/modules/musig/main_impl.h
jonasnick: done

include/secp256k1_musig.h
jonasnick: agree, fixed

src/modules/musig/musig-spec.mediawiki
jonasnick: done

src/modules/musig/main_impl.h
jonasnick: let's do this later

real-or-random:
ACK fc26ca8ddef0629c7df190f1cc92157fce64e370

src/modules/musig/tests_impl.h
```diff
@@ -950,6 +987,145 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+                break;
+            case 3:
+                /* [X1, X1, X2, X2] */
+                n_pks = 4;
+                memcpy(pk_ser_tmp[0], pk_ser[0], sizeof(pk_ser_tmp[0]));
+                memcpy(pk_ser_tmp[1], pk_ser[0], sizeof(pk_ser_tmp[1]));
+                memcpy(pk_ser_tmp[2], pk_ser[1], sizeof(pk_ser_tmp[2]));
+                memcpy(pk_ser_tmp[3], pk_ser[1], sizeof(pk_ser_tmp[3]));
+                has_second_pk = 1;
+                second_pk_idx = 3;
```
RCasatta, line 1120: this should be 2?

I don't think this is the issue but my (toy code)[https://github.com/LLFourn/secp256kfun/pull/67] is failing on this case (while the other 3 are ok)

src/modules/musig/musig-spec.mediawiki
```diff
@@ -0,0 +1,102 @@
...
+The algorithm ''KeyAgg(pk<sub>1..u</sub>)'' is defined as:
+* For ''i = 1 .. u'':
+** Let ''a<sub>i</sub> = KeyAggCoeff(pk<sub>1..u</sub>, i)''.
+** Let ''P<sub>i</sub> = lift_x(int(pk<sub>i</sub>))''; fail if it fails.
+* Let ''S = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(S)''.
+* Return ''bytes(S)''.
+
+The algorithm ''HashKeys(pk<sub>1..u</sub>)'' is defined as:
+* Return ''hash(pk<sub>1</sub> || pk<sub>2</sub> || ... || pk<sub>u</sub>)''
```
RCasatta, line 82: May I know why this is not a tagged hash?

jonasnick: The key agg coefficient should be domain separated since the outer hash is tagged. But since to the best of my knowledge domain separation isn't well defined, I'm leaning towards making this tagged. This would also prevent discussions like this in the future (and we're obviously already using a lot of tagged hashes). How about tag "KeyAgg list"?

src/modules/musig/tests_impl.h
jonasnick: `second_pk_idx = 2` and `second_pk_idx = 3` are both correct since they both get keyagg coefficient = 1. But agree that it's better to use the first one encountered here (and perhaps add comment that second_pk_idx = 3 is equally valid).

src/modules/musig/musig-spec.mediawiki
RCasatta: > How about tag "KeyAgg list"?

Sounds good

PR #124, opened 2021-02-09
ecdsa_s2c: rename anti-klepto to anti-exfil

apoelstra:


real-or-random (approved):
ACK e354c5751d670716791778d86c1dcd599f2e1a9c diff looks good, I haven't checked that no occurrence is missed

apoelstra:
you can just `git grep -i klepto` and see it doesn't appear

benma:
The commit message is unfortunately empty. Just curious, what was the reason for the change? Noticing it now since I am updating to a newer version of this lib.

apoelstra:
Blockstream marketirng saying that "kleptography" is not a widely understood term.

PR #128, opened 2021-03-24
Make function argument name consistent with doc

sanket1729:
I don't know what users usually prefer, `key32` or `seed32`. But, at least the naming should be consistent.


real-or-random (approved):
ACK cc82ad5ab743c6c74793d1e5cd5cee6f60175a53

apoelstra:
utack

PR #132, opened 2021-06-14
Upstream PRs 831, 907, 903, 889, 918, 906, 928, 922, 933, Merge bitcoin-core/secp256k1#936: Fix gen_context/ASM build on ARM, 925, 937, 926, Merge bitcoin-core/secp256k1#940: contrib: Explain explicit header guards, 850, 930, 941, 846, 947, 662, 950

jonasnick:
[bitcoin-core/secp256k1#831]: Safegcd inverses, drop Jacobi symbols, remove libgmp
[bitcoin-core/secp256k1#907]: changed import to use brackets <> for openssl
[bitcoin-core/secp256k1#903]: Make argument of fe_normalizes_to_zero{_var} const
[bitcoin-core/secp256k1#889]: fix uninitialized read in tests
[bitcoin-core/secp256k1#918]: Clean up configuration in gen_context
[bitcoin-core/secp256k1#906]: Use modified divsteps with initial delta=1/2 for constant-time
[bitcoin-core/secp256k1#928]: Define SECP256K1_BUILD in secp256k1.c directly.
[bitcoin-core/secp256k1#922]: Add mingw32-w64/wine CI build
[bitcoin-core/secp256k1#933]: Avoids a missing brace warning in schnorrsig/tests_impl.h on old compilers
Merge bitcoin-core/secp256k1#936: Fix gen_context/ASM build on ARM
[bitcoin-core/secp256k1#925]: changed include statements without prefix 'include/'
[bitcoin-core/secp256k1#937]: Have ge_set_gej_var, gej_double_var and ge_set_all_gej_var initialize all fields of their outputs.
[bitcoin-core/secp256k1#926]: secp256k1.h: clarify that by default arguments must be != NULL
Merge bitcoin-core/secp256k1#940: contrib: Explain explicit header guards
[bitcoin-core/secp256k1#850]: add secp256k1_ec_pubkey_cmp method
[bitcoin-core/secp256k1#930]: Add ARM32/ARM64 CI
[bitcoin-core/secp256k1#941]: Clean up git tree
[bitcoin-core/secp256k1#846]: ci: Run ASan/LSan and reorganize sanitizer and Valgrind jobs
[bitcoin-core/secp256k1#947]: ci: Run PRs on merge result even for i686
[bitcoin-core/secp256k1#662]: Add ecmult_gen, ecmult_const and ecmult to benchmark
[bitcoin-core/secp256k1#950]: ci: Add ppc64le build

This PR can be recreated  with `./contrib/sync-upstream.sh range 1758a92f`.

@jesseposner I had to change one of the test cases in the ECDSA adaptor sig module (see commit message). Does that still test what you wanted to test?


real-or-random:
We need to tweak the cirrus config.

Ubsan builds: `Container errored with 'OOMKilled'`... So maybe increase the memory to 2 GB here.
Valgrind builds: apparently too many iterations for our tests...

real-or-random:
Hm, still timeout... It's also possible to double the timeouts: https://cirrus-ci.org/faq/#instance-timed-out

jonasnick:
Cool, that worked!

jesseposner:
@jonasnick If you run a `gcov` test coverage report with these changes, we are missing coverage for the deserialization error at https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/src/modules/ecdsa_adaptor/main_impl.h#L245 and the serialization error at https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/src/modules/ecdsa_adaptor/main_impl.h#L343.

jonasnick:
@jesseposner Line 245 has coverage with (and without this PR) according to my test run. Also, I don't know how the commit could have affected that line. The error at line 343 is unreachable now, so I added an explanation as a comment and  removed the test case entirely.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -341,6 +341,13 @@ int secp256k1_ecdsa_adaptor_recover(const secp256k1_context* ctx, unsigned char
      * branch point. */
     secp256k1_declassify(ctx, &enckey_expected_ge, sizeof(enckey_expected_ge));
     if (!secp256k1_eckey_pubkey_serialize(&enckey_expected_ge, enckey_expected33, &size, SECP256K1_EC_COMPRESSED)) {
+        /* Unreachable
+         * Proof:
+         *     eckey_pubkey_serialize fails <=> deckey = 0
+         *     deckey = 0 <=> s^-1 = 0 or sp = 0
+         *     case 1: s^-1 = 0 impossible by the definition of multiplicative inverse
```
jesseposner, line 348: While it's true that it's definitionally impossible, if `secp256k1_scalar_inverse` doesn't check for an overflow, then it can return `0` when `s` is equal to the curve order `q`. Neither the [implementation](https://github.com/ElementsProject/secp256k1-zkp/blob/f3708a1ecb445b1b05a0f8fcd1da6a88f83d89c4/src/scalar_impl.h#L68) on master nor the `scalar_low_impl.h` [implementation](https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-e04f40e44c1c6158458b5cfe25b01d18645b047d393d4bc5ae8c90f55d1803dfR130) implementation in this PR check for an overflow.

This PR adds a new `secp256k1_scalar_inverse` implementation in [scalar_4x64_impl.h](https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-334ffa39c358290caf0eadd32018e5d9276ea2e76416f28a34b981e59b56a9d9R933) and [scalar_8x32_impl.h](https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-4112c4b434fd39c8722830f9daf1db7b2feb89cfd73a69c57c4a575575ee9cedR811). This new implementation calls to either [secp256k1_scalar_to_signed62](https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-334ffa39c358290caf0eadd32018e5d9276ea2e76416f28a34b981e59b56a9d9R913) or [secp256k1_scalar_to_signed30](https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-4112c4b434fd39c8722830f9daf1db7b2feb89cfd73a69c57c4a575575ee9cedR786), which in turn check for an overflow (at https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-334ffa39c358290caf0eadd32018e5d9276ea2e76416f28a34b981e59b56a9d9R918 and https://github.com/ElementsProject/secp256k1-zkp/pull/132/files#diff-4112c4b434fd39c8722830f9daf1db7b2feb89cfd73a69c57c4a575575ee9cedR792 respectively), but only if the `VERIFY` macro is defined.

jonasnick: But `s` doesn't overflow (it was just loaded from the ecdsa signature).

jesseposner: The ecdsa signature can contain an `s` that is equal to the curve order `q`. When this happens in the implementation on master then `secp256k1_scalar_inverse` will write `0` to `deckey`, and the error at line 343 will be reached. With this PR, that is no longer the case because the `secp256k1_scalar_inverse` implementation has changed such that, if the `VERIFY` macro is defined then the following check will fail `VERIFY_CHECK(secp256k1_scalar_check_overflow(r) == 0);`.

I also just checked to see what happens if `VERIFY` is not defined, and with the new implementation `secp256k1_scalar_inverse` writes `1` to `deckey` instead of `0`, so even without `VERIFY`, the error at line 343 will not be reached in the new implementation. So it is correct to say that line 343 is unreachable with the new implementation (assuming `scalar_low_impl.h` is not used). But this depends on the implementation, and is not impossible, as is demonstrated by the test case on master which reaches line 343.

src/modules/ecdsa_adaptor/tests_impl.h
```diff
@@ -1105,12 +1105,6 @@ void adaptor_tests(void) {
         secp256k1_ecdsa_signature sig_tmp;
```
jesseposner, line 1105: `sig_tmp` is no longer used with the removal of the test at 1122.

jesseposner:
@jonasnick I'm finding that line 245 doesn't have coverage consistently because of the buffer fix at https://github.com/ElementsProject/secp256k1-zkp/pull/132/commits/01dfb69ab1065425fd1753f224a21b48596f9731. It can get coverage depending on whether the random bit flip causes deserialization to fail, but it doesn't always cause it to fail.

The small buffer was causing the deserialization to fail, although that was undermining the purpose of the random bit flip test, so it's good that you fixed it. To make sure we hit the deserialization error perhaps another test case should be added:

```c
    {
        /* Test failed adaptor sig deserialization */
        unsigned char adaptor_sig_tmp[162];
        memcpy(adaptor_sig_tmp, adaptor_sig, sizeof(adaptor_sig_tmp));
        memset(&adaptor_sig_tmp, 0xFF, 162);
        CHECK(secp256k1_ecdsa_adaptor_verify(ctx, adaptor_sig_tmp, &pubkey, msg, &enckey) == 0);
    }
```

src/modules/ecdsa_adaptor/tests_impl.h
jonasnick: fixed

src/modules/ecdsa_adaptor/main_impl.h
jonasnick: Ok, I tweaked the comment such that it says that it's unreachable by the tests because scalar_inverse checks for valid inputs in that case (since VERIFY is defined).

jonasnick:
@jesseposner thanks for figuring this out. I added your suggested test case.

src/modules/ecdsa_adaptor/main_impl.h
```diff
@@ -341,6 +341,15 @@ int secp256k1_ecdsa_adaptor_recover(const secp256k1_context* ctx, unsigned char
      * branch point. */
     secp256k1_declassify(ctx, &enckey_expected_ge, sizeof(enckey_expected_ge));
     if (!secp256k1_eckey_pubkey_serialize(&enckey_expected_ge, enckey_expected33, &size, SECP256K1_EC_COMPRESSED)) {
+        /* Unreachable in tests
```
real-or-random, line 344: I don't understand. Do you mean "Unreachable in VERIFY builds"? 

If yes, I think we should avoid this. In other words: If `scalar_inverse` expects valid scalars, we should always give it valid scalars.

real-or-random: Ah so it's "unreachable if the user doesn't fiddle around with the data in ecdsa_signature" (but we still return 0 because it's the safest thing to do)?

I guess that's fine but maybe the comment can be improved. It was not clear to me. Maybe mention why this is unreachable in production and mention test coverage.

jonasnick: Mentioned test coverage in the comment.

real-or-random (approved):
ACK f09497ea3e07d7a730a6ff3479dca18b848ef729

PR #144, opened 2021-09-15
Upstream PRs 969, 956, 783, 976

jonasnick:
[bitcoin-core/secp256k1#969]: ci: Fixes after Debian release
[bitcoin-core/secp256k1#956]: Replace ecmult_context with a generated static array.
[bitcoin-core/secp256k1#783]: Make the public API docs more consistent and explicit
[bitcoin-core/secp256k1#976]: `secp256k1_schnorrsig_sign_internal` should be static

This PR can be recreated  with `./sync-upstream.sh range 2a3a97c665475bc00d5d60f2f04830202983a631`.


src/modules/rangeproof/tests_impl.h
```diff
@@ -138,44 +137,42 @@ static void test_rangeproof_api(const secp256k1_context *none, const secp256k1_c
...
+        CHECK(*ecount == 15);
     }
     {
         uint64_t min_value;
         uint64_t max_value;
-        CHECK(secp256k1_rangeproof_verify(none, &min_value, &max_value, &commit, proof, len, ext_commit, ext_commit_len, secp256k1_generator_h) == 0);
-        CHECK(*ecount == 17);
-        CHECK(secp256k1_rangeproof_verify(sign, &min_value, &max_value, &commit, proof, len, ext_commit, ext_commit_len, secp256k1_generator_h) == 0);
-        CHECK(*ecount == 18);
+        CHECK(secp256k1_rangeproof_verify(none, &min_value, &max_value, &commit, proof, len, ext_commit, ext_commit_len, secp256k1_generator_h) == 1);
```
apoelstra, line 155: lol this line should definitely fail. How did CI not catch this?

apoelstra: Oh, actually, I am wrong, this was changed because of the upstream elimination of the verify context.

real-or-random: Yeah, we're in an intermediate state where we made this (backwards-compatible) API change but haven't documented it yet...

PR #145, opened 2021-09-15
sync-upstream: fix quoting

jonasnick:
Otherwise strings in $TITLE and $BODAY that are enclosed in ` are executed in
gh-pr-create.sh.

### DIFF
--- a/contrib/sync-upstream.sh
+++ b/contrib/sync-upstream.sh
@@ -100,11 +100,20 @@ git checkout master
 git pull
 git checkout -b temp-merge-"$PRNUM"
 
+# Escape single quote
+# ' -> '\''
+quote() {
+    local quoted=${1//\'/\'\\\'\'}
+    printf "%s" "$quoted"
+}
+TITLE=$(quote "$TITLE")
+BODY=$(quote "$BODY")
+
 BASEDIR=$(dirname "$0")
 FNAME="$BASEDIR/gh-pr-create.sh"
 cat <<EOT > "$FNAME"
 #!/bin/sh
-gh pr create -t "$TITLE" -b "$BODY" --web
+gh pr create -t '$TITLE' -b '$BODY' --web
 # Remove temporary branch
 git checkout master
 git branch -D temp-merge-"$PRNUM"


real-or-random:
Oh, is this really an issue? Do you have an example case?

I always assumed double-quoting is fine. https://stackoverflow.com/questions/4273074/sanitize-user-input-in-bash-for-security-purposes

jonasnick:
It is fine, but we're doing something different here: write into a file that we execute later.
An example is calling ./sync-upstream.sh on master. The generated gh-pr-create.sh file contains the string ``secp256k1_schnorrsig_sign_internal`` which will be executed.
A better solution is to write the PR title and body into a file and add a command to sync-upstream.sh that reads the file and executes the gh command.

contrib/sync-upstream.sh
```diff
@@ -100,11 +100,20 @@ git checkout master
...
+    local quoted=${1//\'/\'\\\'\'}
+    printf "%s" "$quoted"
+}
+TITLE=$(quote "$TITLE")
+BODY=$(quote "$BODY")
+
 BASEDIR=$(dirname "$0")
 FNAME="$BASEDIR/gh-pr-create.sh"
 cat <<EOT > "$FNAME"
 #!/bin/sh
```
real-or-random, line 115:
```suggestion
#!/bin/bash
```

`${1// ... }` is a bashism, see https://stackoverflow.com/a/13210909 . (This has a few POSIX solutions but no need to bother, `sync-upstream.sh` is already bash-specific.)

real-or-random (commented):
Ok sure, that makes sense.

ACK mod nit

contrib/sync-upstream.sh
jonasnick: `sync-upstream.sh` already uses bash. The suggested line change doesn't affect `${1// ... }`.

real-or-random: Oh man, this PR really confuses me for no reason.  You're right of course.

real-or-random:
ACK 95ee1fa0303fc72ca10f594db5b1c4a8551fed3d

contrib/sync-upstream.sh
jonasnick: It's mostly just my dumb design of writing yet another shell script to execute. I should fix this at some point.

PR #149, opened 2021-10-17
rangeproof: add a bunch more testing

apoelstra:
Add two new fixed rangeproof vectors; check that various extracted values are correct; add a test for creating and verifying single-value proofs.

I am going to do some pretty invasive work on the rangeproof logic to fix the excessive stack usage and we should have better fixed-vector tests to detect any mistakes.


src/modules/rangeproof/tests_impl.h
```diff
@@ -541,8 +541,95 @@ static void test_rangeproof(void) {
...
+    secp256k1_pedersen_commitment commit;
+    unsigned char blind[32];
+    unsigned char blind_out[32];
+    unsigned char nonce[32];
+    const unsigned char message[1] = " "; /* no message will fit into a single-value proof */
+    unsigned char message_out[sizeof(proof)] = { 0 };
+    size_t plen = sizeof(proof);
+    size_t min_val_out = 0;
+    size_t max_val_out = 0;
+    size_t val_out = 0;
```
real-or-random, line 555:
```suggestion
    uint64_t min_val_out = 0;
    uint64_t max_val_out = 0;
    uint64_t val_out = 0;
```

apoelstra: Thanks. I'm disappointed that my local gcc didn't catch that.

apoelstra:
Closing since this is included in #195

PR #131, opened 2021-05-13
Replace MuSig(1) module with MuSig2

jonasnick:
The main commit comprises `905 insertions(+), 1253 deletions(-)`. The diff isn't as small as I had hoped, but that's mostly because it was possible to simplify the API quite substantially which required rewriting large parts. Sorry, almost all of the changes are in one big commit which makes the diff very hard to read. Perhaps best to re-review most parts from scratch.

A few key changes:

- Obviously no commitment round. No big session struct and no `verifier` sessions. No `signer` struct.
- There's a new `secnonce` struct that is the output of musig_nonce_gen and derived from a uniformly random session_id32. The derivation can be strengthened by adding whatever session parameters (combined_pk, msg) are available. The nonce function is my ad-hoc construction that allows for these optional inputs. Please have a look at that.
- The secnonce is made invalid after being used in partial_sign.
- Adaptor signatures basically work as before, according to https://github.com/ElementsProject/scriptless-scripts/pull/24 (with the exception that they operate on aggregate instead of partial sigs)
- To avoid making this PR overly complex I did not consider how this implementation interacts with nested-MuSig, sign-to-contract, and antiklepto.
- Testing should be close to complete. There's no reachable line or branch that isn't exercised by the tests.
- [x] ~In the current implementation when a signer sends an invalid nonce (i.e. some garbage that can't be mapped to a group element), it is ignored when combining nonces. Only after receiving the signers partial signature and running `partial_sig_verify` will we notice that the signer misbehaved. The reason for this is that 1) this makes the API simpler and 2) malicious peers don't gain any additional powers because they can always interrupt the protocol by refusing to sign. However, this is up for discussion.~ EDIT: this is not the case anymore since invalid nonces are rejected when they're parsed.
- [x] ~For every partial signature we verify we have to parse the pubnonce (two compressed points), despite having parsed it in `process_nonces` already. This is not great. `process_nonces` could optionally output the array of parsed pubnonces.~ EDIT: fixed by having a dedicated type for nonces.
- [x] ~I left `src/modules/musig/musig.md` unchanged for now. Perhaps we should merge it with the `musig-spec`~ EDIT: musig.md is updated
- [x] partial verification should use multiexp to compute `R1 + b*R2 + c*P`, but this can be done in a separate PR
- [x] renaming wishlist
    - pre_session -> keyagg_cache (because there is no session anymore)
[...]

PR #151, opened 2021-10-26
MuSig: Add Minimal Compatibility with BIP32 Tweaking

jonasnick:
In short, `musig_pubkey_tweak_add` now allows for xonly _and_ "ordinary" tweaking. Also, in order to allow using `ec_pubkey_tweak_add` on the non-xonly aggregate public key, there's a new function `musig_pubkey_get` that allows obtaining it from the `keyagg_cache`.

One alternative would be that instead of adding `musig_pubkey_get`, we could change `pubkey_agg` to output an ordinary (non-xonly) pubkey. Then users of the API who do not need ordinary (BIP32) tweaking would be forced to call `xonly_pubkey_from_pubkey`. And we'd probably want to change the spec. And it would be a bit weird to output a pubkey that can't be directly schnorrsig_verify'd.

Based on #131

### DIFF
--- a/examples/musig.c
+++ b/examples/musig.c
@@ -52,14 +52,51 @@ int create_keypair(const secp256k1_context* ctx, struct signer_secrets *signer_s
     return 1;
 }
 
+/* Tweak the pubkey corresponding to the provided keyagg cache, update the cache
+ * and return the tweaked aggregate pk. */
+int tweak(const secp256k1_context* ctx, secp256k1_xonly_pubkey *agg_pk, secp256k1_musig_keyagg_cache *cache) {
+    secp256k1_pubkey output_pk;
+    unsigned char ordinary_tweak[32] = "this could be a BIP32 tweak....";
+    unsigned char xonly_tweak[32] = "this could be a taproot tweak..";
+
+
+    /* Ordinary tweaking which, for example, allows deriving multiple child
+     * public keys from a single aggregate key using BIP32 */
+    if (!secp256k1_musig_pubkey_ec_tweak_add(ctx, NULL, cache, ordinary_tweak)) {
+        return 0;
+    }
+    /* Note that we did not provided an output_pk argument, because the
+     * resulting pk is also saved in the cache and so if one is just interested
+     * in signing the output_pk argument is unnecessary. On the other hand, if
+     * one is not interested in signing, the same output_pk can be obtained by
+     * calling `secp256k1_musig_pubkey_get` right after key aggregation to get
+     * the full pubkey and then call `secp256k1_ec_pubkey_tweak_add`. */
+
+    /* Xonly tweaking which, for example, allows creating taproot commitments */
+    if (!secp256k1_musig_pubkey_xonly_tweak_add(ctx, &output_pk, cache, xonly_tweak)) {
+        return 0;
+    }
+    /* Note that if we wouldn't care about signing, we can arrive at the same
+     * output_pk by providing the untweaked public key to
+     * `secp256k1_xonly_pubkey_tweak_add` (after converting it to an xonly pubkey
+     * if necessary with `secp256k1_xonly_pubkey_from_pubkey`). */
+
+    /* Now we convert the output_pk to an xonly pubkey to allow to later verify
+     * the Schnorr signature against it. For this purpose we can ignore the
+     * `pk_parity` output argument; we would need it if we would have to open
+     * the taproot commitment. */
+    if (!secp256k1_xonly_pubkey_from_pubkey(ctx, agg_pk, NULL, &output_pk)) {
+        return 0;
+    }
+    return 1;
+}
+
 /* Sign a message hash with the given key pairs and store the result in sig */
-int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, struct signer *signer, const unsigned char* msg32, unsigned char *sig64) {
+int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, struct signer *signer, const secp256k1_musig_keyagg_cache *cache, const unsigned char *msg32, unsigned char *sig64) {
     int i;
-    const secp256k1_xonly_pubkey *pubkeys[N_SIGNERS];
     const secp256k1_musig_pubnonce *pubnonces[N_SIGNERS];
     const secp256k1_musig_partial_sig *partial_sigs[N_SIGNERS];
     /* The same for all signers */
-    secp256k1_musig_keyagg_cache cache;
     secp256k1_musig_session session;
 
     for (i = 0; i < N_SIGNERS; i++) {
@@ -86,7 +123,6 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
         if (!secp256k1_musig_nonce_gen(ctx, &signer_secrets[i].secnonce, &signer[i].pubnonce, session_id, seckey, msg32, NULL, NULL)) {
             return 0;
         }
-        pubkeys[i] = &signer[i].pubkey;
         pubnonces[i] = &signer[i].pubnonce;
     }
     /* Communication round 1: A production system would exchange public nonces
@@ -94,21 +130,18 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
     for (i = 0; i < N_SIGNERS; i++) {
         secp256k1_musig_aggnonce agg_pubnonce;
 
-        /* Create aggregate pubkey, aggregate nonce and initialize signer data */
-        if (!secp256k1_musig_pubkey_agg(ctx, NULL, NULL, &cache, pubkeys, N_SIGNERS)) {
-            return 0;
-        }
+        /* Create aggregate nonce and initialize the session */
         if (!secp256k1_musig_nonce_agg(ctx, &agg_pubnonce, pubnonces, N_SIGNERS)) {
             return 0;
         }
-        if (!secp256k1_musig_nonce_process(ctx, &session, &agg_pubnonce, msg32, &cache, NULL)) {
+        if (!secp256k1_musig_nonce_process(ctx, &session, &agg_pubnonce, msg32, cache, NULL)) {
             return 0;
         }
         /* partial_sign will clear the secnonce by setting it to 0. That's because
          * you must _never_ reuse the secnonce (or use the same session_id to
          * create a secnonce). If you do, you effectively reuse the nonce and
          * leak the secret key. */
-        if (!secp256k1_musig_partial_sign(ctx, &signer[i].partial_sig, &signer_secrets[i].secnonce, &signer_secrets[i].keypair, &cache, &session)) {
+        if (!secp256k1_musig_partial_sign(ctx, &signer[i].partial_sig, &signer_secrets[i].secnonce, &signer_secrets[i].keypair, cache, &session)) {
             return 0;
         }
         partial_sigs[i] = &signer[i].partial_sig;
@@ -127,7 +160,7 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
          * fine to first verify the aggregate sig, and only verify the individual
          * sigs if it does not work.
          */
-        if (!secp256k1_musig_partial_sig_verify(ctx, &signer[i].partial_sig, &signer[i].pubnonce, &signer[i].pubkey, &cache, &session)) {
+        if (!secp256k1_musig_partial_sig_verify(ctx, &signer[i].partial_sig, &signer[i].pubnonce, &signer[i].pubkey, cache, &session)) {
             return 0;
         }
     }
@@ -141,6 +174,7 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
     struct signer signers[N_SIGNERS];
     const secp256k1_xonly_pubkey *pubkeys_ptr[N_SIGNERS];
     secp256k1_xonly_pubkey agg_pk;
+    secp256k1_musig_keyagg_cache cache;
     unsigned char msg[32] = "this_could_be_the_hash_of_a_msg!";
     unsigned char sig[64];
 
@@ -156,13 +190,21 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
     }
     printf("ok\n");
     printf("Combining public keys...");
-    if (!secp256k1_musig_pubkey_agg(ctx, NULL, &agg_pk, NULL, pubkeys_ptr, N_SIGNERS)) {
+    /* If you just want to aggregate and not sign the cache can be NULL */
+    if (!secp256k1_musig_pubkey_agg(ctx, NULL, &agg_pk, &cache, pubkeys_ptr, N_SIGNERS)) {
+        printf("FAILED\n");
+        return 1;
+    }
+    printf("ok\n");
+    printf("Tweaking................");
+    /* Optionally tweak the aggregate key */
+    if (!tweak(ctx, &agg_pk, &cache)) {
         printf("FAILED\n");
         return 1;
     }
     printf("ok\n");
     printf("Signing message.........");
-    if (!sign(ctx, signer_secrets, signers, msg, sig)) {
+    if (!sign(ctx, signer_secrets, signers, &cache, msg, sig)) {
         printf("FAILED\n");
         return 1;
     }
--- a/include/secp256k1_musig.h
+++ b/include/secp256k1_musig.h
@@ -223,6 +223,24 @@ SECP256K1_API int secp256k1_musig_pubkey_agg(
     size_t n_pubkeys
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(5);
 
+/** Obtain the aggregate public key from a keyagg_cache.
+ *
+ *  This is only useful if you need the non-xonly public key, in particular for
+ *  ordinary (non-xonly) tweaking or batch-verifying multiple key aggregations
+ *  (not implemented).
+ *
+ *  Returns: 0 if the arguments are invalid, 1 otherwise
+ *  Args:        ctx: pointer to a context object
+ *  Out:      agg_pk: the MuSig-aggregated public key.
+ *  In: keyagg_cache: pointer to a `musig_keyagg_cache` struct initialized by
+ *                    `musig_pubkey_agg`
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
+    const secp256k1_context* ctx,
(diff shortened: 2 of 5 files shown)


jonasnick:
As @real-or-random remarked, instead of an `xonly` argument it's probably better to have two functions, `ec_tweak` and `xonly_tweak`, because they'd correspond to the existing tweaking functions.

jonasnick:
rebased

robot-dreams (commented):
Nice, looks good on a first pass. I'm convinced that if you generate an aggregate pk, apply ordinary tweak to derive a BIP32 child key, then apply an x-only tweak to derive a BIP341 output key, you will get the correct result.

Before I finalize my review, I still want to:
- [x] Check changes to `tests_impl.h`
- [ ] Try adding a test vector with a BIP32 tweak
- [ ] Think about if there's a simpler approach (e.g. keeping track of different invariants)

Regarding interaction with BIP32, am I understanding correctly that:
- To sign for a descendent pubkey, you just start with the aggregate pubkey (root of the tree), apply `ec_tweak_add` repeatedly with the `parse_256(I_L)` value at each level of the tree, and then proceed with the protocol as usual
- It generally shouldn't be possible to derive a hardened child (nor is this desirable) because no individual signer should have the aggregate *private* key

src/modules/musig/keyagg_impl.h
```diff
@@ -277,7 +277,7 @@ int secp256k1_musig_pubkey_tweak_add(const secp256k1_context* ctx, secp256k1_pub
     if (overflow) {
         return 0;
     }
-    if (secp256k1_extrakeys_ge_even_y(&cache_i.pk)) {
+    if (xonly && secp256k1_extrakeys_ge_even_y(&cache_i.pk)) {
```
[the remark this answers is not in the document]
jonasnick: To ease maintainability we try to avoid changing files that we didn't add in secp-zkp. `secp256k1_extrakeys_ge_even_y` is a function we get from upstream, so we should change it upstream.

jonasnick:
@robot-dreams Your understanding appears to match mine. As an aside, hardened derivation is [not impossible](https://hackmd.io/@elichai/ry082c4QK).

src/modules/musig/session_impl.h
```diff
@@ -538,56 +538,65 @@ int secp256k1_musig_partial_sign(const secp256k1_context* ctx, secp256k1_musig_p
...
+     *     P_agg[i] := f(i, P_agg[i-1]) + t[i]*G for i = 1, ..., m
+     *       where f(i, X) := |X| if xonly(i)
+     *             f(i, X) := X   otherwise
      *
      * Note that our goal is to produce a partial signature corresponding to
-     * the final public key after m tweaking operations P_final = |P_tweak[m]|.
+     * the final public key after m tweaking operations P_final = |P_agg[m]|.
      *
-     * Define d[i], d_agg, and d_tweak[i] so that:
+     * Define d[i] for 1 <= i <= n-1 and d_agg[i] for 0 <= i <= m so that:
```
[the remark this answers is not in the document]
jonasnick: fixed, thanks!

src/modules/musig/tests_impl.h
```diff
@@ -857,22 +873,35 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
     CHECK(secp256k1_musig_pubkey_agg(ctx, scratch, &P_xonly[0], &keyagg_cache, pk_ptr, 2) == 1);
     musig_tweak_test_helper(&P_xonly[0], sk[0], sk[1], &keyagg_cache);
+    CHECK(secp256k1_musig_pubkey_get(ctx, &P[0], &keyagg_cache));
 
-    /* Compute Pi = |Pj| + tweaki*G where where j = i-1 and try signing for
-     * that key. The function |.| normalizes the point to have an even
-     * X-coordinate. This results in ordinary "xonly-tweaking". */
+    /* Compute Pi = f(Pj) + tweaki*G where where j = i-1 and try signing for
+     * that key. If xonly is set to true, the function f is normalizes the input
+     * point to have an even X-coordinate (ordinary "xonly-tweaking").
```
robot-dreams, line 880: Since there's already "ordinary" vs. "xonly" tweaking, would it be more clear to just say "xonly-tweaking" here?

robot-dreams (commented):
ACK 7474bc15d85b6ac04e36acd432158608e4fb7ced aside from remaining comments

I added a test vector for ordinary tweaking, feel free to grab the commit (see https://github.com/jonasnick/secp256k1-zkp/pull/2) or just copy the vectors directly

I haven't thought about a way to simplify the invariants, but in retrospect that should probably happen in a different PR (if at all) instead of blocking this one. By the way I think it was a good idea to define `f(i, P_agg[i - 1])` in the explanation.

examples/musig.c
```diff
@@ -52,14 +52,51 @@ int create_keypair(const secp256k1_context* ctx, struct signer_secrets *signer_s
     return 1;
 }
 
+/* Tweak the pubkey corresponding to the provided keyagg cache, update the cache
+ * and return the tweaked aggregate pk. */
+int tweak(const secp256k1_context* ctx, secp256k1_xonly_pubkey *agg_pk, secp256k1_musig_keyagg_cache *cache) {
+    secp256k1_pubkey output_pk;
+    unsigned char ordinary_tweak[32] = "this could be a BIP32 tweak..";
```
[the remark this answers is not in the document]
jonasnick: fixed

src/modules/musig/tests_impl.h
jonasnick: Yeah that's better

jonasnick:
Thanks @robot-dreams ! I added your commit.

robot-dreams:
ACK 560aa3d3a98081963d3c5d276773d20e07a3932d

jonasnick:
Squashed fixups.

include/secp256k1_musig.h
```diff
@@ -223,6 +223,24 @@ SECP256K1_API int secp256k1_musig_pubkey_agg(
...
+ *
+ *  This is only useful if you need the non-xonly public key, in particular for
+ *  ordinary (non-xonly) tweaking or batch-verifying multiple key aggregations
+ *  (not implemented).
+ *
+ *  Returns: 0 if the arguments are invalid, 1 otherwise
+ *  Args:        ctx: pointer to a context object
+ *  Out:      agg_pk: the MuSig-aggregated public key.
+ *  In: keyagg_cache: pointer to a key aggregation cache initialized in
+ *                    `musig_pubkey_agg`
```
real-or-random, line 236:
```suggestion
 *  In: keyagg_cache: pointer to a `musig_keyagg_cache` struct initialized by
 *                    `musig_pubkey_agg`
```

to match the existing docs.

include/secp256k1_musig.h
```diff
@@ -223,6 +223,24 @@ SECP256K1_API int secp256k1_musig_pubkey_agg(
     size_t n_pubkeys
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(5);
 
+/** Obtain the aggregate public key from a keyagg_cache.
+ *
+ *  This is only useful if you need the non-xonly public key, in particular for
+ *  ordinary (non-xonly) tweaking or batch-verifying multiple key aggregations
+ *  (not implemented).
```
real-or-random, line 230: (not this PR:) We should think about the terminology here. "Non-xonly" is totally fine to get this merged but I think it can be improved.

src/modules/musig/tests_impl.h
```diff
@@ -243,6 +244,15 @@ void musig_api_tests(secp256k1_scratch_space *scratch) {
...
     CHECK(secp256k1_musig_pubkey_agg(vrfy, scratch, &agg_pk, &keyagg_cache, pk_ptr, 2) == 1);
 
+    /* pubkey_get */
+    ecount = 0;
+    CHECK(secp256k1_musig_pubkey_get(none, &full_agg_pk, &keyagg_cache) == 1);
+    CHECK(secp256k1_musig_pubkey_get(none, NULL, &keyagg_cache) == 0);
+    CHECK(ecount == 1);
+    CHECK(secp256k1_musig_pubkey_get(none, &full_agg_pk, NULL) == 0);
+    CHECK(ecount == 2);
+    CHECK(memcmp(&full_agg_pk, zeros68, sizeof(full_agg_pk)) == 0);
```
real-or-random, line 254:
```suggestion
    CHECK(secp256k1_memcmp_var(&full_agg_pk, zeros68, sizeof(full_agg_pk)) == 0);
```

real-or-random (commented):
Got interrupted, but let me leave some nits on the first commit here. I will continue later

include/secp256k1_musig.h
```diff
@@ -241,16 +241,59 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
...
+ *  The tweaking method is the same as `secp256k1_ec_pubkey_tweak_add`. So after
+ *  the following pseudocode buf and buf2 have identical contents (absent
+ *  earlier failures).
+ *
+ *  secp256k1_musig_pubkey_agg(..., keyagg_cache, pubkeys, ...)
+ *  secp256k1_musig_pubkey_get(..., agg_pk, keyagg_cache)
+ *  secp256k1_musig_pubkey_ec_tweak_add(..., output_pk, tweak32, keyagg_cache)
+ *  secp256k1_ec_pubkey_serialize(..., buf, output_pk)
+ *  secp256k1_ec_pubkey_tweak_add(..., agg_pk, tweak32)
+ *  secp256k1_ec_pubkey_serialize(..., buf2, agg_pk)
```
real-or-random, line 257: I think it's easier to follow in this order:
```suggestion
 *  secp256k1_musig_pubkey_agg(..., keyagg_cache, pubkeys, ...)
 *  secp256k1_musig_pubkey_ec_tweak_add(..., output_pk, tweak32, keyagg_cache)
 *  secp256k1_ec_pubkey_serialize(..., buf, output_pk)
 * 
 *  secp256k1_musig_pubkey_get(..., agg_pk, keyagg_cache)
 *  secp256k1_ec_pubkey_tweak_add(..., agg_pk, tweak32)
 *  secp256k1_ec_pubkey_serialize(..., buf2, agg_pk)
```

include/secp256k1_musig.h
```diff
@@ -241,16 +241,59 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
...
+ *  to create a signature for it, you can just use
+ *  `secp256k1_ec_pubkey_tweak_add`.
+ *
+ *  Returns: 0 if the arguments are invalid or the resulting public key would be
+ *           invalid (only when the tweak is the negation of the corresponding
+ *           secret key). 1 otherwise.
+ *  Args:            ctx: pointer to a context object initialized for verification
+ *  Out:   output_pubkey: pointer to a public key to store the result. Will be set
+ *                        to an invalid value if this function returns 0. If you
+ *                        do not need it, this arg can be NULL.
```
real-or-random, line 270:
```suggestion
 *  Out:   output_pubkey: pointer to a public key to store the result (can be NULL).
 *                        Will be set to an invalid value if this function returns 0.
```
My eyes were grepping for "(can be NULL)" with the brackets and I felt some unease. I think it's good to be this consistent.

edit: Ahh I notice this part and other are copied from below.  Well I suggest either ignore it or change it below. Sorry, can't "suggest" the code for the other function here (not in context of changed lines). Next time I'll just make a nit comment on top of it...

include/secp256k1_musig.h
```diff
@@ -241,16 +241,59 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
...
+ *  In/Out: keyagg_cache: pointer to a `musig_keyagg_cache` struct initialized by
+ *                       `musig_pubkey_agg`
+ *  In:          tweak32: pointer to a 32-byte tweak. If the tweak is invalid
+ *                        according to secp256k1_ec_seckey_verify, this function
```
real-or-random, line 274:
```suggestion
 *                        according to `secp256k1_ec_seckey_verify`, this function
```

include/secp256k1_musig.h
```diff
@@ -241,16 +241,59 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_pubkey_get(
...
  *
  *  The tweaking method is the same as `secp256k1_xonly_pubkey_tweak_add`. So in
  *  the following pseudocode xonly_pubkey_tweak_add_check (absent earlier
  *  failures) returns 1.
  *
  *  secp256k1_musig_pubkey_agg(..., agg_pk, keyagg_cache, pubkeys, ...)
- *  secp256k1_musig_pubkey_tweak_add(..., output_pubkey, tweak32, keyagg_cache)
- *  secp256k1_xonly_pubkey_serialize(..., buf, output_pubkey)
+ *  secp256k1_musig_pubkey_xonly_tweak_add(..., output_pk, tweak32, keyagg_cache)
+ *  secp256k1_xonly_pubkey_serialize(..., buf, output_pk)
```
real-or-random, line 296: Why change this (and above)?

examples/musig.c
```diff
@@ -86,29 +123,25 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
...
     for (i = 0; i < N_SIGNERS; i++) {
         secp256k1_musig_aggnonce agg_pubnonce;
 
-        /* Create aggregate pubkey, aggregate nonce and initialize signer data */
-        if (!secp256k1_musig_pubkey_agg(ctx, NULL, NULL, &cache, pubkeys, N_SIGNERS)) {
-            return 0;
-        }
-        if (!secp256k1_musig_nonce_agg(ctx, &agg_pubnonce, pubnonces, N_SIGNERS)) {
+        /* Create aggregate nonce and initialize the session */
+        if(!secp256k1_musig_nonce_agg(ctx, &agg_pubnonce, pubnonces, N_SIGNERS)) {
```
real-or-random, line 134:
```suggestion
        if (!secp256k1_musig_nonce_agg(ctx, &agg_pubnonce, pubnonces, N_SIGNERS)) {
```

examples/musig.c
```diff
@@ -156,13 +190,21 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
     }
     printf("ok\n");
     printf("Combining public keys...");
-    if (!secp256k1_musig_pubkey_agg(ctx, NULL, &agg_pk, NULL, pubkeys_ptr, N_SIGNERS)) {
+    if (!secp256k1_musig_pubkey_agg(ctx, NULL, &agg_pk, &cache, pubkeys_ptr, N_SIGNERS)) {
+        /* If you just want to aggregate and not sign the cache can be NULL */
```
real-or-random, line 194:
```suggestion 
    /* If you just want to aggregate and not sign the cache can be NULL */
    if (!secp256k1_musig_pubkey_agg(ctx, NULL, &agg_pk, &cache, pubkeys_ptr, N_SIGNERS)) {
```
(btw github's diff algorithm is really stupid, I'm just swapping lines here except indentation.)

examples/musig.c
```diff
@@ -52,14 +52,51 @@ int create_keypair(const secp256k1_context* ctx, struct signer_secrets *signer_s
...
+     * calling `secp256k1_musig_pubkey_get` right after key aggregation to get
+     * the full pubkey and then call `secp256k1_ec_pubkey_tweak_add`. */
+
+    /* Xonly tweaking which, for example, allows creating taproot commitments */
+    if (!secp256k1_musig_pubkey_xonly_tweak_add(ctx, &output_pk, cache, xonly_tweak)) {
+        return 0;
+    }
+    /* Note that if we wouldn't care about signing, we can arrive at the same
+     * output_pk by providing the untweaked public key to
+     * `secp256k1_xonly_pubkey_tweak_add` (after converting it to an xonly ubkey
```
real-or-random, line 81:
```suggestion
     * `secp256k1_xonly_pubkey_tweak_add` (after converting it to an xonly pubkey
```

real-or-random (commented):
ACK mod nits.

Really all of this are only nits. It's good that we have this example code, it's very helpful.

real-or-random:
We're also making progress towards a better understanding of when to use which tweaks, and then how to deal with them during signing. I think we should write that up somewhere.

include/secp256k1_musig.h
jonasnick: I don't think this is correct. `pubkey_get` will return the already tweaked pk.

include/secp256k1_musig.h
jonasnick: `(can be NULL)` is imho less informative than `If you do not need it, this arg can be NULL.`.

include/secp256k1_musig.h
jonasnick: `secp256k1_musig_pubkey_tweak_add` doesn't exist anymore

real-or-random: What I meant: Why change `output_pubkey` to `output_pk`? Ok it really doesn't matter.

real-or-random:
> Hm, not sure what you mean. The straight forward answer is x-only tweaks for taproot, ordinary tweaks for everything else (including BIP32).

I may have meant "when to use x-only keys in protocols". But I'm not sure to be honest, ignore this part of my comment.

include/secp256k1_musig.h
real-or-random: nevermind.

include/secp256k1_musig.h
real-or-random: ok.

include/secp256k1_musig.h
jonasnick: Ah, sorry. Because it's shorter and everyone knows what a pk is because this is used in a lot of places.

include/secp256k1_musig.h
jonasnick: Perhaps we can find a (tree-wide) wording that's both easily "eye-greppable" and understandable :)

jonasnick:
Thx for the review @real-or-random. Squashed fixups.

robot-dreams:
ACK 8088eddc534cbbb89dd5f892828c4013416c4f2b based on https://github.com/ElementsProject/secp256k1-zkp/pull/151#issuecomment-1005198409 and the following `range-diff`:

```
git range-diff \
	a5b5909e8d56eb3c23aeb6eb4cdca0a621a0930b \
	560aa3d3a98081963d3c5d276773d20e07a3932d \
	8088eddc534cbbb89dd5f892828c4013416c4f2b
```

include/secp256k1_musig.h
real-or-random: I think just putting the entire sentence in parentheses would do the trick. But let's not hold up this PR further.

PR #153, opened 2021-11-02
PoC: Add internals module

jonasnick:
The idea with exposing internal functions is that instead of (only) having experimental modules, we could have experimental _libraries_ that depend on libsecp and live outside of this repo. This has a couple of advantages, such as being able to phase out experimental modules, which simplifies versioning. One downside, however, is that the benefits of libsecp being a single compilation unit is being lost. But another big benefit is that complicated, stateful protocols (such as threshold sigs) can be written in rust (for example), using libsecp internals.

I only played with exposing scalars so far, but the module could also provide `field.h`, `group.h`, `ecmult`, `ecmult_gen`, etc. It would have no API or ABI guarantees. One idea mentioned by @rustyrussell is to make each call fail unless secp256k1_experimental_enable(version) has been called (with the exact version of the library).

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -203,3 +203,8 @@ endif
 if ENABLE_MODULE_ECDSA_ADAPTOR
 include src/modules/ecdsa_adaptor/Makefile.am.include
 endif
+
+if ENABLE_MODULE_INTERNALS
+include src/modules/internals/Makefile.am.include
+endif
+
--- a/configure.ac
+++ b/configure.ac
@@ -191,6 +191,11 @@ AC_ARG_ENABLE(module_ecdsa-adaptor,
     [enable_module_ecdsa_adaptor=$enableval],
     [enable_module_ecdsa_adaptor=no])
 
+AC_ARG_ENABLE(module_internals,
+    AS_HELP_STRING([--enable-module-musig],[enable internals module (experimental)]),
+    [enable_module_internals=$enableval],
+    [enable_module_internals=no])
+
 AC_ARG_ENABLE(external_default_callbacks,
     AS_HELP_STRING([--enable-external-default-callbacks],[enable external default callback functions [default=no]]),
     [use_external_default_callbacks=$enableval],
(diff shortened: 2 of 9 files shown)


real-or-random:
Nice to see a concrete proposal.

The approach to have this a module is pretty lightweight. 

Here are some observations / thoughts:

If we want a

real-or-random:
Nice to see a concrete proposal.

The approach to make this a module is pretty lightweight.

I believe there's agreement that we don't want to provide any API/ABI stability. That implies that callers need to depend on the exact version. With that mind, I wonder why the approach in the PR will be better than vendoring a specific version.

Normally the advantage is that people are able create library X and library Y independently and both X and Y can depend on secp256k1(-zkp) and link to it. But if X and Y need to depend on the exact same version of secp256k1(-zkp), this advantage is rather moot, and they're not really independent.

Another advantage you mention is that we can have foreign language bindings for the internals. But if that is the primary reason, wouldn't it be more straight-forward to create those bindings directly, e.g., for Rust create a rust-secp256k1-internals crate that vendors secp256k1(-zkp), or have an internals feature in rust-secp256k1?

jonasnick:
> they're not really independent.

They're still independent in that they can have separate maintainers, a different focus and more adequate versioning. Also, users are able to pick exactly what they need.

> wouldn't it be more straight-forward to create those bindings directly

What do you mean exactly by more directly? Or rather, in what way is an internals module not direct?

elsirion:
> One idea mentioned by @rustyrussell is to make each call fail unless `secp256k1_experimental_enable(version)` has been called (with the exact version of the library).

Would it be reasonable to give the internal API its own version? That way the check could be SemVer-aware and also allow the use of newer versions if there were no breaking changes. Even if in practice the respective language bindings just vendor a  specific version this could still be a good fail-safe to check for API compatibility. Also tracking changes in an internal API changelog would be very useful for downstream users.

Regarding the exposed scalar functionality I think it would be interesting to expose multiplication so that e.g. Shamir Secret Sharing can be implemented. On the same note having `secp256k1_scalar_set_u64` available would probably also be nice (although that seems covered by `secp256k1_internals_scalar_set_b32` in a more clunky way).

Issue #155, opened 2021-12-21
Various MuSig Improvements


jonasnick:
- [ ] Fix `TODO` items: "use multiexp to compute -s*G + e*mu*pubkey + aggnonce[0] + b*aggnonce[1]" and "Cache mu".
- [ ] `musig_nonce_gen` currently rejects `session_id32=0` if there's no `seckey`. This check could be stronger and reject weak entropy counters, for example in a endianness-independent way by checking that the middle 16-bytes are not all 0 (H/T @robot-dreams).
- Strengthen nonce generation by feeding addresses of arguments to the function (#152)
- [ ] Support sign-to-contract (@dr-orlovsky)
- [ ] Reenable efficient multiexponentiation (`ecmult_multi`) in `pubkey_agg` when confidence in its implementation is higher.
- [ ] Allow signing variable length messages (i.e. with length != 32) as proposed for BIP-340
- [ ] Consider treating `pre_sig64` argument of `musig_adapt` and `musig_extract_adaptor` as secret (for constant-timeness test)

PR #157, opened 2022-01-01
Add description of MuSig signing to musig-spec.md

jonasnick:
Will wait before adding tweaking until #151 is merged.


doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+* The message ''m'': a 32-byte array
+
+The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined as:
+* Let ''R<sub>1</sub> = lift_x(int(aggnonce[0:33])), R<sub>2</sub> = lift_x(int(aggnonce[33:66]))''; fail if that fails
+* Let ''Q = KeyAgg(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce,bytes(Q),m)) mod n''
+* Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
+* Fail if ''is_infinite(R)''
+* Let ''k'<sub>1</sub> = int(secnonce[0:32]), k'<sub>2</sub> = int(secnonce[32:64])''
+* Fail if ''k'<sub>i</sub> = 0'' or ''k'<sub>i</sub> &ge; n'' for ''i = 1..2''
```
robot-dreams, line 142: I think the implementation in this repo checks for 0 but not for overflow (see https://github.com/ElementsProject/secp256k1-zkp/blob/master/src/modules/musig/session_impl.h#L31), should the spec match this?

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+* Let ''k<sub>1</sub> = k'<sub>1</sub>, k<sub>2</sub> = k'<sub>2</sub> '' if ''has_even_y(R)'', otherwise let ''k<sub>1</sub> = n - k'<sub>1</sub>, k<sub>2</sub> = n - k<sub>2</sub>''
+* Let ''d' = int(sk)''
+* Fail if ''d' = 0'' or ''d' &ge; n''
+* Let ''P = d'⋅G''
+* Let ''d = d' '' if ''has_even_y(P) `XOR` has_even_y(Q)'', otherwise let ''d = n - d' ''
+* Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
+* Let ''mu = KeyAggCoeff(pk<sub>1..u</sub>, bytes(P))''
+* Let ''s = (k<sub>1</sub> + b⋅k<sub>2</sub> + e⋅mu⋅d) mod n''
+* Let ''psig = bytes(s)''
+* If ''PartialSigVerify(psig, pk<sub>1..u</sub>, pubnonce<sub>1..u</sub>, m, bytes(P))'' (see below) returns failure, abort<ref>Verifying the signature before leaving the signer prevents random or attacker provoked computation errors. This prevents publishing invalid signatures which may leak information about the secret key. It is recommended, but can be omitted if the computation cost is prohibitive.</ref>.
```
robot-dreams, line 152: Is it worth adding a checklist item to #155 to add a flag to `secp256k1_musig_partial_sign` to allow verification before returning?

robot-dreams (commented):
Looking great so far! Aside from inline comments:

- For each of the various functions that take a list of input public keys, is there a precondition on the input public keys being already sorted, or is the function expected to sort the keys itself, or neither?

- There are a few places where you're using lift_x(int(...)) where the input is a 33-byte compressed representation, rather than a 32-byte x-only representation. Is it necessary to define a new helper to handle this case?

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+
+=== Motivation ===
+
+== Description ==
+
+=== Design ===
+
+* A function for sorting public keys allows to aggregate keys independent of the (initial) order.
+* The KeyAgg coefficient is computed by hashing the key instead of key index. Otherwise, if the pubkey list gets sorted, the signer needs to translate between key indices pre- and post-sorting.
+* The second unique key in the pubkey list gets the constant KeyAgg coefficient 1 which saves an exponentiation (see the MuSig2* appendix in the [https://eprint.iacr.org/2020/1261 MuSig2 paper]).
```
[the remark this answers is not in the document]
jonasnick: It is the second in the input list of keyagg. Added that to the sentence.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+** The function ''int(x)'', where ''x'' is a 32-byte array, returns the 256-bit unsigned integer whose most significant byte first encoding is ''x''.
+** The function ''lift_x(x)'', where ''x'' is an integer in range ''0..p-1'', returns the point ''P'' for which ''x(P) = x''<ref>
+    Given a candidate X coordinate ''x'' in the range ''0..p-1'', there exist either exactly two or exactly zero valid Y coordinates. If no valid Y coordinate exists, then ''x'' is not a valid X coordinate either, i.e., no point ''P'' exists for which ''x(P) = x''. The valid Y coordinates for a given candidate ''x'' are the square roots of ''c = x<sup>3</sup> + 7 mod p'' and they can be computed as ''y = &plusmn;c<sup>(p+1)/4</sup> mod p'' (see [https://en.wikipedia.org/wiki/Quadratic_residue#Prime_or_prime_power_modulus Quadratic residue]) if they exist, which can be checked by squaring and comparing with ''c''.</ref> and ''has_even_y(P)'', or fails if ''x'' is outside the range or no such point exists. The function ''lift_x(x)'' is equivalent to the following pseudocode:
```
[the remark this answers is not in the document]
jonasnick: Yeah, I rewrote this a bit since it sounded weird that x was defined to be in the range, yet it could be called with an argument that exceeded that.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+*** Let ''c = x<sup>3</sup> + 7 mod p''.
+*** Let ''y = c<sup>(p+1)/4</sup> mod p''.
+*** Fail if ''c &ne; y<sup>2</sup> mod p''.
+*** Return the unique point ''P'' such that ''x(P) = x'' and ''y(P) = y'' if ''y mod 2 = 0'' or ''y(P) = p-y'' otherwise.
```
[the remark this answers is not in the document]
jonasnick: Ok, I hope it's still clear that this is the same `lift_x` as in bip-schnorr.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+==== Key Aggregation ====
+
+Input:
+* The number ''u'' of public keys with ''0 < u < 2^32''
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+
+The algorithm ''KeyAgg(pk<sub>1..u</sub>)'' is defined as:
+* For ''i = 1 .. u'':
+** Let ''a<sub>i</sub> = KeyAggCoeff(pk<sub>1..u</sub>, pk<sub>i</sub>)''.
+** Let ''P<sub>i</sub> = lift_x(int(pk<sub>i</sub>))''; fail if it fails.
```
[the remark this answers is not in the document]
jonasnick: done

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+
+The algorithm ''NonceAgg(pubnonce<sub>1..u</sub>)'' is defined as:
+* For ''i = 1 .. 2'':
+** For ''j = 1 .. u'':
+*** Let ''R<sub>i,j</sub> = lift_x(int(pubnonce<sub>j</sub>[i*33:i*33+33]))''; fail if that fails
+** Let ''R'<sub>i</sub> = R<sub>i,1</sub> + R<sub>i,2</sub> + ... + R<sub>i,u</sub>''
+** Let ''R<sub>i</sub> = R'<sub>i</sub>'' if not ''is_infinite(R'<sub>i</sub>)'', otherwise let R<sub>i</sub> = G''
+* Return ''aggnonce = cbytes(R<sub>1</sub>) || cbytes(R<sub>2</sub>)''
+
+Note: If ''is_infinite(R'<sub>i</sub>)'' there is at least one dishonest signer.
```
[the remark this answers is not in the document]
jonasnick: Ok, done

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+If we would fail here, we will never be able to determine who it is.
+Therefore, we should continue such that the culprit is revealed when collecting and verifying partial signatures.
+However, dealing with the point at infinity requires defining a serialization and may require extra code complexity in implementations.
+Instead, we set the aggregate nonce to some arbitrary point, the generator.
+This is secure, because it only restricts the abilities of the attacker: an attacker that forces the sum of nonces to be infinity by sending some maliciously generated nonce pairs can be turned into an attacker that forces the sum to be the generator (by simply adding the generator to one of the malicious nonces), and this does not change the winning condition of the signature security game (EUF-CMA).
```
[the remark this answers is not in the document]
jonasnick: I don't understand. You're saying that the attacker can send `R_neg = - R_others` and `R_neg + G` and both will result in the same final nonce R. If you get two partial sigs with the same R but two different values of c, then the secnonce has been reused. The attacker can also get same R different c by choosing a different message if the secnonce is reused.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,222 @@
...
+* The secret key ''sk'': a 32-byte array
+* The aggregate public nonce ''aggnonce'': a 66-byte array
+* The number ''u'' of public keys with ''0 < u < 2^32''
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The message ''m'': a 32-byte array
+
+The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined as:
+* Let ''R<sub>1</sub> = lift_x(int(aggnonce[0:33])), R<sub>2</sub> = lift_x(int(aggnonce[33:66]))''; fail if that fails
+* Let ''Q = KeyAgg(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce,bytes(Q),m)) mod n''
```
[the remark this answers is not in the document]
jonasnick: Besides instantiating the random oracle with a hash function, how is it different from the paper?

jonasnick:
@robot-dreams

> For each of the various functions that take a list of input public keys, is there a precondition on the input public keys being already sorted, or is the function expected to sort the keys itself, or neither?

Neither. Added a note that the output of KeyAgg is dependent on the order of the inputs. So either implementors of the spec make sure that that all signers get the same order of pubkeys or they sort them according to the spec (see discussion here https://github.com/ElementsProject/secp256k1-zkp/pull/120).

>  Is it necessary to define a new helper to handle this case?

Oops, added `point` and `pointc` function.

doc/musig-spec.mediawiki
jonasnick: In our implementation we never generate a secnonce out of range and this would surely be the case in other implementations. This is just added here to be sure that we can serialize `k'*G` and that scalar multiplication is well definied.

doc/musig-spec.mediawiki
jonasnick: I don't think there's much overhead to calling `secp256k1_musig_partial_verify` after `secp256k1_musig_partial_sign`. But interesting idea because it could result in more users who don't care so much about signing performance to actually do verify.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: I was just referring to the ordering (paper has X, R, m); probably not a big deal in retrospect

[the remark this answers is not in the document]
jonasnick: Hm I don't think that's noteworthy because there's a clear difference between an abstract description of the scheme (MuSig paper) and a spec. One can imagine that `RO(X, R, m)` is instantiated as `hash(bytes(R),bytes(X),m)`. But generally, if there are other significant differences (for example setting aggnonce to G if infinite otherwise) we should definitely point them out.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: Oops, what I said before was nonsense. I forgot that the adversary also needs the private nonce, haha.

Is it correct if an adversary A can win the EUF-CMA game against the modified scheme (call it MuSig2'), then A can be used directly to also win the EUF-CMA game against MuSig2*?

I'm thinking every query that A makes against a MuSig2' challenger (H_agg, H_non, H_sig, Sign, Sign') is also a valid query against a MuSig2* challenger (I assume this is what you meant by "only restricts the abilities of the attacker"), and a valid forgery against MuSig2' is also a valid forgery against MuSig2*; thus the "reduction" is completely trivial.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: Ah, I missed the fact that this part matches BIP-0340. If you think it's better to keep the wording consistent with BIP-0340, that would also make sense to me.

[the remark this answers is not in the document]
jonasnick: If there are obvious, simple improvements I think we should make them.

doc/musig-spec.mediawiki
robot-dreams: You're right, `nonce_gen` indeed checks for overflow.

robot-dreams: Just confirming, where is the `k >= n` check in the implementation? I tried looking in `secp256k1_musig_nonce_gen` but I didn't see it:

https://github.com/ElementsProject/secp256k1-zkp/blob/master/src/modules/musig/session_impl.h#L313

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
jonasnick: Hm, I had to think about this quite a bit again which is an indication that we should expand on this argument a bit. Well, now we have the space for it.

> then A can be used directly to also win the EUF-CMA game against MuSig2*?

I don't think so. If A sends a Sign' query with aggnonce[0] = infinity, it expects a partial signature using aggnonce[0] = G. Therefore, a reduction would need to add G to A's query if it is infinity.

> a valid forgery (L, m, (R, s)) against MuSig2' is also a valid forgery against MuSig2*

What confused me a bit is that R = infinity is allowed in principle in MuSig2*. The scheme we're comparing this to is where R = infinity in a forgery is disallowed already (which is what BIP340 also implements).

doc/musig-spec.mediawiki
jonasnick: `secp256k1_scalar_set_b32` in `musig_nonce_function` computes the scalar mod `n` if it overflows.

robot-dreams: Great, that's what I missed, thanks!

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,227 @@
...
+*** Fail if ''x &gt; p-1''.
+*** Let ''c = x<sup>3</sup> + 7 mod p''.
+*** Let ''y' = c<sup>(p+1)/4</sup> mod p''.
+*** Fail if ''c &ne; y'<sup>2</sup> mod p''.
+*** Let ''y = y' '' if ''y' mod 2 = 0'', otherwise let ''y = p - y' ''.
+*** Return the unique point ''P'' such that ''x(P) = x'' and ''y(P) = y''.
+** The function ''point(x)'', where ''x'' is a 32-byte array ("x-only" serialization), returns ''lift_x(int(x))''. Fail if ''lift_x'' fails.
+** The function ''pointc(x)'', where ''x'' is a 33-byte array (compressed serialization), sets ''P = lift_x(int(x[1:33]))'' and fails if that fails. If ''x[0] = 2'' it returns ''P'' and if ''x[0] = 3'' it returns ''-P''. Otherwise, it fails.
```
brandonblack, line 61: Interesting to define EC Point Decompression in terms of `lift_x` instead of the reverse. Would it be simpler and more historically traceable to do the reverse?
```
pointc(x) = ec_decompress(x)
point(x) = ec_decompress(0x02 || x)
```

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,227 @@
...
+* Fail if ''is_infinite(Q)''.
+* Return ''bytes(Q)''.
+
+The algorithm ''HashKeys(pk<sub>1..u</sub>)'' is defined as:
+* Return ''hash<sub>KeyAgg list</sub>(pk<sub>1</sub> || pk<sub>2</sub> || ... || pk<sub>u</sub>)''
+
+The algorithm ''IsSecond(pk<sub>1..u</sub>, pk')'' is defined as:
+* For ''j = 1 .. u'':
+** If ''pk<sub>j</sub> &ne; pk<sub>1</sub>'':
+*** Return ''true'' if ''pk<sub>j</sub> = pk' '', otherwise return ''false''.
```
brandonblack, line 94:
```suggestion
** If ''pk<sub>j</sub> &ne; pk<sub>2</sub>'':
*** Return ''true''
```

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,227 @@
...
+* Return ''false''
+
+The algorithm ''KeyAggCoeff(pk<sub>1..u</sub>, pk)'' is defined as:
+* Let ''L = HashKeys(pk<sub>1..u</sub>)''.
+* Return 1 if ''IsSecond(pk<sub>1..u</sub>, pk<sub>i</sub>)'', otherwise return  ''int(hash<sub>KeyAgg coefficient</sub>(L || pk)) mod n''.
```
brandonblack, line 99:
```suggestion
* If ''IsSecond(pk<sub>1..u</sub>, pk<sub>i</sub>)''
** Return 1
* Return  ''int(hash<sub>KeyAgg coefficient</sub>(L || pk)) mod n''.
```

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: Ah, I think I see what you're saying.

I was previously considering`NonceAgg` as being done by the adversary, including the part where it possibly replaces infinity with the generator (because `NonceAgg` only depends on public data). In this case, I think if Sign' for MuSig2' receives infinity as one of the aggnonces, it should just fail.

But if replacing infinity with the generator is done inside Sign', by the MuSig2' challenger, then I agree with you that the reduction needs to replace infinity with the generator before passing the query onto the MuSig2* challenger.

brandonblack (commented):
Great to see this in detail and with clear definitions for inputs!

doc/musig-spec.mediawiki
jonasnick: I considered this but didn't want to define a new function ec_decompress that's largely redundant with BIP340's lift-x.

doc/musig-spec.mediawiki
jonasnick: Hm, I don't see how that would work.

brandonblack: You're right, I misunderstood the original code.

doc/musig-spec.mediawiki
jonasnick: thx, that's clearer.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: Here's a suggestion for an alternate way to argue about the security of the modification. I'm not 100% confident in it but if this general idea works I think it might be easier to understand:

> Note that NonceAgg (both the original and modified version) only depends on publicly available data (the set of public pre-nonces from every signer). Thus in the context of the signature security game (EUF-CMA), we can consider NonceAgg to be performed by the adversary (rather than the challenger) without loss of generality.
>
> This modification changes neither the behavior of the EUF-CMA challenger nor the condition required to win the security game (the adversary still has to output a valid forgery according to the unmodified MuSig2* scheme). Since we've already proved that MuSig2* is secure against an arbitrary adversary, we can conclude that the modified scheme is still secure.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,229 @@
...
+* Let ''Q = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(Q)''.
+* Return ''bytes(Q)''.
+
+The algorithm ''HashKeys(pk<sub>1..u</sub>)'' is defined as:
+* Return ''hash<sub>KeyAgg list</sub>(pk<sub>1</sub> || pk<sub>2</sub> || ... || pk<sub>u</sub>)''
+
+The algorithm ''IsSecond(pk<sub>1..u</sub>, pk')'' is defined as:
+* For ''j = 1 .. u'':
+** If ''pk<sub>j</sub> &ne; pk<sub>1</sub>'':
```
brandonblack, line 93:
```suggestion
** If ''pk<sub>j</sub> &eq; pk<sub>2</sub>'':
```

jonasnick: That's not equivalent if `pk1 = pk2`. We really want the "second pubkey" in the sense that it's unequal from the first.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
jonasnick: > But if replacing infinity with the generator is done inside Sign', by the MuSig2' challenger, then I agree with you

You're right, what I said is nonsense. Aggregation happens outside of Sign' exactly because we don't want to trust aggregation.

The sentence "it only restricts the abilities of the attacker" is wrong, because the abilities of the attacker are exactly the same (arbitrary `NonceAgg`). I pushed your suggestion (after minor modifications).

CC @real-or-random

doc/musig-spec.mediawiki
jonasnick: Added to #155

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* The number ''u'' of public keys with ''0 < u < 2^32''
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+
+The algorithm ''KeyAgg(pk<sub>1..u</sub>)'' is defined as:
+* For ''i = 1 .. u'':
+** Let ''a<sub>i</sub> = KeyAggCoeff(pk<sub>1..u</sub>, pk<sub>i</sub>)''.
+** Let ''P<sub>i</sub> = point(pk<sub>i</sub>)''; fail if that fails.
+* Let ''Q = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(Q)''.
+* Return ''bytes(Q)''.
```
robot-dreams, line 86: Should KeyAgg return Q instead of bytes(Q), based on how it's used in Sign and PartialSigVerify?

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce,bytes(Q),m)) mod n''
+* Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
+* Fail if ''is_infinite(R)''
+* Let ''k'<sub>1</sub> = int(secnonce[0:32]), k'<sub>2</sub> = int(secnonce[32:64])''
+* Fail if ''k'<sub>i</sub> = 0'' or ''k'<sub>i</sub> &ge; n'' for ''i = 1..2''
+* Let ''k<sub>1</sub> = k'<sub>1</sub>, k<sub>2</sub> = k'<sub>2</sub> '' if ''has_even_y(R)'', otherwise let ''k<sub>1</sub> = n - k'<sub>1</sub>, k<sub>2</sub> = n - k<sub>2</sub>''
+* Let ''d' = int(sk)''
+* Fail if ''d' = 0'' or ''d' &ge; n''
+* Let ''P = d'⋅G''
+* Let ''d = d' '' if ''has_even_y(P) `XOR` has_even_y(Q)'', otherwise let ''d = n - d' ''
```
robot-dreams, line 160: Should this be reversed? Negate if `has_even_y(P) XOR has_even_y(Q)`?

robot-dreams (commented):
I did a second review pass by just trying to [implement the spec directly](https://gist.github.com/robot-dreams/a51b1e73df1f604fe77a5cb2b9cdb078) and seeing if any issues came up.

Aside from the inline comments below, the spec was enough to produce the expected test vectors for KeyAgg and Sign :)

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+
+==== Signing ====
+
+Input:
+* The secret nonce ''secnonce'' that has never been used as input to ''Sign'' before: a 64-byte array
+* The secret key ''sk'': a 32-byte array
+* The aggregate public nonce ''aggnonce'': a 66-byte array
+* The number ''u'' of public keys with ''0 < u < 2^32''
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The message ''m'': a 32-byte array
```
roconnor-blockstream, line 146: To be used with taproot, signing also needs a tweak parameter.  I don't know if this needs to be passed into the nonce stuff.

jonasnick: Yes, see the PR description:

> Will wait before adding tweaking until #151 is merged.

The implementation already has x-only tweaking and #151 adds "ordinary" tweaking.

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+
+==== Nonce Aggregation ====
+
+* The number ''u'' of ''pubnonces'' with ''0 < u < 2^32''
+* The public nonces ''pubnonce<sub>1..u</sub>'': ''u'' 66-byte arrays
+
+The algorithm ''NonceAgg(pubnonce<sub>1..u</sub>)'' is defined as:
+* For ''i = 1 .. 2'':
+** For ''j = 1 .. u'':
+*** Let ''R<sub>i,j</sub> = pointc(pubnonce<sub>j</sub>[i*33:i*33+33])''; fail if that fails
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* The secret key ''sk'': a 32-byte array
+* The aggregate public nonce ''aggnonce'': a 66-byte array
+* The number ''u'' of public keys with ''0 < u < 2^32''
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The message ''m'': a 32-byte array
+
+The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined as:
+* Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
+* Let ''Q = KeyAgg(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce,bytes(Q),m)) mod n''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* Let ''k<sub>1</sub> = k'<sub>1</sub>, k<sub>2</sub> = k'<sub>2</sub> '' if ''has_even_y(R)'', otherwise let ''k<sub>1</sub> = n - k'<sub>1</sub>, k<sub>2</sub> = n - k<sub>2</sub>''
+* Let ''d' = int(sk)''
+* Fail if ''d' = 0'' or ''d' &ge; n''
+* Let ''P = d'⋅G''
+* Let ''d = d' '' if ''has_even_y(P) `XOR` has_even_y(Q)'', otherwise let ''d = n - d' ''
+* Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
+* Let ''mu = KeyAggCoeff(pk<sub>1..u</sub>, bytes(P))''
+* Let ''s = (k<sub>1</sub> + b⋅k<sub>2</sub> + e⋅mu⋅d) mod n''
+* Let ''psig = bytes(s)''
+* If ''PartialSigVerify(psig, pk<sub>1..u</sub>, pubnonce<sub>1..u</sub>, m, bytes(P))'' (see below) returns failure, abort<ref>Verifying the signature before leaving the signer prevents random or attacker provoked computation errors. This prevents publishing invalid signatures which may leak information about the secret key. It is recommended, but can be omitted if the computation cost is prohibitive.</ref>.
```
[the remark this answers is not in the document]
jonasnick: I added a `PartialSigVerifyInternal` function that takes the aggregate nonce.
I didn't want to replace the pubnonces argument of PartialSigVerify with aggnonce because that would wrongly indicate that verifying makes sense with untrusted nonce aggregation

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The public key of the signer ''pk<sup>*</sup>'' (in ''pk<sub>1..u</sub>''): a 32-byte array
+* The message ''m'': a 32-byte array
+
+The algorithm ''PartialSigVerify(psig, pubnonce<sub>1..u</sub>, pk<sub>1..u</sub>, pk<sup>*</sup>, m)'' is defined as:
+* Let ''s = int(psig)''; fail if ''s &ge; n''
+* Let ''aggnonce = NonceAgg(pubnonce<sub>1..u</sub>)''; fail if that fails
+* Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
+* Let ''Q = KeyAgg(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce,bytes(Q),m)) mod n''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,235 @@
...
+* Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
+* Let ''R<sup>*</sup><sub>1</sub> = pointc(pubnonce<sub>i</sub>[0:33]), R<sup>*</sup><sub>2</sub> = pointc(pubnonce<sub>i</sub>[33:66])''
```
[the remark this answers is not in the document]
jonasnick: Added separate pubnonce argument.

doc/musig-spec.mediawiki
jonasnick: KeyAgg should return bytes(Q) because a public key is by definition a 32 byte array. I instead converted the KeyAgg output to a point at the relevant places.

doc/musig-spec.mediawiki
jonasnick: Negation fail. Fixed.

jonasnick:
@robot-dreams

> Aside from the inline comments below, the spec was enough to produce the expected test vectors for KeyAgg and Sign :)

Excellent, that's great to hear. Thank you for the review!

doc/musig-spec.mediawiki
```diff
@@ -147,7 +147,7 @@ Input:
 
 The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined as:
 * Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
-* Let ''Q = KeyAgg(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''Q = point(KeyAgg(pk<sub>1..u</sub>))''; fail if that fails
```
robot-dreams, line 150: `point(bytes(Q)) drops the parity of Q, but do you need the parity here to know if you need to negate the signing key?

jonasnick:
Rebased

doc/musig-spec.mediawiki
```diff
@@ -172,17 +173,32 @@ Input:
 * The number ''u'' of public nonces and public keys with ''0 < u < 2^32''
 * The public nonces ''pubnonce<sub>1..u</sub>'': ''u'' 66-byte arrays
 * The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The message ''m'': a 32-byte array
+* The index of the signer ''i'' in the public nonces and public keys with ''0 < i <= u''
+
+The algorithm ''PartialSigVerify(psig, pubnonce<sub>1..u</sub>, pk<sub>1..u</sub>, m, i)'' is defined as:
+* Let ''aggnonce = NonceAgg(pubnonce<sub>1..u</sub>)''; fail if that fails
+* Ren ''PartialSigVerifyInternal(psig, pubnonce<sub>i</sub>, aggnonce, pk<sub>1..u</sub>, pk<sub>i</sub>, m)''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
```diff
@@ -162,7 +162,8 @@ The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined
 * Let ''mu = KeyAggCoeff(pk<sub>1..u</sub>, bytes(P))''
 * Let ''s = (k<sub>1</sub> + b⋅k<sub>2</sub> + e⋅mu⋅d) mod n''
 * Let ''psig = bytes(s)''
-* If ''PartialSigVerify(psig, pk<sub>1..u</sub>, pubnonce<sub>1..u</sub>, m, bytes(P))'' (see below) returns failure, abort<ref>Verifying the signature before leaving the signer prevents random or attacker provoked computation errors. This prevents publishing invalid signatures which may leak information about the secret key. It is recommended, but can be omitted if the computation cost is prohibitive.</ref>.
+* Let ''pubnonce = cbytes(k'<sub>1</sub>⋅G) || cbytes(k'<sub>1</sub>⋅G)''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
jonasnick: Of course. Added `KeyAggInternal`.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
robot-dreams: OK, sounds good for now, but I do want to think about this more if you add adaptors to the spec (since the implementation has comments that it's much more important to call `PartialSigVerify` in that case)

robot-dreams:
ACK 623f3cb5d0da9edc61386be73c200522ed1d399c 🚀

jonasnick:
Squashed fixups

doc/musig-spec.mediawiki
```diff
@@ -0,0 +1,255 @@
...
+* Let ''k<sub>1</sub> = k'<sub>1</sub>, k<sub>2</sub> = k'<sub>2</sub> '' if ''has_even_y(R)'', otherwise let ''k<sub>1</sub> = n - k'<sub>1</sub>, k<sub>2</sub> = n - k<sub>2</sub>''
+* Let ''d' = int(sk)''
+* Fail if ''d' = 0'' or ''d' &ge; n''
+* Let ''P = d'⋅G''
+* Let ''d = n - d' '' if ''has_even_y(P) `XOR` has_even_y(Q)'', otherwise let ''d = d' ''
+* Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
+* Let ''mu = KeyAggCoeff(pk<sub>1..u</sub>, bytes(P))''
+* Let ''s = (k<sub>1</sub> + b⋅k<sub>2</sub> + e⋅mu⋅d) mod n''
+* Let ''psig = bytes(s)''
+* Let ''pubnonce = cbytes(k'<sub>1</sub>⋅G) || cbytes(k'<sub>2</sub>⋅G)''
```
jonasnick, line 169: I'm tending towards changing the serialization of the nonce pair from `compressed(R1) || compressed(R2)` (66 bytes) to `parity(R1) & parity(R2)  << 1 || xonly(R1) || xonly(R2)` (65 bytes).

brandonblack: `parity(R1) | parity(R2) << 1`? (`|` instead of `&`)

Although I'm having a hard time thinking of a case where this 1 byte matters, given that the partial nonces never appear in the blockchain.

jonasnick: Eh sure, edited. OTOH there's no reason to add a useless byte.

jonasnick:
Squashed fixup & rebased

robot-dreams:
ACK 69b392f3cbd4dbff953ec8f2ff44f6a8f612b661 based on:

- https://github.com/ElementsProject/secp256k1-zkp/pull/157#issuecomment-1006114758
- 623f3cb5d0da9edc61386be73c200522ed1d399c  -> 703f74144e91e9ed9b783a57a61059036dc003dd was a squash
- Inspected output of "git range-diff a5b5909e8d56eb3c23aeb6eb4cdca0a621a0930b 703f74144e91e9ed9b783a57a61059036dc003dd 69b392f3cbd4dbff953ec8f2ff44f6a8f612b661"

real-or-random (approved):
ACK https://github.com/ElementsProject/secp256k1-zkp/pull/157/commits/69b392f3cbd4dbff953ec8f2ff44f6a8f612b661 I haven't looked at every detail but it's certainly ready to be merged as draft spec

I think we should move this to a separate repo (fork of bips repo) soon.

doc/musig-spec.mediawiki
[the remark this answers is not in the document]
real-or-random: I was also confused first but I think now the argument makes sense. Wording can still be improved. Right now it sounds like we're coming up with this argument here in the spec but the EUF-CMA game in the paper already lets the adversary perform NonceAgg (=SignAgg in the paper).

So I believe a formal argument is as simple as this. 

Let MuSig2* be the scheme from the paper, let MuSig2† be MuSig2* with the modification that the scheme simply aborts if one of the aggregate nonces is infinity, and let MuSig2†† be the scheme in the spec. (Note that MuSig2†† also aborts if the one of the aggregate nonces is infinity because then parsing fails.)

MuSig2* is EUF-CMA by the theorem in the paper. Since Sign† only adds aborts to Sign*, and the adversary can simulate internally whether Sign† will hit those aborts (by checking whether one of the aggregate nonces is infinity), MuSig2† is EUF-CMA. Since MuSig2† and MuSig2†† differ only in NonceAgg, the EUF-CMA games of MuSig2† and MuSig2†† are identical. Thus MuSig2†† is EUF-CMA.

(This could also be written nicely in game hopping style.)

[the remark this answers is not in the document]
robot-dreams: OK, I think I'm following—my problem with my informal argument above is that it doesn't account for the extra "abort" possibility (it used without proof the fact that MuSig2† is secure). Here's my understanding of your formal argument, is this correct?

- Suppose we have adversary A†† that wins MuSig2†† game[1]; here's how we build A†:
	- Use A†† directly, since MuSig2† and MuSig2†† are the same game

- Suppose we have adversary A† that wins MuSig2† game; here's how we build A*:
	- Start with transparent wrapper around A†
	- Whenever A† makes call to Sign†, first check public nonces
		- If either of them is infinity, return "abort" to A†
		- Otherwise, forward call to Sign* and return result to A†
	- Wrapper simulates Sign† perfectly to A†, and everything else is unchanged
	- The forgery that A† outputs wins MuSig2† (of course) but it also wins MuSig2*

Thus if we have adversary A†† that wins MuSig2†† game, then we could build A* that wins MuSig2* game. From the paper, we can't have adversary A* that wins MuSig2* game. So taking the contrapositive, we can't have adversary A†† that wins MuSig2†† game. So the scheme in the spec is secure.

[1] I mean "wins ... game *with non-negligible probability*" everywhere, but I'll leave that out to keep it shorter.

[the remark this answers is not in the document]
real-or-random: Yes, exactly correct! So your wrapper around A† is the formal implementation of my sentence "the adversary can simulate internally whether Sign† will hit those aborts (by checking whether one of the aggregate nonces is infinity)". 

I think the "abort" thing is a very minor detail here -- but I believe it's necessary in a fully rigorous proof: by observing aborts, the adversary could learn new information (say Sign† aborts iff the last bit of the secret key is 1).

But since the aborts in fact depend only on publicly available information, this is not the case here. And your reduction formalizes this intuition: whatever can be learned from a Sign† oracle (which can abort) can also be learned from a Sign* oracle (which does not abort), simply by determining locally whether Sign† would abort.

This also means that the current text is slightly wrong, when it says "This modification changes neither the behavior of the EUF-CMA challenger". The modification *does* change the behavior of the EUF-CMA challenger, namely by introducing aborts -- but it does so only in a way that anyhow can be simulated by the adversary.

PR #160, opened 2022-01-04
rangeproof: clean up legacy rangeproof code

apoelstra:
This is the first round of cleanups with the goal of reducing the stack space usage of the rangeproof proving and verification code. This is needed for many embedded applications, and even on hosted machines it's a bit obscene that we're throwing tens of kilobytes of stuff onto the stack unnecessarily.

There is some stack reduction in this PR itself (dropping the 4k `prep` array from `rangeproof_genrand_sign` and the 1k `k` array from `rangeproof_sign_impl` but that isn't the main goal of this PR. The goal is to get the code into a state where it's easier to read and reason about.

This PR does **not** include the addition of closures or other complex APIs that might be controversial. It is a collection of cherry-picks from my work in November on stack space reduction, where I excluded everything I thought was "experimental" or unlikely to work out.


src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -267,7 +267,81 @@ SECP256K1_INLINE static void secp256k1_rangeproof_init_rng(
...
+            secp256k1_rfc6979_hmac_sha256_generate(rng, tmp, 32);
+            if (message && i < header->n_rings - 1) {
+                /* encode the message side-channel */
+                int b;
+                for (b = 0; b < 32; b++) {
+                    if ((i * 4 + j) * 32 + b < msg_len) {
+                        tmp[b] ^= message[(i * 4 + j) * 32 + b];
+                    }
+                }
+            } else if (i == header->n_rings - 1) {
```
apoelstra, line 316: There is a change in signing behavior here, which is that we no longer encode data in the (potentially two) `s`-values in the final ring.

The tests don't catch this, and I hate that this behavior exists, but I think I should restore it so that there is no potential change in behavior. I will come up with a fixed test vector for this.

apoelstra: Actually, we are okay. In e7a8a5f6383c705159f191b6f4c165efdcfb4334 (July 2016) I made it illegal (fails `rangeproof_sign`) to produce signatures that used these last two `s` values for sidechannel purposes, even though it looks like gmax did actually implement the logic for it in 2015.

So it has never been possible, using the libsecp-zkp API, to produce a rangeproof that embedded data in this space, so I'm comfortable removing the ability to as part of these cleanups.

apoelstra: Made this explicit in the new commit 0db19e5170e6096744dac5fa6de85a185ed70bd1

real-or-random:
Is this ready for review or are you working on further changes?

apoelstra:
@real-or-random ready for review. I am building stuff on top of this, which seem to be working now (I eliminated all the massive arrays in `rangeproof_rewind_inner` :)) so I think this is the right foundation.

apoelstra:
My current tentative roadmap:

1. This "cleanup"/testing PR
2. A followup which moves some rewinding logic into `borromean_verify`, decreases the speed of rewinding a fair bit (but speeds up failed rewinding by letting us bail out before doing any EC ops), but eliminates all the huge stack objects for rewinding
3. A followup which does the same for verification (all that is left is the `pubs` array)
4. A followup(s) which makes similar changes for rangeproof signing.
5. (Maybe) an overall followup which tries to do microoptimizations and other stuff, once everything else has settled out.

src/modules/rangeproof/main_impl.h
```diff
@@ -231,17 +231,23 @@ int secp256k1_pedersen_blind_generator_blind_sum(const secp256k1_context* ctx, c
...
  uint64_t *min_value, uint64_t *max_value, const unsigned char *proof, size_t plen) {
+    secp256k1_rangeproof_header header;
     size_t offset;
-    uint64_t scale;
     ARG_CHECK(exp != NULL);
     ARG_CHECK(mantissa != NULL);
     ARG_CHECK(min_value != NULL);
     ARG_CHECK(max_value != NULL);
     ARG_CHECK(proof != NULL);
     offset = 0;
```
jonasnick, line 241: `secp256k1_rangeproof_getheader_impl` took offset as an input, but `header_parse` doesn't. Remove it?

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -59,188 +249,196 @@ SECP256K1_INLINE static void secp256k1_rangeproof_serialize_point(unsigned char*
...
     memcpy(rngseed, nonce, 32);
     secp256k1_rangeproof_serialize_point(rngseed + 32, commit);
     secp256k1_rangeproof_serialize_point(rngseed + 32 + 33, genp);
     memcpy(rngseed + 33 + 33 + 32, proof, len);
-    secp256k1_rfc6979_hmac_sha256_initialize(&rng, rngseed, 32 + 33 + 33 + len);
-    secp256k1_scalar_clear(&acc);
+    secp256k1_rfc6979_hmac_sha256_initialize(rng, rngseed, 32 + 33 + 33 + len);
+}
+
+SECP256K1_INLINE static int secp256k1_rangeproof_genrand_sign(
```
jonasnick, line 270: Is there anything preventing using `genrand_sign` on an unitialized RNG? If not, can we VERIFY_CHECK this?

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -59,188 +249,196 @@ SECP256K1_INLINE static void secp256k1_rangeproof_serialize_point(unsigned char*
...
-    size_t rsizes[32];                /* How many possible values there are for each place. */
     size_t secidx[32];                /* Which digit is the correct one. */
+    secp256k1_rfc6979_hmac_sha256 genrand_rng;
     size_t len;                       /* Number of bytes used so far. */
     size_t i;
-    int overflow;
-    size_t npub;
+    size_t pub_idx;
     len = 0;
-    if (*plen < 65 || min_value > value || min_bits > 64 || min_bits < 0 || exp < -1 || exp > 18) {
```
jonasnick, line 216: Shouldn't we still check that value `value > min_value`? (and if we do, we should add a test).

src/modules/rangeproof/tests_impl.h
```diff
@@ -653,25 +746,175 @@ void test_rangeproof_fixed_vectors(void) {
...
+    ));
+
+    CHECK(memcmp(blind, blind_2, 32) == 0);
+    CHECK(value == 11);
+    CHECK(min_value == 0);
+    CHECK(max_value == 15);
+    CHECK(m_len == 128); /* length of the sidechannel in the proof */
+    CHECK(memcmp(message, message_2, sizeof(message_2)) == 0);
+    for (i = sizeof(message_2); i < m_len; i++) {
+        /* No message encoded in this vector */
```
jonasnick, line 861: This seems to be wrong. There is a message in this vector.

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -421,21 +585,11 @@ SECP256K1_INLINE static int secp256k1_rangeproof_rewind_inner(secp256k1_scalar *
...
-        }
-        return 0;
-    }
-    skip1 += (rings - 1) << 2;
-    skip2 += (rings - 1) << 2;
     /* Like in the rsize[] == 1 case, Having figured out which s is the one which was not forged, we can recover the blinding factor. */
-    secp256k1_rangeproof_recover_x(&stmp, &s_orig[skip2], &ev[skip2], &s[skip2]);
-    secp256k1_scalar_negate(&sec[rings - 1], &sec[rings - 1]);
-    secp256k1_scalar_add(blind, &stmp, &sec[rings - 1]);
+    final_x_pos = 4 * (header->n_rings - 1) + ((*v >> (2 * (header->n_rings - 1))) & 3);
```
jonasnick, line 589: AFAICs `*v` isn't the same as `value`. If `!if ((tmp[0] & 128) && (memcmp(&tmp[16], &tmp[24], 8) == 0) && (memcmp(&tmp[8], &tmp[16], 8) == 0))` then `*v = 0` and `value` is `UINT_MAX`. And in the other places where we access `v` we check that it's not NULL.

src/modules/rangeproof/rangeproof.h
```diff
@@ -12,6 +12,79 @@
 #include "ecmult.h"
 #include "ecmult_gen.h"
 
+/** Structure representing data directly encoded into a rangeproof header
+ *
+ * A rangeproof is a proof, associated with a Pedersen commitment, that a
+ * "proven value" in is the range [0, 2^mantissa]. The committed value is
```
jonasnick, line 18: Shouldn't it be `range [0, 2^mantissa-1]`? If so, then same for `max_value` doc below.

jonasnick (commented):
First (shallow) review pass. Nice to see all those cleanups!

src/modules/rangeproof/rangeproof_impl.h
apoelstra: We should remove any checks that `v` is non-NULL. We have always been inconsistent about this will always have segfaulted if `v` were every NULL (but fortunately it never is).

In later commits (maybe not in this PR) I move away from the name `v` and toward `proven_value` vs `value`. You are correct that `*v` is not the same as `value`. I'm not sure what point you are making about it being 0 sometimes.

src/modules/rangeproof/main_impl.h
apoelstra: `header_parse` does take it as an input.

apoelstra: `header_parse` does take it as an input.

apoelstra: `header_parse` does take it as an input.

src/modules/rangeproof/rangeproof_impl.h
apoelstra: There is no method (currently) which can tell whether a RNG has been initialized. But the bigger risk in this code is that you may pass a "used" RNG which is initialized, but not to the right state. I don't know how we could VERIFY_CHECK something like that.

src/modules/rangeproof/rangeproof_impl.h
apoelstra: We do, at the start of `secp256k1_rangeproof_header_set_for_value`. But yes, we should add a test.

src/modules/rangeproof/main_impl.h
jonasnick: Sorry that was  ambiguous. `secp256k1_rangeproof_getheader_impl` reads `*offset` (but is always called with `*offset = 0` and `header_parse` doesn't.

apoelstra: Ah I understand! `header_parse` treats this field as write-only, so there is no longer any need to zero- initialize it. Ok, I will delete this line.

src/modules/rangeproof/tests_impl.h
```diff
@@ -653,25 +746,175 @@ void test_rangeproof_fixed_vectors(void) {
...
+        CHECK(message[i] == 0);
+    }
+}
+
+    /* Vector 3: single-value proof of UINT64_MAX */
```
jonasnick, line 866: "single-value proof" means `exp = -1`, correct? Do we have a test with `exp > 0`?

jonasnick (commented):
I haven't fully investigated this but 

```
./tests 64 106275f9d56e923d4698c32cae3b1634
```

fails with
```
src/modules/rangeproof/tests_impl.h:517: test condition failed: secp256k1_rangeproof_sign(ctx, proof, &len, vmin, &commit, blind, commit.data, exp, min_bits, v, NULL, 0, NULL, 0, secp256k1_generator_h)
```

It seems like `secp256k1_rangeproof_header_set_for_value` fails because `secp256k1_rangeproof_header_expand` fails because `header->max_value > UINT64_MAX - header->min_value`.

jonasnick (commented):
Added three more fixed test vectors to my branch (https://github.com/jonasnick/secp256k1-zkp/commits/rangeproof-cleanups-jn). Feel free to cherry-pick (and modify).

apoelstra:
Thanks! Pulled it into my branch.

jonasnick (commented):
I wonder about the implications of removing the 64-byte extra space for Elements.

Elements has this reject condition `msg_size != SIDECHANNEL_MSG_SIZE` where `msg_size` is the `m_len` argument of `rewind`,  `msg_size = 64` and `SIDECHANNEL_MSG_SIZE = 64`. If one creates a rangeproof with a single ring of size 4, non-updated nodes will get `msg_size = 64` and won't reject the transaction. Updated nodes however will have `msg_size = 0` and therefore reject the tx.

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -59,188 +257,196 @@ SECP256K1_INLINE static void secp256k1_rangeproof_serialize_point(unsigned char*
...
+            secp256k1_rfc6979_hmac_sha256_generate(rng, tmp, 32);
             do {
-                secp256k1_rfc6979_hmac_sha256_generate(&rng, tmp, 32);
+                secp256k1_rfc6979_hmac_sha256_generate(rng, tmp, 32);
                 secp256k1_scalar_set_b32(&sec[i], tmp, &overflow);
             } while (overflow || secp256k1_scalar_is_zero(&sec[i]));
             secp256k1_scalar_add(&acc, &acc, &sec[i]);
         } else {
             secp256k1_scalar_negate(&acc, &acc);
             sec[i] = acc;
```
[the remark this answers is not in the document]
apoelstra: I am actually thinking about removing all of the checks for conditions like this from the signing code, since they are cryptographically impossible to hit.

apoelstra:
@jonasnick regarding Elements, that check is actually wrong right now because the returned `msg_size` from `rangeproof_rewind` is 64 bytes larger than the actual capacity available from `rangeproof_sign`.

It is true that currently wallets will accept 2-bit rangeproofs that have embedded messages, and this change will cause those to be rejected, but *producing* such rangeproofs would require a custom rangeproof signing implementation. I'm not aware of any such implementations, nor am I aware of 2-bit rangeproofs being used in practice.

There is a sorta-attack here where somebody could *make* such a rangeproof, have it be accepted by an Elements 22 wallet, then the user updates to Elements 23 and loses her backup and needs to recover the data from the rangeproof again, and now can't. But this is a really long chain of implausible scenarios, which results in a situation which is recoverable (with some extra crypto work from us) if anybody actually encounters it.

src/modules/rangeproof/rangeproof_impl.h
jonasnick: I was wrong, `*v` is the same as `value` at this point.

jonasnick (commented):
Needs rebase but LGTM. `header_set_for_value` makes sense to me but I can't fully verify that it's exactly equivalent to the `proveparams` function it replaces. Luckily the tests are quite comprehensive.

I've added another test to my branch to exercise the following condition in `proveparams` which wasn't hit by the existing tests:
```
        if ((*min_value && value > INT64_MAX) || (value && *min_value >= INT64_MAX))
            return 0;
```
See commit `rangeproof: add test for large value and minimum value` and `fixup! rangeproof: use rangeproof_header structure for proving` [here](https://github.com/jonasnick/secp256k1-zkp/commits/rangeproof-cleanups-jn).

apoelstra:
Rebased and pulled both your commits into `rangeproof: use rangeproof_header structure for proving`

jonasnick (commented):
ACK ￼d1a3479fe61330a2d262738e8359ea5839ce5d89

real-or-random (commented):
only reviewed 3/10 commits so far

real-or-random (commented):
5/10...

real-or-random (commented):
7/10

src/modules/whitelist/main_impl.h
```diff
@@ -47,19 +47,7 @@ int secp256k1_whitelist_sign(const secp256k1_context* ctx, secp256k1_whitelist_s
...
-            if (!ret) {
-                break;
-            }
-            secp256k1_scalar_set_b32(&non, nonce32, &overflow);
-            memset(nonce32, 0, 32);
-            if (overflow || secp256k1_scalar_is_zero(&non)) {
-                count++;
-                continue;
-            }
-            done = 1;
```
real-or-random, line 62: nit: I believe this change is fine. But in general, I'm not worth making similar changes if the only short term goal is to reduce stack usage and and if we're going to replace this anyway.

real-or-random:
10/10

apoelstra:
(Pushed rebase onto current master, resulting in f25f555fd938ef8cecc66c5c0d3182c480d3a1b3. Will followup with fixes for Tim's comments.)

src/modules/rangeproof/tests_impl.h
```diff
@@ -547,8 +547,95 @@ static void test_rangeproof(void) {
...
 }
 
+static void test_single_value_proof(uint64_t val) {
+    unsigned char proof[5000];
+    secp256k1_pedersen_commitment commit;
+    unsigned char blind[32];
+    unsigned char blind_out[32];
+    unsigned char nonce[32];
+    const unsigned char message[1] = " "; /* no message will fit into a single-value proof */
+    unsigned char message_out[sizeof(proof)] = { 0 };
```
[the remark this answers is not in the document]
apoelstra: I'm going to leave this be -- I did consider using a nonzero value, but decided against it, because
* 0 is very easy to notice in debug output (though I suppose so would any constant value)
* the failure mode in message decoding will be to output random-looking data (since rewinding extracts from a RNG and then xors the output). So it's hard to see how rewinding could fail in a way that would be confused with constant 0s

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -109,6 +109,105 @@ static int secp256k1_rangeproof_header_parse(
...
+    /* Sanity checks */
+    if (min_value > value || min_bits > 64 || exp < -1 || exp > 18) {
+        return 0;
+    }
+
+    /* Start by just using the user's requested values, then adjust them in
+     * various ways to make them compatible. This is probably not advisable
+     * from a privacy point-of-view but it's important to be compatible with
+     * the 2015-era API, and all of these issues will go away when we merge
+     * Bulletproofs. */
```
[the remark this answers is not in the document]
apoelstra: I think this one is fine :) BPs simply have none of these insane options so there is no room for complex adjustment logic.

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -202,88 +301,12 @@ SECP256K1_INLINE static int secp256k1_rangeproof_genrand(secp256k1_scalar *sec,
...
-    if (*min_value == UINT64_MAX) {
-        /* If the minimum value is the maximal representable value, then we cannot code a range. */
-        *exp = -1;
-    }
-    if (*exp >= 0) {
-        int max_bits;
-        uint64_t v2;
-        if ((*min_value && value > INT64_MAX) || (value && *min_value >= INT64_MAX)) {
-            /* If either value or min_value is >= 2^63-1 then the other must by zero to avoid overflowing the proven range. */
-            return 0;
```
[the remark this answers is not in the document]
apoelstra: I think the old check was just wrong, in multiple ways.
* First, if `min_value` is greater than `INT64_MAX` or indeed greater than 0, then `value` must also be greater than 0, since it must be >= `min_value`. In the old code this was checked before even calling `proveparams`.
* If this instead intended to check that `min_value == value` in the case that either one was > `INT64_MAX`, this still makes no sense. There is no reason I shouldn't be able to have `min_value` be `INT64_MAX + 1` and `value` be `INT64_MAX + 2`, say.

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -109,6 +109,105 @@ static int secp256k1_rangeproof_header_parse(
...
+        }
+        header->exp = i;
+    }
+
+
+    /* Increase the mantissa from min_bits until it actually covers the proven value */
+    if (!secp256k1_rangeproof_header_expand(header)) {
+        return 0;
+    }
+    *proven_value = (value - header->min_value) / header->scale;
```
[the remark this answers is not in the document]
apoelstra: Dividing by max is definitely wrong, and I don't think the old code does anything like that.

src/modules/whitelist/main_impl.h
apoelstra: This removed logic is incompatible with moving the nonce-computation logic into `borromean_sign`, which is (a) necessary to do any useful amount of stack space reduction; (b) reduces the reviewing burden by no longer requiring that array indices of secret data in far-away files be compatible.

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -294,44 +317,43 @@ SECP256K1_INLINE static int secp256k1_rangeproof_sign_impl(const secp256k1_ecmul
...
         }
-        len += 8;
+    } else {
+        secidx[0] = 0;
     }
+
+    VERIFY_CHECK(v * header.scale + header.min_value == value);
+    VERIFY_CHECK(header.n_rings > 0);
+    VERIFY_CHECK(header.n_rings <= 32);
+    VERIFY_CHECK(header.n_pubs <= 128);
```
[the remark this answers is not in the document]
apoelstra: Sure. In fact, it looks like they were already there, and these `VERIFY_CHECK`s are redundant.

src/modules/whitelist/main_impl.h
apoelstra: I'm also unsure what you mean by "we're going to replace this anyway"

src/modules/rangeproof/rangeproof_impl.h
```diff
@@ -149,9 +149,17 @@ static int secp256k1_rangeproof_header_set_for_value(
          */
         header->exp = 0;
     }
+    /* Reduce min_bits to keep within a uint64_t's range (essentially copied from 2015 code) */
+    {
+        const unsigned int max_bits = min_value ? secp256k1_clz64_var(min_value) : 64;
+        if (min_bits > max_bits) {
+            header->mantissa = max_bits;
+            min_bits = max_bits;
```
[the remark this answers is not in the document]
apoelstra: Made this change. Hopefully the resulting range-diff isn't too messy (this was a pretty short patch so changing only a few lines pretty-much rewrote it).

apoelstra:
Addressed all comments in 5e36b324a98f7ea276979c3e846d4284da20f99a (which I just pushed). I left a :+1: emoji on the ones that I accepted, and replied to all the ones that I didn't.

apoelstra:
Is this good to merge actually? (Other than needing a rebase?)

PR #162, opened 2022-01-06
whitelist: remove ability to specific nonce function

apoelstra:
This functionality is inappropriate to expose for a zero-knowledge proof, and was confusingly (and potentially dangerously) implemented.

For now just replace it with code that uses the default ECDSA 6979 nonce, which we at least know will do something safe with the input that this code gives it (a "message" which is a commitment to all the pubkeys involved, xor'd with the index of which `s` value we are generating, alongside the secret key and a retry count in case of overflow). In a later PR I will just replace this mess with a sanely-seeded chacha.

(Ideally in future we would be using [Merlin](https://medium.com/@hdevalence/merlin-flexible-composable-transcripts-for-zero-knowledge-proofs-28d9fda22d9a) or something but for this borderline-legacy code I think it's best not to try anything too extreme.)

### DIFF
--- a/include/secp256k1_whitelist.h
+++ b/include/secp256k1_whitelist.h
@@ -101,8 +101,6 @@ SECP256K1_API int secp256k1_whitelist_signature_serialize(
  *         online_seckey: the secret key to the signer's online pubkey
  *         summed_seckey: the secret key to the sum of (whitelisted key, signer's offline pubkey)
  *         index: the signer's index in the lists of keys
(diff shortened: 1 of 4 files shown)


jonasnick (commented):
ACK 11d675dce8edab6cece4524c3ffb3de809bea72f

real-or-random (commented):
utACK 11d675dce8edab6cece4524c3ffb3de809bea72f

PR #164, opened 2022-01-15
Change IsSecond to match the design section

kallerosenbaum:
The third bullet in the design section says that the second unique key gets coefficient 1, but the algorithm doesn't match that.

This change will make the algo description match the design section.

### DIFF
--- a/src/modules/musig/musig-spec.mediawiki
+++ b/src/modules/musig/musig-spec.mediawiki
@@ -84,7 +84,7 @@ The algorithm ''HashKeys(pk<sub>1..u</sub>)'' is defined as:
 The algorithm ''IsSecond(pk<sub>1..u</sub>, i)'' is defined as:
 * For ''j = 1 .. u'':
 ** If ''pk<sub>j</sub> &ne; pk<sub>1</sub>'':
-*** Return ''true'' if ''pk<sub>j</sub> = pk<sub>i</sub>'', otherwise return ''false''.
+*** Return ''true'' if ''pk<sub>j</sub> != pk<sub>i</sub>'', otherwise return ''false''.
 * Return ''false''
 
 The algorithm ''KeyAggCoeff(pk<sub>1..u</sub>, i)'' is defined as:


jonasnick (commented):
Hm, no the algorithm is correct. With your change, `IsSecond(pk<sub>1..u</sub>, 1)` will return true which is not what we intend.

real-or-random:
It should be correct as it is. Say we have pk1 != pk2

Then if I'm not mistaken IsSecond([pk1, pk2], pk2) = true.

kallerosenbaum:
Run KeyAggCoeff([77, 43], 2) // 1-based index
IsSecond([77,43],2) will return false and a_2 will become != 1
The result is that the second unique key will not get coefficient 1, which contradicts the Design section.

kallerosenbaum:
Oops, forget about this. IsSecond is correct.

jonasnick:
This part of the spec has already confused more people https://github.com/ElementsProject/secp256k1-zkp/pull/157#discussion_r777616214 and others. I'm wondering if we can write this down more clearly.

real-or-random:
Maybe we can write a FindSecondUniqueKey which differs because
 * it will return a single pubkey and not work on indices (I think this is be clearer)
 * it's more natural because it iterates over they keys only once (this is how you'd implement it)

But I think the core of the problem is that the algorithm looks wrong at first glance. (I thought the same when I first saw it). But that's hard to fix. The algorithm itself is correct and as simple as it gets. Our brains here are wrong, not the spec. ;)

real-or-random:
No matter how we write the algorithm, we could add a few input/output examples

kallerosenbaum:
One slightly related question:

If the keys are [44, 80, 80], then IsSecond([44, 80, 80], 3) would return true, right? This means that if the second unique key is repeated, all of them will get coefficient 1. Is that intentional? It seems intentional because the check compares actual keys, and not indices in "Return true if pkj = pki, otherwise return false"

jonasnick:
@kallerosenbaum Yes this is intentional and the variant that is proven correct in the appendix of the MuSig2 paper (we call this variant MuSig2*).

PR #167, opened 2022-02-13
Add ordinary and x-only tweaking to spec and simplify implementation

jonasnick:
~~This is a draft because I still need to add some comment to the implementation that explains how exactly we make `tweak` identical to `t` in `tweak_add`.~~

Based on #171

### DIFF
--- a/doc/musig-spec.mediawiki
+++ b/doc/musig-spec.mediawiki
@@ -81,18 +81,25 @@ The algorithm '''''KeySort(pk<sub>1..u</sub>)''''' is defined as:
 Input:
 * The number ''u'' of public keys with ''0 < u < 2^32''
 * The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The number ''v'' of tweaks with ''0 &le; v < 2^32''
+* The tweaks ''tweak<sub>1..v</sub>'': ''v'' 32-byte arrays
+* The tweak methods ''is_xonly_t<sub>1..v</sub>'' : ''v'' booleans
 
-The algorithm '''''KeyAgg(pk<sub>1..u</sub>)''''' is defined as:
-* Let ''Q = KeyAggInternal(pk<sub>1..u</sub>)''; fail if that fails.
+The algorithm '''''KeyAgg(pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)''''' is defined as:
+* Let ''(Q,_,_) = KeyAggInternal(pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)''; fail if that fails.
 * Return ''bytes(Q)''.
 
-The algorithm '''''KeyAggInternal(pk<sub>1..u</sub>)''''' is defined as:
+The algorithm '''''KeyAggInternal(pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)''''' is defined as:
 * For ''i = 1 .. u'':
 ** Let ''a<sub>i</sub> = KeyAggCoeff(pk<sub>1..u</sub>, pk<sub>i</sub>)''.
 ** Let ''P<sub>i</sub> = point(pk<sub>i</sub>)''; fail if that fails.
-* Let ''Q = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
-* Fail if ''is_infinite(Q)''.
-* Return ''Q''.
+* Let ''Q<sub>0</sub> = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(Q<sub>0</sub>)''.
+* Let ''tacc<sub>0</sub> = 0''
+* Let ''gacc<sub>0</sub> = 1''
+* For ''i = 1 .. v'':
+** Let ''(Q<sub>i</sub>, gacc<sub>i</sub>, tacc<sub>i</sub>) = Tweak(Q<sub>i-1</sub>, gacc<sub>i-1</sub>, tweak<sub>i</sub>, tacc<sub>i-1</sub>, is_xonly_t<sub>i</sub>)''; fail if that fails
+* Return ''(Q<sub>v</sub>, gacc<sub>v</sub>, tacc<sub>v</sub>)''.
 
 The algorithm '''''HashKeys(pk<sub>1..u</sub>)''''' is defined as:
 * Return ''hash<sub>KeyAgg list</sub>(pk<sub>1</sub> || pk<sub>2</sub> || ... || pk<sub>u</sub>)''
@@ -108,6 +115,17 @@ The algorithm '''''KeyAggCoeff(pk<sub>1..u</sub>, pk')''''' is defined as:
 * If ''IsSecond(pk<sub>1..u</sub>, pk')'':
 ** Return 1
 * Return ''int(hash<sub>KeyAgg coefficient</sub>(L || pk')) mod n''
+
+The algorithm '''''Tweak(Q<sub>i-1</sub>, gacc<sub>i-1</sub>, tweak<sub>i</sub>, tacc<sub>i-1</sub>, is_xonly_t<sub>i</sub>)''''' is defined as:
+* If ''is_xonly_t<sub>i</sub>'' and ''not has_even_y(Q<sub>i-1</sub>)'':
+** Let ''g<sub>i-1</sub> = -1 mod n''
+* Else: let ''g<sub>i-1</sub> = 1''
+* Let ''t<sub>i</sub> = int(tweak<sub>i</sub>)''; fail if ''t &ge; n''
+* Let ''Q<sub>i</sub> = g<sub>i-1</sub>⋅Q<sub>i-1</sub> + t<sub>i</sub>⋅G''
+** Fail if ''is_infinite(Q<sub>i</sub>)''
+* Let ''gacc<sub>i</sub> = g<sub>i-1</sub>⋅gacc<sub>i-1</sub> mod n''
+* Let ''tacc<sub>i</sub> = t<sub>i</sub> + g<sub>i-1</sub>⋅tacc<sub>i-1</sub> mod n''
+* Return ''(Q<sub>i</sub>, gacc<sub>i</sub>, tacc<sub>i</sub>)''
 
 ==== Nonce Generation ====
 
@@ -137,22 +155,25 @@ The Session Context is a data structure consisting of the following elements:
 * The aggregate public nonce ''aggnonce'': a 66-byte array
 * The number ''u'' of public keys with ''0 < u < 2^32''
 * The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The number ''v'' of tweaks with ''0 &le; v < 2^32''
+* The tweaks ''tweak<sub>1..v</sub>'': ''v'' 32-byte arrays
+* The tweak methods ''is_xonly_t<sub>1..v</sub>'' : ''v'' booleans
 * The message ''m'': a 32-byte array
 
-We write "Let ''(aggnonce, u, pk<sub>1..u</sub>, m) = session_ctx''" to assign names to the elements of a Session Context.
+We write "Let ''(aggnonce, u, pk<sub>1..u</sub>, v, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>, m) = session_ctx''" to assign names to the elements of a Session Context.
 
 The algorithm '''''GetSessionValues(session_ctx)''''' is defined as:
-* Let ''(aggnonce, u, pk<sub>1..u</sub>, m) = session_ctx''
-* Let ''Q = KeyAggInternal(pk<sub>1..u</sub>)''; fail if that fails
+* Let ''(aggnonce, u, pk<sub>1..u</sub>, v, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>, m) = session_ctx''
+* Let ''(Q, gacc<sub>v</sub>, tacc<sub>v</sub>) = KeyAggInternal(pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)''; fail if that fails
 * Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce || bytes(Q) || m)) mod n''
 * Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
 * Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
 * Fail if ''is_infinite(R)''
 * Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
-* Return ''(Q, b, R, e)''
+* Return ''(Q, gacc<sub>v</sub>, tacc<sub>v</sub>, b, R, e)''
 
 The algorithm '''''GetSessionKeyAggCoeff(session_ctx, P)''''' is defined as:
-* Let ''(_, u, pk<sub>1..u</sub>, _) = session_ctx''
+* Let ''(_, u, pk<sub>1..u</sub>, _, _, _, _) = session_ctx''
 * Return ''KeyAggCoeff(pk<sub>1..u</sub>, bytes(P))''
 
 ==== Signing ====
@@ -163,7 +184,7 @@ Input:
 * The ''session_ctx'': a [[#session-context|Session Context]] data structure
 
 The algorithm '''''Sign(secnonce, sk, session_ctx)''''' is defined as:
-* Let ''(Q, b, R, e) = GetSessionValues(session_ctx)''; fail if that fails
+* Let ''(Q, gacc<sub>v</sub>, _, b, R, e) = GetSessionValues(session_ctx)''; fail if that fails
 * Let ''k'<sub>1</sub> = int(secnonce[0:32]), k'<sub>2</sub> = int(secnonce[32:64])''
 * Fail if ''k'<sub>i</sub> = 0'' or ''k'<sub>i</sub> &ge; n'' for ''i = 1..2''
 * Let ''k<sub>1</sub> = k'<sub>1</sub>, k<sub>2</sub> = k'<sub>2</sub> '' if ''has_even_y(R)'', otherwise let ''k<sub>1</sub> = n - k'<sub>1</sub>, k<sub>2</sub> = n - k<sub>2</sub>''
@@ -171,7 +192,9 @@ The algorithm '''''Sign(secnonce, sk, session_ctx)''''' is defined as:
 * Fail if ''d' = 0'' or ''d' &ge; n''
 * Let ''P = d'⋅G''
 * Let ''mu = GetSessionKeyAggCoeff(session_ctx, P)''; fail if that fails
-* Let ''d = n - d' '' if ''has_even_y(P) `XOR` has_even_y(Q)'', otherwise let ''d = d' ''
+* Let ''g<sub>v</sub> = 1'' if ''has_even_y(Q)'', otherwise let ''g<sub>v</sub> = -1 mod n''
+* Let ''gp = 1'' if ''has_even_y(P)'', otherwise let ''gp = -1 mod n''
+* Let ''d = g<sub>v</sub>⋅gacc<sub>v</sub>⋅gp⋅d' ''
 * Let ''s = (k<sub>1</sub> + b⋅k<sub>2</sub> + e⋅mu⋅d) mod n''
 * Let ''psig = bytes(s)''
 * Let ''pubnonce = cbytes(k'<sub>1</sub>⋅G) || cbytes(k'<sub>2</sub>⋅G)''
@@ -185,12 +208,15 @@ Input:
 * The number ''u'' of public nonces and public keys with ''0 < u < 2^32''
 * The public nonces ''pubnonce<sub>1..u</sub>'': ''u'' 66-byte arrays
 * The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
+* The number ''v'' of tweaks with ''0 &le; v < 2^32''
+* The tweaks ''tweak<sub>1..v</sub>'': ''v'' 32-byte arrays
+* The tweak methods ''is_xonly_t<sub>1..v</sub>'' : ''v'' booleans
 * The message ''m'': a 32-byte array
 * The index of the signer ''i'' in the public nonces and public keys with ''0 < i &le; u''
 
-The algorithm '''''PartialSigVerify(psig, pubnonce<sub>1..u</sub>, pk<sub>1..u</sub>, m, i)''''' is defined as:
+The algorithm '''''PartialSigVerify(psig, pubnonce<sub>1..u</sub>, pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>, m, i)''''' is defined as:
 * Let ''aggnonce = NonceAgg(pubnonce<sub>1..u</sub>)''; fail if that fails
-* Let ''session_ctx = (aggnonce, u, pk<sub>1..u</sub>, m)''
+* Let ''session_ctx = (aggnonce, u, pk<sub>1..u</sub>, v, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>, m)''
 * Run ''PartialSigVerifyInternal(psig, pubnonce<sub>i</sub>, pk<sub>i</sub>, session_ctx)''
 * Return success iff no failure occurred before reaching this point.
 
@@ -203,13 +229,14 @@ Input:
 * The ''session_ctx'': a [[#session-context|Session Context]] data structure
 
 The algorithm '''''PartialSigVerifyInternal(psig, pubnonce, pk<sup>*</sup>, session_ctx)''''' is defined as:
-* Let ''(Q, b, R, e) = GetSessionValues(session_ctx)''; fail if that fails
+* Let ''(Q, gacc<sub>v</sub>, _, b, R, e) = GetSessionValues(session_ctx)''; fail if that fails
 * Let ''s = int(psig)''; fail if ''s &ge; n''
 * Let ''R<sup>*</sup><sub>1</sub> = pointc(pubnonce[0:33]), R<sup>*</sup><sub>2</sub> = pointc(pubnonce[33:66])''
 * Let ''R<sup>*</sup>' = R<sup>*</sup><sub>1</sub> + b⋅R<sup>*</sup><sub>2</sub>''
 * Let ''R<sup>*</sup> = R<sup>*</sup>' '' if ''has_even_y(R)'', otherwise let ''R<sup>*</sup> = -R<sup>*</sup>' ''
-* Let ''P' = point(pk<sup>*</sup>)''; fail if that fails
-* Let ''P = P' '' if ''has_even_y(Q)'', otherwise let ''P = -P' ''
+* Let ''g<sub>v</sub> = 1'' if ''has_even_y(Q)'', otherwise let ''g<sub>v</sub> = -1 mod n''
+* Let ''g' = g<sub>v</sub>⋅gacc<sub>v</sub> mod n''
+* Let ''P = g'⋅point(pk<sup>*</sup>)''; fail if that fails
 * Let ''mu = GetSessionKeyAggCoeff(session_ctx, P)''; fail if that fails
 * Fail if ''s⋅G &ne; R<sup>*</sup> + e⋅mu⋅P''
 * Return success iff no failure occurred before reaching this point.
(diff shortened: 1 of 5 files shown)


doc/musig-spec.mediawiki
```diff
@@ -147,35 +147,47 @@ The algorithm ''NonceAgg(pubnonce<sub>1..u</sub>)'' is defined as:
...
+The algorithm ''GetSessionValues(session_ctx)'' is defined as:
 * Let ''Q = KeyAggInternal(pk<sub>1..u</sub>)''; fail if that fails
 * Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce || bytes(Q) || m)) mod n''
+* Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
 * Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
 * Fail if ''is_infinite(R)''
+* Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
+* Return ''(Q, b, R, e)''
+
+The algorithm ''GetSessionKeyCoef(session_ctx, P)'' is defined as:
```
robot-dreams, line 282: Nit: `Coeff` for consistency?

robot-dreams (commented):
Reviewed up to dcc485daaa303a78ba0ccb80090db5648e1d38c8, maybe those first few commits can be split out and merged soon?

Overall the changes (in that range) look good, I just have some cosmetic comments. It was straightforward to directly adapt the spec updates to make the reference implementation use a "Session Context", see 9e773e6b73283096a6dd5a3b56c42803ae8947fa

doc/musig-spec.mediawiki
```diff
@@ -147,35 +147,47 @@ The algorithm ''NonceAgg(pubnonce<sub>1..u</sub>)'' is defined as:
...
-* The secret key ''sk'': a 32-byte array
+The Session Context is a data structure consisting of the following elements:
 * The aggregate public nonce ''aggnonce'': a 66-byte array
 * The number ''u'' of public keys with ''0 < u < 2^32''
 * The public keys ''pk<sub>1..u</sub>'': ''u'' 32-byte arrays
 * The message ''m'': a 32-byte array
 
-The algorithm ''Sign(secnonce, sk, aggnonce, pk<sub>1..u</sub>, m)'' is defined as:
-* Let ''R<sub>1</sub> = pointc(aggnonce[0:33]), R<sub>2</sub> = pointc(aggnonce[33:66])''; fail if that fails
+The algorithm ''GetSessionValues(session_ctx)'' is defined as:
```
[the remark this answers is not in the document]
jonasnick: Added the assignmen, an explanation what that means and a definition for _tuples_.

doc/musig-spec.mediawiki
```diff
@@ -190,44 +202,39 @@ Input:
...
-* Let ''Q = KeyAggInternal(pk<sub>1..u</sub>)''; fail if that fails
-* Let ''b = int(hash<sub>MuSig/noncecoef</sub>(aggnonce || bytes(Q) || m)) mod n''
-* Let ''R = R<sub>1</sub> + b⋅R<sub>2</sub>''
 * Let ''R<sup>*</sup><sub>1</sub> = pointc(pubnonce[0:33]), R<sup>*</sup><sub>2</sub> = pointc(pubnonce[33:66])''
 * Let ''R<sup>*</sup>' = R<sup>*</sup><sub>1</sub> + b⋅R<sup>*</sup><sub>2</sub>''
 * Let ''R<sup>*</sup> = R<sup>*</sup>' '' if ''has_even_y(R)'', otherwise let ''R<sup>*</sup> = -R<sup>*</sup>' ''
-* Let ''e = int(hash<sub>BIP0340/challenge</sub>(bytes(R) || bytes(Q) || m)) mod n''
-* Let ''mu = KeyAggCoeff(pk<sub>1..u</sub>, pk<sup>*</sup>)''
 * Let ''P' = point(pk<sup>*</sup>)''; fail if that fails
+* Let ''mu = SessionGetKeyCoef(session_ctx, P)''; fail if that fails
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
jonasnick: fixed

jonasnick:
After receiving some feedback (H/T @brandonblack) and looking at this PR again, I'm not sure that this is the right way to specify tweaking. Yes, the spec itself is elegant, but the implementation is absolutely not.

In short, the spec is elegant because it only has one keygen algorithm that does keygen and all tweaks at once. Therefore, we can avoid having keygen state. However, in an implementation we want to be more flexible. For example, we want to apply two different tweaks on the same aggregate key without having to recompute the aggregate key. This requires deviating quite a bit from the spec pseudocode. I made it extra complicated in order to be able to say that `parity_factor` always equals `g` in the spec. Perhaps we should change the spec to be more practical, at the cost of introducing key aggregation/tweaking state. I'll have to try this out.

> maybe those first few commits can be split out and merged soon? 

@robot-dreams done, see #171

brandonblack:
I'm futzing with my implementation of tweaking to see if I can reverse it back to something that would make sense as a specification. So far all I've accomplished is introducing ~21 million state variables.

robot-dreams:
Quick question: can you give an example of where you'd want to interleave ordinary (BIP32) tweaks and x-only (BIP341) tweaks, as opposed to doing (1) all the ordinary tweaks, followed by (2) all the x-only tweaks?

doc/musig-spec.mediawiki
```diff
@@ -104,6 +223,22 @@ The algorithm ''KeyAggCoeff(pk<sub>1..u</sub>, pk')'' is defined as:
...
+** Let ''t<sub>i</sub> = int(tweak<sub>i</sub>)''; fail if ''t &ge; n''
+** Let ''Q<sub>i</sub> = g<sub>i-1</sub>⋅Q<sub>i-1</sub> + t<sub>i</sub>⋅G
+** Fail if ''is_infinite(Q<sub>i</sub>)''
+* Let ''g<sub>v</sub> = 1'' if ''has_even_y(Q<sub>v</sub>)'', otherwise let ''g<sub>v</sub> = -1''
+* Let ''g = g<sub>0</sub>⋅...⋅g<sub>v</sub>''
+* If ''v > 0'':
+** Let ''t = sum<sub>i=1</sub><sup>v</sup> t<sub>i</sub>⋅(prod<sub>j=i</sub><sup>v</sup> g<sub>j</sub> ) mod n''
+* Else: let ''t = 0''
+* Return ''Q<sub>v</sub>, g, t''
+
```
brandonblack, line 241:
```suggestion
The algorithm ''Tweak(Q, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)'' is defined as:
* Let ''Q<sub>0</sub> = Q''
* Let ''g<sub>0</sub> = not has_even_y(Q)''
* Let ''t<sub>0</sub> = 0''
* For ''i = 1 .. v'':
** Let ''t'<sub>i</sub> = int(tweak<sub>i</sub>)''; fail if ''t<sub>i</sub> &ge; n''
** Let ''t<sub>i</sub> = t<sub>i-1</sub> + t'<sub>i</sub> mod n'' if ''is_xonly_t<sub>i</sub>'' otherwise let ''t<sub>i</sub> = -(t<sub>i-1</sub> + t'<sub>i</sub>) mod n''
** Let ''Q'<sub>i</sub> = Q<sub>i-1</sub>'' if ''not is_xonly_t<sub>i</sub>'' or ''has_even_y(Q<sub>i-1</sub>)'' otherwise let ''Q'<sub>i</sub> = -Q<sub>i-1</sub>''
** Let ''Q<sub>i</sub> = Q'<sub>i</sub> + t'<sub>i</sub>⋅G''
** Fail if ''is_infinite(Q<sub>i</sub>)''
** Let ''g'<sub>i</sub> = g<sub>i-1</sub>'' if ''is_xonly_t<sub>i</sub>'' or ''has_even_y(Q<sub>i-1</sub>)'' otherwise let ''g'<sub>i</sub> = not g<sub>i-1</sub>''
** Let ''g<sub>i</sub> = g'<sub>i</sub>'' if ''has_even_y(Q<sub>i</sub>)'' otherwise let ''g<sub>i</sub> = not g'<sub>i</sub>''
* Return ''Q<sub>v</sub>, g<sub>v</sub>, t<sub>v</sub>''
```
I believe this matches the logic of my code. Not saying it is necessarily better, but it does have a certain flow to it.

jonasnick:
@robot-dreams I can't imagine an example for this. But I don't think allowing interleaving adds any complexity.

brandonblack:
OK, sorry for the first false start there -- I thought I had a fully working tweak function, but discovered bugs when testing it against the vectors in `tests_impl.h` instead of just my own happy path (thanks for those).

Here's a specification for my working (really this time) algorithm:
```
The algorithm ''Tweak(Q, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)'' is defined as:
* Let ''Q<sub>0</sub> = Q''
* Let ''g<sub>0</sub> = not has_even_y(Q)''
* Let ''t<sub>0</sub> = 0''
* For ''i = 1 .. v'':
** Let ''t'<sub>i</sub> = int(tweak<sub>i</sub>)''; fail if ''t<sub>i</sub> &ge; n''
** Let ''Q'<sub>i</sub> = -Q<sub>i-1</sub> if ''is_xonly_t<sub>i</sub>'' and ''not has_even_y(Q<sub>i-1</sub>)'' otherwise let ''Q'<sub>i</sub> = Q<sub>i-1</sub>
** Let ''Q<sub>i</sub> = Q'<sub>i</sub> + t'<sub>i</sub>⋅G''
** If ''is_xonly_t<sub>i</sub>'' or ''not has_even_y(Q<sub>i-1</sub>)'':
*** Let ''t"<sub>i</sub> = t<sub>i-1</sub> + t'<sub>i</sub> mod n''
*** Let ''g'<sub>i</sub> = g<sub>i-1</sub>''
** Else:
*** Let ''t"<sub>i</sub> = n - t<sub>i-1</sub> + t'<sub>i</sub> mod n''
*** Let ''g'<sub>i</sub> = not g<sub>i-1</sub>''
** Let ''(t<sub>i</sub>, g<sub>i</sub>) = (t"<sub>i</sub>, g'<sub>i</sub>)'' if ''has_even_y(Q<sub>i</sub>)'' otherwise let ''(t<sub>i</sub>, g<sub>i</sub>) = (n - t"<sub>i</sub>, not g'<sub>i</sub>)''
* Return ''Q<sub>v</sub>, g<sub>v</sub>, t<sub>v</sub>''
```

This treats `g` as parity, rather than introducing the "parity factor" concept. It can also be easily adapted to adding a tweak given `G_p`, `g_p`, `t_p`, which would require more changes elsewhere. This gist has both this version, and the "add tweak" version: https://gist.github.com/brandonblack/8862eed68619dc52555d77027f962a1b

Here's my implementation that logically matches the "add tweak"-style spec:
https://github.com/brandonblack/musig-js/blob/0d67928b6263096c9655304b3d370fbb47c3fe3b/index.ts#L283

jonasnick (commented):
> I'm not sure that this is the right way to specify tweaking. Yes, the spec itself is elegant, but the implementation is absolutely not.

Ok, now I have a much better solution that is also just slightly different to the previous proposal. It allows having a simple implementation that is close to the spec and the spec doesn't require to include a lot of keygen state. This was achieved by not including `g[v]` in `g`. I archived the old branch [here](https://github.com/jonasnick/secp256k1-zkp/commits/musig-spec-tweak-single-g).

@brandonblack Thank you! I think a downside of your proposal is that it's not as easy to justify as the current algorithm in the "Negation Of The Secret Key When Signing" section. In particular, switching the `g` values to booleans and letting `g[i]` depend on `g[i-1]` seems harder to deal with.

brandonblack (commented):
Removed comment about mis-push

doc/musig-spec.mediawiki
```diff
@@ -47,6 +48,7 @@ The following conventions are used, with constants as defined for [https://www.s
 ** The function ''bytes(x)'', where ''x'' is an integer, returns the 32-byte encoding of ''x'', most significant byte first.
 ** The function ''bytes(P)'', where ''P'' is a point, returns ''bytes(x(P))''.
 ** The function ''has_even_y(P)'', where ''P'' is a point for which ''not is_infinite(P)'', returns ''y(P) mod 2 = 0''.
+** The function ''set_even_y(P)'', where ''P'' is a point, returns ''P'' if ''is_infinite(P)'' or ''has_even_y(P)''. Otherwise,  ''set_even_y(P)'' returns ''-P''.
```
real-or-random, line 51:
```suggestion
** The function ''with_even_y(P)'', where ''P'' is a point, returns ''P'' if ''is_infinite(P)'' or ''has_even_y(P)''. Otherwise,  ''with_even_y(P)'' returns ''-P''.
```

nit: "set" is misleading for a pure function.

doc/musig-spec.mediawiki
```diff
@@ -28,8 +28,9 @@ This document is licensed under the 2-clause BSD license.
 * The second unique key in the pubkey list given to ''KeyAgg'' (as well as any keys identical to this key) gets the constant KeyAgg coefficient 1 which saves an exponentiation (see the MuSig2* appendix in the [https://eprint.iacr.org/2020/1261 MuSig2 paper]).
 * The public key inputs are serialized using x-only (32 byte) instead of compressed (33 byte) serialization. The reason for this is that as x-only keys are becoming more common, the full key may not be available.
 * The public nonces are serialized in compressed format (33 bytes). We accept the small overhead compared to x-only serialization to avoid complicating the specification.
+* This specification supports signing for ''tweaked'' aggregate public keys. There are two modes of tweaking. ''Ordinary'' tweaking allows deriving child aggregate public keys per [https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki|BIP32]. ''X-only'' tweaking allows creating a [https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki|BIP341] Taproot tweak. See section [#tweaking|Tweaking] below for details.
```
real-or-random, line 31:
```suggestion
* This specification supports signing for ''tweaked'' aggregate public keys. There are two modes of tweaking. ''Ordinary'' tweaking allows deriving child aggregate public keys per [https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki BIP32]. ''X-only'' tweaking allows creating a [https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki BIP341] Taproot tweak. See section [#tweaking Tweaking] below for details.
```

This is wrong also in other places below.

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+
+=== Negation Of The Secret Key When Signing ===
+
+In order to produce a partial signature for an x-only public key that is an aggregate of x-only keys and x-only or ordinarily tweaked, the ''[[#Sign negation|Sign]]'' algorithm may need to negate the secret key during the signing process.
+
+<poem>
+The following public keys arise as intermediate steps in the MuSig protocol:
+• ''P<sub>i</sub>'' as computed in ''KeyAggInternal'' is the point corresponding to the ''i''-th signer's x-only public key. Defining ''d'<sub>i</sub>'' to be the ''d' '' value as computed in the ''Sign'' algorithm of the ''i''-th signer, we have
+    ''P<sub>i</sub> = set_even_y(d'<sub>i</sub>⋅G) ''.
+• ''Q<sub>0</sub>'' as defined in ''Tweak'' is the aggregate public key computed in ''KeyAggInternal'' as
```
real-or-random, line 88: defined in which Tweak? "OrdinaryTweak or XonlyTweak"?
But even then I can't fully parse this. There's no Q_0 defined in either of the tweak algorithms..

edit: Ah you mean the algorithm below. So this was very confusing to me. Maybe point out that it can be found below. But I think then it's still not optimal for the reader because this here requires that one has read later parts. Maybe it's better to move this entire "correctness proof"/ explanation below the spec algos?

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+* Return P + t⋅G
+
+The algorithm ''XonlyTweak(P, t)'' is defined as:
+* Return set_even_y(P) + t⋅G
+
+=== Negation Of The Secret Key When Signing ===
+
+In order to produce a partial signature for an x-only public key that is an aggregate of x-only keys and x-only or ordinarily tweaked, the ''[[#Sign negation|Sign]]'' algorithm may need to negate the secret key during the signing process.
+
+<poem>
```
real-or-random, line 84: nit: Could need some rhymes.

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+
+The algorithm ''OrdinaryTweak(P, t)'' is defined as:
+* Return P + t⋅G
+
+The algorithm ''XonlyTweak(P, t)'' is defined as:
+* Return set_even_y(P) + t⋅G
+
+=== Negation Of The Secret Key When Signing ===
+
+In order to produce a partial signature for an x-only public key that is an aggregate of x-only keys and x-only or ordinarily tweaked, the ''[[#Sign negation|Sign]]'' algorithm may need to negate the secret key during the signing process.
```
real-or-random, line 82: I think it will be good to introduce u and v already in this sentence.

edit: I think this is related to my comment below that the sections could be reordered.

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+    ''P<sub>i</sub> = set_even_y(d'<sub>i</sub>⋅G) ''.
+• ''Q<sub>0</sub>'' as defined in ''Tweak'' is the aggregate public key computed in ''KeyAggInternal'' as
+    ''Q<sub>0</sub> = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''.
+• ''Q<sub>i</sub>'' as computed in ''Tweak'' for ''1 &le; i &le; v'' is the tweaked public key after the ''i''-th tweaking operation. It holds that
+    ''Q<sub>i</sub> = f(i-1) + t<sub>i</sub>⋅G'' for ''i = 1, ..., v'' where
+        ''f(i) := set_even_y(Q<sub>i</sub>)'' if ''is_xonly_t<sub>i+1</sub>'' and
+        ''f(i) := Q<sub>i</sub>'' otherwise.
+</poem>
+
+The goal is to produce a partial signature corresponding to the output of ''KeyAgg'', i.e., the final (x-only) public key point after ''v'' tweaking operations ''set_even_y(Q<sub>u</sub>)''.
```
real-or-random, line 96: Should this be Q_v instead of Q_u?

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+    ''set_even_y(Q<sub>v</sub>) - t*G
+        = g⋅Q<sub>0</sub>
+        = g⋅(a<sub>1</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>)
+        = g⋅(a<sub>1</sub>⋅gp<sub>1</sub>⋅d'<sub>1</sub>⋅G + ... + a<sub>u</sub>⋅gp<sub>u</sub>⋅d'<sub>u</sub>⋅G)
+        = sum<sub>i=1</sub><sup>u</sup>(g⋅gp<sub>i</sub>⋅a<sub>i</sub>⋅d'<sub>i</sub>)*G''.
+</poem>
+
+Thus, signer ''i'' multiplies its secret key ''d'<sub>i</sub>'' with ''g⋅gp<sub>i</sub>'' in the ''[[#Sign negation|Sign]]'' algorithm.
+
+=== Negation Of The Public Key When Verifying ===
```
real-or-random, line 137:
```suggestion
=== Negation Of The Public Key When Partially Verifying ===
```

doc/musig-spec.mediawiki
```diff
@@ -76,18 +191,22 @@ The algorithm ''KeySort(pk<sub>1..u</sub>)'' is defined as:
...
+The algorithm ''KeyAggInternal(pk<sub>1..u</sub>, tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)'' is defined as:
 * For ''i = 1 .. u'':
 ** Let ''a<sub>i</sub> = KeyAggCoeff(pk<sub>1..u</sub>, pk<sub>i</sub>)''.
 ** Let ''P<sub>i</sub> = point(pk<sub>i</sub>)''; fail if that fails.
-* Let ''Q = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
-* Fail if ''is_infinite(Q)''.
-* Return ''Q''.
+* Let ''Q' = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(Q')''.
+* Let ''Q, g, t = Tweak(Q', tweak<sub>1..v</sub>, is_xonly_t<sub>1..v</sub>)''; fail if that fails
```
real-or-random, line 208: I think further down you parse tuples using a parentheses syntax.

doc/musig-spec.mediawiki
```diff
@@ -60,7 +62,120 @@ The following conventions are used, with constants as defined for [https://www.s
...
+        = g<sub>v</sub>⋅Q<sub>v</sub>
+        = g<sub>v</sub>⋅(f(v-1) + t<sub>v</sub>⋅G)
+        = g<sub>v</sub>⋅(g<sub>v-1</sub>⋅(f(v-2) + t<sub>v-1</sub>⋅G) + t<sub>v</sub>⋅G)
+        = g<sub>v</sub>⋅...⋅g<sub>0</sub>⋅Q<sub>0</sub> + (g<sub>v</sub>⋅t<sub>v</sub>+g<sub>v</sub>⋅g<sub>v-1</sub>⋅t<sub>v-1</sub> + ... + g<sub>v</sub>⋅g<sub>v-1</sub>⋅...⋅g<sub>1</sub>⋅t<sub>1</sub>)⋅G''.
+</poem>
+
+<poem>
+The ''KeyAgg'' algorithm computes
+    ''g = g<sub>v</sub>⋅...⋅g<sub>0</sub>
+    t = sum<sub>i=1</sub><sup>v</sup> t<sub>i</sub>⋅(prod<sub>j=i</sub><sup>v</sup> g<sub>j</sub> ) mod n
```
real-or-random, line 121: sub and sup don't look nice together. I'd suggest to switch to a `sum<sub>i=1..v</sub>` style

real-or-random (commented):
I haven't read the algorithms in detail (and I think it will be easier to do this once you've addressed some of the comments) but I think it's a very good idea to have this tweaking section that explains why the algorithms are as they are.

doc/musig-spec.mediawiki
jonasnick: true, fixed

doc/musig-spec.mediawiki
jonasnick: I did find it truly poetic that after quite a while of searching I found that this is the best way of formatting things. But, formatting still sucks and I didn't pay much attention to optimize it (yet).

doc/musig-spec.mediawiki
jonasnick: thx, fixed

doc/musig-spec.mediawiki
jonasnick: done

doc/musig-spec.mediawiki
```diff
@@ -184,64 +318,64 @@ Input:
...
+* The ''session_ctx'': a [[#session-context|Session Context]] data structure
 
-The algorithm ''SigAgg(R, sig<sub>1..u</sub>)'' is defined as:
+The algorithm ''PartialSigAgg(psig<sub>1..u</sub>, session_ctx)'' is defined as:
+* Let ''(Q, _, t, _, _, R, e) = GetSessionValues(session_ctx)''; fail if that fails
 * For ''i = 1 .. u'':
-** Let ''s<sub>i</sub> = int(sig<sub>i</sub>)''; fail if ''s<sub>i</sub> &ge; n''.
-* Let ''s = s<sub>1</sub> + ... + s<sub>u</sub> mod n''
+** Let ''s<sub>i</sub> = int(psig<sub>i</sub>)''; fail if ''s<sub>i</sub> &ge; n''.
+* Let ''g<sub>v</sub> = 1'' if ''has_even_y(Q)'', otherwise let ''g<sub>v</sub> = -1''
```
brandonblack, line 365: Not sure how much this matters, but this is a notable change in requirement for the partial signature aggregator role. It now needs to have access to the XY aggregate public key.

doc/musig-spec.mediawiki
```diff
@@ -76,18 +191,27 @@ The algorithm ''KeySort(pk<sub>1..u</sub>)'' is defined as:
...
-* Let ''Q = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
-* Fail if ''is_infinite(Q)''.
-* Return ''Q''.
+* Let ''Q<sub>0</sub> = a<sub>1</sub>⋅P<sub>1</sub> + a<sub>2</sub>⋅P<sub>1</sub> + ... + a<sub>u</sub>⋅P<sub>u</sub>''
+* Fail if ''is_infinite(Q<sub>0</sub>)''.
+* For ''i = 1 .. v'':
+** Let ''Q<sub>i</sub>, g<sub>i-1</sub>, t<sub>i</sub> = Tweak(Q<sub>i-1</sub>, tweak<sub>i</sub>, is_xonly_t<sub>i</sub>)''; fail if that fails
+* If ''v > 0'':
+** Let ''g = g<sub>0</sub>⋅...⋅g<sub>v-1</sub>''
+** Let ''t = sum<sub>i=1</sub><sup>v</sup> t<sub>i</sub>⋅(prod<sub>j=i</sub><sup>v-1</sup> g<sub>j</sub> ) mod n''
```
brandonblack, line 212: It's not immediately obvious just how nicely this lends itself to implementing as an accumulator for `t`. The sum of product notation is difficult to translate to 'just negate the accumulator conditionally each step'. Perhaps some color text on that topic would be nice. Not necessary, as it can _also_ be implemented correctly by collecting all of the values, and doing the sums at the end.

brandonblack (commented):
Spec tACK (if that's a thing).

I modified my KeyAgg, Tweak, PartialSign, PartialVerify, SignAgg, and Session code to match this latest spec, and it was indeed much easier to get right.

Should the `Abstract` be updated to reflect that this document describes much more than Aggregation?

doc/musig-spec.mediawiki
jonasnick: > New way better.

Yeah agreed. In the secp-zkp implementation we do not require the final pubkey in aggregation but instead store the final (parity-included) tweak in the session object (which is a subset of the session datastructure of the spec).

doc/musig-spec.mediawiki
jonasnick: Good point. I think it's worth changing, but would require introducing g and t accumulators (as you suggested initially) and adjusting the explantion. I'll try this out.

jonasnick:
@brandonblack Great to hear that it's now easier to get right. Thanks for the feedback. I fixed the sentence in the abstract.

jonasnick (commented):
As @brandonblack suggested I moved the compution of t from KeyAggInternal to Tweak which demonstrates how to incrementally apply tweaks like we do in the implementation. I renamed `t` to `tacc[v]` and adjusted the explanation. Similarly, I moved the g computation inside Tweak and renamed it `gacc[v]`.

doc/musig-spec.mediawiki
jonasnick: I moved this section below the specification section. That seems moved reasonable because it's content not immediately relevant to someone implementing the specification.

doc/musig-spec.mediawiki
jonasnick: done

doc/musig-spec.mediawiki
```diff
@@ -60,53 +62,74 @@ The following conventions are used, with constants as defined for [https://www.s
...
 ** The function ''hash<sub>tag</sub>(x)'' where ''tag'' is a UTF-8 encoded tag name and ''x'' is a byte array returns the 32-byte hash ''SHA256(SHA256(tag) || SHA256(tag) || x)''.
+* Other:
+** Tuples are written by listing the elements within parentheses and separated by commas. For example, ''(2, 3, 1)'' is a tuple.
 
+=== Specification ===
 
 ==== Key Sorting ====
 
 Input:
 * The number ''u'' of signatures with ''0 < u < 2^32''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
```diff
@@ -60,53 +62,74 @@ The following conventions are used, with constants as defined for [https://www.s
...
 
+The algorithm '''''Tweak(Q<sub>i-1</sub>, gacc<sub>i-1</sub>, tweak<sub>i</sub>, tacc<sub>i-1</sub>, is_xonly_t<sub>i</sub>)''''' is defined as:
+* If ''is_xonly_t<sub>i</sub>'' and ''not has_even_y(Q<sub>i-1</sub>)'':
+** Let ''g<sub>i-1</sub> = -1 mod n''
+* Else: let ''g<sub>i-1</sub> = 1''
+* Let ''t<sub>i</sub> = int(tweak<sub>i</sub>)''; fail if ''t &ge; n''
+* Let ''Q<sub>i</sub> = g<sub>i-1</sub>⋅Q<sub>i-1</sub> + t<sub>i</sub>⋅G''
+* Let ''gacc<sub>i</sub> = g<sub>i-1</sub>⋅gacc<sub>i-1</sub> mod n''
+* Let ''tacc<sub>i</sub> = t<sub>i</sub> + g<sub>i-1</sub>⋅tacc<sub>i-1</sub> mod n''
+** Fail if ''is_infinite(Q<sub>i</sub>)''
```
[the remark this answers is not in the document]
jonasnick: fixed

doc/musig-spec.mediawiki
real-or-random: I think it's much better now. :)

real-or-random:
This is very nice.

ACK f1fad27181421b061a7b9b586c56bb04a68c41ec -- I haven't checked all the indices etc, so this is more of a Concept ACK than a "pseudocode review ACK" but we we have the ACK by Brandon and this is anyway still a draft, so I think this is good to be merged. 

Here are a few high-level comments that you could address here or in a later PR (whatever you feel seems easier):
 - The tweaking section should remind the reader that the purpose of specifying tweaking here is only to ensure compatibility with existing uses of tweaking. Draft:  "The purpose of specifying tweaking here is only to ensure compatibility with existing uses of tweaking, i.e., that the result of the signing is a valid signature with the tweaks applied. The algorithms here take as input arbitrary tweaks but allowing for arbitrary tweaks will render signatures insecure, and so the tweaks itself still need to obtained according to some other specification (e.g., [insert here]), which preserves security of tweaked signatures even if MuSig2 is used to generate the signatures. This typically involves deriving the tweaks from a hash of the public key given as input to the tweaking algorithm and some other information."
 - When describing the algorithms, we could write "x-only public keys"  instead of "public keys".
 - Since we have so many algorithms now, it's more elegant to drop the "the algorithm ... is defined as:" and just have the name of the algorithm in bold.

I share your impression that mediawiki is not optimal for algorithms but I have no good suggestion currently.

jonasnick:
@real-or-random Good points, but I agree that it's better to address this in a separate PR given that this one is already pretty extensive and has ACKs. I'll put this on the todo list.

Here's a short summary of design decisions taken during the lifetime of this PR:
- **Why don't we compute `g_v` already in `KeyAggInternal` and multiply it into `g`?** There is an important difference between the spec and an (efficient) implementation: in the spec we do tweaking inside keyagg, but an implementation would have tweaking functions separate from key aggregation. As a result, such an implementation does not know `v` when applying tweaks. It's much (!) simpler for an implementation to compute `g[v]` in the signing phase. The spec simply reflects that. Why don't we separate tweaking from key aggregation in the spec as well? That's possible, but would introduce a bit more bloat.
- **Why do we compute tacc inside Tweak and not in KeyAggInternal which would be simpler?** Again, this is to show that tweaking can easily be done separately from key aggregation.

PR #168, opened 2022-02-18
Avoid recomputing Ge normalized from Gej in Nonce process

sanket1729:
In secp256k1_musig_nonce_process(), we already have aggnonce in
normalized form. We don't need to recompute them again in
secp256k1_musig_nonce_process_internal().


src/modules/musig/session_impl.h
```diff
@@ -450,7 +447,7 @@ int secp256k1_musig_nonce_process(const secp256k1_context* ctx, secp256k1_musig_
         }
         secp256k1_gej_add_ge_var(&aggnonce_ptj[0], &aggnonce_ptj[0], &adaptorp, NULL);
```
real-or-random, line 448: Thanks for the PR! I think you need to take this line into account. The [0] aggnonce may have changed, so we need to convert the changed nonce back to ge. (I don't think the same is necessary for [1])

sanket1729: Moved the adaptor logic to the internal function.

1) When the adaptor is NULL, there is no need to convert between `gej` and `ge`.
2) When the adaptor is not NULL, we only convert the required aggnonce[0]

src/modules/musig/session_impl.h
```diff
@@ -442,15 +450,8 @@ int secp256k1_musig_nonce_process(const secp256k1_context* ctx, secp256k1_musig_
...
-    if (adaptor != NULL) {
-        secp256k1_ge adaptorp;
-        if (!secp256k1_pubkey_load(ctx, &adaptorp, adaptor)) {
-            return 0;
-        }
-        secp256k1_gej_add_ge_var(&aggnonce_ptj[0], &aggnonce_ptj[0], &adaptorp, NULL);
-    }
-    if (!secp256k1_musig_nonce_process_internal(&session_i.fin_nonce_parity, fin_nonce, &session_i.noncecoef, aggnonce_ptj, agg_pk32, msg32)) {
+
+    if (!secp256k1_musig_nonce_process_internal(ctx, &session_i.fin_nonce_parity, fin_nonce, &session_i.noncecoef, aggnonce_ptj, aggnonce_pt, agg_pk32, msg32, adaptor)) {
```
real-or-random, line 454: Wouldn't it be clearer to pass only the `ge`s to `secp256k1_musig_nonce_process_internal` and compute the `gej`s only there?

real-or-random (commented):
Concept ACK

PR #175, opened 2022-03-31
configure: Check compile+link when checking existence of functions

real-or-random:
Undeclared functions are fine in C but linking will fail.

This is a quick fix. In the long-term we should probably resort to `#ifdef`s as common in upstream.


jonasnick (commented):
ACK 79472c7ee556a7a3d2abc2e1ab077cfe120c78ff

PR #177, opened 2022-04-01
musig-spec: add optional arguments to strengthen nonce function

jonasnick:


doc/musig-spec.mediawiki
```diff
@@ -212,8 +214,20 @@ Input:
...
 ==== Nonce Generation ====
 
-'''''NonceGen()''''':
-* Generate two random integers ''k<sub>1</sub>, k<sub>2</sub>'' in the range ''1...n-1''
+Input:
+* The secret signing key ''sk'': a 32-byte array or 0-byte array (optional argument)
+* The aggregate public key ''aggpk'': a 32-byte array or 0-byte array (optional argument)
+* The message ''m'': a 32-byte array or 0-byte array (optional argument)
+* The length ''l'' of auxiliary entropy ''ent'': an integer ''&ge; 0''
+* The auxiliary entropy ''ent'': an l-byte array (optional argument)
```
real-or-random, line 222: If your pseudocode language is expressive compute `len(sk)`, couldn't it also compute `len(ent)` then?

real-or-random, line 222: We should also say explain the "security considerations" here, similarly to how the aux argument is explained in BIP340:
 * What is a good "ent" string? (For example, even some session identifier / context string can be good, or a counter can be good).
 * What are the implications of providing or not providing the arguments? This may be difficult to explain because there are so many optional arguments. But I think it boils down to  mostly case distinction between sk present vs sk not present.
 
Is "entropy" the right term here?  Maybe "context" is better? I don't know. A counter is useful here but what's the entropy of a counter? But in this case,  my suggestion "context" isn't much better... Another idea is just "aux" plus have a proper explanation.

real-or-random (commented):
concept ACK

doc/musig-spec.mediawiki
jonasnick: I didn't do this for consistency with other variable length arguments (number of pubkeys, nonces, etc) but I reconsidered and `len(ent)` is quite a bit nicer.

jonasnick: I tried avoiding `aux` to avoid confusion with the `aux` argument in BIP340 which also affects nonce generation but has a different purpose. I renamed the auxiliary input to `in` and added explanations to the NonceGeneration section.

PR #181, opened 2022-04-04
musig-spec: clarify hashing in noncegen by converting ints to bytes

jonasnick:


real-or-random (approved):
ACK 376733b58b282a4985dd78d0125749473f0aeff3

PR #186, opened 2022-04-05
musig-spec: Minor cleanup

robot-dreams:
- Remove all trailing whitespace `:%s/\s\+$//g`
- Say "second distinct key" instead of "second key" when describing Musig2* optimization
- Remove comments about sorting algorithms, since (1) even with thousands of keys sorting is unlikely to be a bottleneck, and (2) the majority of standard library implementations handle pathological cases, e.g. by using a worst-case n log n algorithm or falling back to one when necessary


jonasnick:
needs rebase

jonasnick (commented):
ACK 67247e53afdf32f414a9fbd0fb008b3935b1e6d9

PR #188, opened 2022-04-18
Add MuSig2 adaptor sig exercise in example code

instagibbs:
Code examples are clearer than doc explanations in a few ways, so figured I'd just submit this.


rage-proof:
I would like to see in the example `secp256k1_musig_extract_adaptor`, so to verify that the input and output secret-adaptor are identical.

examples/musig.c
```diff
@@ -156,7 +170,25 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
...
+    if (!secp256k1_musig_nonce_parity(ctx, &nonce_parity, &session)) {
+        return 0;
+    }
+    if (!secp256k1_musig_partial_sig_agg(ctx, presig, &session, partial_sigs, N_SIGNERS)){
+        return 0;
+    }
+    if (!secp256k1_musig_adapt(ctx, sig64, presig, adaptor_key, nonce_parity)) {
+        return 0;
+    }
+	/* With sig64 "on-chain" now, other party can grab the revealed adaptor secret */
```
real-or-random, line 184:
```suggestion
    /* With sig64 "on-chain" now, other party can grab the revealed adaptor secret */
```
wrong indentation

examples/musig.c
```diff
@@ -156,7 +170,25 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
             return 0;
         }
     }
-    return secp256k1_musig_partial_sig_agg(ctx, sig64, &session, partial_sigs, N_SIGNERS);
+
+    /* Since we are doing adaptor sig, complete pre-signature */
+    if (!secp256k1_musig_nonce_parity(ctx, &nonce_parity, &session)) {
+        return 0;
+    }
+    if (!secp256k1_musig_partial_sig_agg(ctx, presig, &session, partial_sigs, N_SIGNERS)){
```
real-or-random, line 178:
```suggestion
    if (!secp256k1_musig_partial_sig_agg(ctx, presig, &session, partial_sigs, N_SIGNERS)) {
```

examples/musig.c
```diff
@@ -156,7 +170,25 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
...
+        return 0;
+    }
+    if (!secp256k1_musig_adapt(ctx, sig64, presig, adaptor_key, nonce_parity)) {
+        return 0;
+    }
+	/* With sig64 "on-chain" now, other party can grab the revealed adaptor secret */
+    if (!secp256k1_musig_extract_adaptor(ctx, extracted_adaptor, sig64, presig, nonce_parity)) {
+        return 0;
+    }
+    if (memcmp(extracted_adaptor, adaptor_key, sizeof(adaptor_key)) != 0) {
```
real-or-random, line 188: Can you `assert` this? We normally use `assert` for these comparisons in the other examples.

examples/musig.c
```diff
@@ -201,7 +233,7 @@ int sign(const secp256k1_context* ctx, struct signer_secrets *signer_secrets, st
         return 1;
     }
     printf("ok\n");
-    printf("Verifying signature.....");
+    printf("Verifying signature and revealed adaptor.....");
```
real-or-random, line 236: I don't think we're verifying the adaptor here?

real-or-random (commented):
Concept ACK

I wonder if this example should go to a separate function. Adaptors add quite a lot to the flow and increase the brain power necessary to follow this example, so this is probably confusing for users who want to see a minimal example. If that means duplicating some of the code, this is not a problem, I think.

examples/musig.c
instagibbs: errr extracting revealed secret, I guess is the phrase?

real-or-random: I'm only saying that the printf not match the code below, which still only verifies a sig.

instagibbs: oh, I added this in the wrong spot

instagibbs: removed

examples/musig.c
instagibbs: done

examples/musig.c
instagibbs: done

examples/musig.c
instagibbs: removed tabs

PR #192, opened 2022-05-23
musig: update nonce generation to version 0.2.0 of the spec

jonasnick:
Draft until python reference implementation produces the same nonce generation test vectors.


src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,55 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
     return 1;
 }
 
+/* Write optional inputs into the hash */
+static void secp256k1_nonce_function_musig_helper(secp256k1_sha256 *sha, unsigned int n_bytes_len, const unsigned char *extra_in32) {
```
robot-dreams, line 230: Nit: I can see two possible points of confusion here:
- `extra_in32` could get confused with the `extra_in` auxiliary input
- `n_bytes_len` could get confused for the length of `extra_in32` because the `n, data` pattern is so common

How would you feel about these renames?
- `extra_in32` -> `data32`
- `n_bytes_len` -> `prefix_size`

robot-dreams (commented):
utACK ffedd26160b0253b79a1e95313f1eeb58394ce4d aside from renaming nit

src/modules/musig/session_impl.h
jonasnick: Oh yeah, done.

jonasnick:
Updated to latest version of https://github.com/jonasnick/bips/pull/19#issuecomment-1136309831

src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,54 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
     return 1;
 }
 
+/* Write optional inputs into the hash */
+static void secp256k1_nonce_function_musig_helper(secp256k1_sha256 *sha, unsigned int prefix_size, const unsigned char *data32) {
+    if (prefix_size > 1) {
+        /* This is necessary because the length of dataput in the spec is
```
robot-dreams, line 232: Nit: I assume it did find-and-replace `extra_input` -> `dataput` here?

robot-dreams (commented):
utACK f923f63b8f8a7274b42562acc04574d7603d2fe6

Using {0, 1} instead of {1, 2} makes sense. I did a quick skim of the spec to make sure there are no other places where we're using {0, 1} and didn't see any.

src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,54 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
     return 1;
 }
 
+/* Write optional inputs into the hash */
+static void secp256k1_nonce_function_musig_helper(secp256k1_sha256 *sha, unsigned int prefix_size, const unsigned char *data32) {
```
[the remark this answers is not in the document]
jonasnick: ok, done

src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,55 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
...
     for (i = 0; i < 2; i++) {
         unsigned char buf[32];
-        secp256k1_sha256_initialize(&sha);
-        secp256k1_sha256_write(&sha, seed, 32);
-        secp256k1_sha256_write(&sha, &i, sizeof(i));
-        secp256k1_sha256_finalize(&sha, buf);
+        sha_tmp = sha;
+        secp256k1_sha256_write(&sha_tmp, &i, 1);
+        secp256k1_sha256_finalize(&sha_tmp, buf);
         secp256k1_scalar_set_b32(&k[i], buf, NULL);
```
rage-proof, line 278: Has `buf` after this point the value of secnonce2 stored and should it be overwritten with zeros?

src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,55 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
...
-            secp256k1_sha256_write(&sha, extra_in[i], 32);
-        } else {
-            len = 0;
-            secp256k1_sha256_write(&sha, &len, 1);
-        }
-    }
-    secp256k1_sha256_finalize(&sha, seed);
+    secp256k1_sha256_write(&sha, rand, sizeof(rand));
+    secp256k1_nonce_function_musig_helper(&sha, 1, agg_pk32);
+    secp256k1_nonce_function_musig_helper(&sha, 1, msg32);
```
NicolasDorier, line 270: In the reference implementation msg doesn't have to be 32 bytes.

real-or-random:
@NicolasDorier Note that the spec has evolved further and this is indeed not yet reflected in this PR.

NicolasDorier:
just found out this PR is from May!

jonasnick:
Superseded by #211

PR #194, opened 2022-07-18
extrakeys: rename swap/swap64 to fix OpenBSD 7.1 compilation

jgriffiths:
OpenBSD defines swap64 in <endian.h>.

### DIFF
--- a/src/modules/extrakeys/hsort_impl.h
+++ b/src/modules/extrakeys/hsort_impl.h
@@ -23,7 +23,7 @@ static SECP256K1_INLINE size_t child2(size_t i) {
     return child1(i)+1;
 }
 
-static SECP256K1_INLINE void swap64(unsigned char *a, size_t i, size_t j, size_t stride) {
+static SECP256K1_INLINE void heap_swap64(unsigned char *a, size_t i, size_t j, size_t stride) {
     unsigned char tmp[64];
     VERIFY_CHECK(stride <= 64);
     memcpy(tmp, a + i*stride, stride);
@@ -31,12 +31,12 @@ static SECP256K1_INLINE void swap64(unsigned char *a, size_t i, size_t j, size_t
     memcpy(a + j*stride, tmp, stride);
 }
 
-static SECP256K1_INLINE void swap(unsigned char *a, size_t i, size_t j, size_t stride) {
+static SECP256K1_INLINE void heap_swap(unsigned char *a, size_t i, size_t j, size_t stride) {
     while (64 < stride) {
-        swap64(a + (stride - 64), i, j, 64);
+        heap_swap64(a + (stride - 64), i, j, 64);
         stride -= 64;
     }
-    swap64(a, i, j, stride);
+    heap_swap64(a, i, j, stride);
 }
 
 static SECP256K1_INLINE void heap_down(unsigned char *a, size_t i, size_t heap_size, size_t stride,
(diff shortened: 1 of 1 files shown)


jgriffiths:
This error was reported against wally by @grubles building c-lightning, the build failure log can be seen here: https://github.com/ElementsProject/libwally-core/issues/334.

> Do you use the library by including it instead of linking against it?

When building the normal libwallycore.a/.so file for linking into an application, no - we have a workaround for including the secp objects into the shared library (1). However for windows builds, cordova plugins, javascript and and the python packages we do, as there is no workaround to including the  secp256k1.c file in these cases (we don't control the build machinery).

We don't explicitly include <endian.h> and only BSD seems to be affected, so it seems to be transitively included by one of the python module build dependencies. In theory we could undefine the conflicting macro before including the source, and I am happy to ask grubles to test that. But we are not alone in this requirement, c-lightning separately builds its libhsm python module via secp256k1.c inclusion, and I believe any other python module will do as well (at least if they want to be manylinux/PyPI compatible). So it seems easier to namespace the calls here and avoid continuing fallout down the road.

real-or-random:
> > Do you use the library by including it instead of linking against it?
> 
> When building the normal libwallycore.a/.so file for linking into an application, no - we have a workaround for including the secp objects into the shared library (1). However for windows builds, cordova plugins, javascript and and the python packages we do, as there is no workaround to including the secp256k1.c file in these cases (we don't control the build machinery).
> 

Oh I see... I wasn't aware that people are actually doing this...


Just for my understanding: Wouldn't the normal way be to declare secp256k1-zkp to be a dependency and assume it's installed? What stops you from doing so? The missing releases/versioning? Or would this not work even if we had proper releases (because of all the special builds such as js and python).

real-or-random (approved):
ACK db648478c3cc600dbe42c5badb45a1e1fa217f13

jgriffiths:
> Just for my understanding: Wouldn't the normal way be to declare secp256k1-zkp to be a dependency and assume it's installed? What stops you from doing so? The missing releases/versioning? Or would this not work even if we had proper releases (because of all the special builds such as js and python).

A possibly overlong overview:

If you were building an executable then yes, you could reasonably expect that to work with both static and dynamic linking. We do support building against an externally specified libsecp, but only for builds without elements support, and only for the case where you are building wally to link with your app specifically - at which point, we disclaim responsibility for you massaging your build environment to make it work.

However this libsecp fork has a lot of experimental modules, while we don't use them all yet, a system-wide installed libsecp would need to have the ones we need enabled, and that would open us up to missing modules, interface changes and working around bugs in installed versions which isn't feasible for us to support given our available resources and the large number of platform and language targets. Currently that's c, c++, android native, python, java, swift, cordova, javascript, embedded, wasm and more if you count our forks with languages we haven't merged yet (go, rust, ...), in BTC only and BTC + Elements variants.

On top of that, we support different options for embedded dev (Jade, bitbox02, homebrew HW) to reduce memory usage (e.g. secp contexts), and we need to be able to CI multiple build types on the same machine (dozens, for the manylinux python builds, 5 for native android cross-compiles).  And we've had to patch in our own bug fixes in the past. So the only solution we've found is bundling secp-zkp as a submodule, and building as we currently do.

In the specific case of python, its build system expects c-extensions to be fully self contained source files which are built by the packaging system, resulting in a single shared library with all symbols resolved except libpython.so and some blessed target specific libs (which go back to 2010 and obviously don't include libsecp). We used to have more flexibility in part-building our python module outside of the python tools but as time goes on various approaches have been shut down until all we have left is the current one.

Each other platform we build secp via include for has a similar story.

jonasnick (commented):
ACK db648478c3cc600dbe42c5badb45a1e1fa217f13

real-or-random:
@jgriffiths I see, that makes sense... I kind of hoped that this is something we don't need to support but it seems I'm wrong. :D

jgriffiths:
@real-or-random Thanks!

PR #195, opened 2022-07-22
Add dedicated methods for verifying asset/value proofs

apoelstra:
- [x] implement value proof verification
- [ ] implement asset proof verification
- [ ] add more tests


include/secp256k1_rangeproof.h
```diff
@@ -286,6 +286,27 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_rangeproof_info(
   size_t plen
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
+/** Verify a rangeproof with a single-value range. Useful as a "proof of value"
+ *  of a Pedersen commitment. Such proofs can be created with `secp256k1_rangeproof_sign`
+ *  by passing an `exp` parameter of -1 and the target value as both `value` and `min_value`.
+ *  Returns 1: Proof was valid and proved the given value
```
jgriffiths, line 292: Is it also the case that `min_bits` must be 0 for the rangeproof, (as elements generates it) - should this be documented here?

include/secp256k1_surjectionproof.h
```diff
@@ -263,6 +263,24 @@ SECP256K1_API int secp256k1_surjectionproof_verify(
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(5);
 #endif
 
+/** Verify a single-input surjectionproof. Such proofs are sometimes useful to
+ *  prove in zero knowledge that a given commitment commits to a specific asset.
+ *  They can be verified with much less memory than generalr proofs.
```
jgriffiths, line 268: typo `generalr`

src/modules/rangeproof/main_impl.h
```diff
@@ -304,4 +304,98 @@ int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof
...
+        if (value != 0) {
+            return 0;
+        }
+        offset = 1;
+    } else {
+        int i;
+        uint64_t claimed = 0;
+        for (i = 0; i < 8; i++) {
+            claimed = (claimed << 8) | proof[1 + i];
+        }
```
jgriffiths, line 349: nit: you can loop here with `for (offset = 1; offset < 9; offset++)`, then get rid of the `int i` declaration, the `+1` in the indexing, and the assignment `offset = 9` below.

apoelstra: Thank you! The target here is very memory-constrained devices so any reduction in the number of variables is good.

include/secp256k1_rangeproof.h
apoelstra: I just checked, and interestingly no -- you can set `min_bits` to anything and it'll just be ignored.

src/modules/rangeproof/main_impl.h
jgriffiths: The compiler would probably make this change for you, I suspect. But you could also initialize offset to 1 where its declared, remove the assignment to 1 in the if branch above, and use `while (offset < 9) { ..; offset++; }` as your loop here.

You could probably reorder the stack vars to align them and use less stack space too.

include/secp256k1_rangeproof.h
apoelstra: Added a comment to this effect.

include/secp256k1_surjectionproof.h
apoelstra: Fixed.

src/modules/rangeproof/main_impl.h
apoelstra: Updated the loop but didn't reorder any variables (*that* I think even the most limited compiler would do)

jgriffiths:
utack 9eea0819bd08e1f98cfcb932728d6bbb5290922d

I don't have time at this stage to test this; wally uses the full proof validation code for blind asset/value proofs since we use that code already for blinding. Once PSBTv2 is in I should be able to switch over and make sure our tests still pass, may be a while though.

From an API perspective, it may be nice to have create_value variants for creating single value proofs that take the reduced set of parameters and call the full functions with the repeated/defaulted values (even if its not worth a custom implementation of the actual proof creation).

apoelstra:
Agreed, `create_` variants would be nice. I'll add those to this PR.

apoelstra:
I have added a `rangeproof_create_value` function to create a single-value rangeproof. I don't think there's much value in doing the same for surjection proofs -- to create a single-value surjection proof you just use the normal surjectionproof in the "obvious" way by providing a single input and setting `n_ephemeral_input_tags` to 1. No magic or unused parameters.

jgriffiths:
@apoelstra rangeproof was the painful case for me, so thanks for that, LGTM.

wally (following the Elements impl) only creates proofs with a maximum of 3 used inputs, so for generation even the reduced mode has 12 more unused stack elements than we need. However, we don't enable reduced mode so that we can verify proofs created by others since wally is available for use as a general purpose lib.  In an ideal world we'd be able to specify the number we use instead of having a fixed limit, and even better be able to generate with one limit and verify with another. I can see Jade/bitbox using 3/16 for create/verify for example assuming they can both verify on device.

jgriffiths:
> but not nearly as bad as rangeproof_sign

Note for `secp256k1_rangeproof_sign`, the final `memset(prep, 0, 4096);` is redundant, having been zeroed above. If the compiler doesn't eliminate that as a dead store that would keep `prep` alive for the whole function which would prevent some other arrays like `pubs` from reusing its space.

You could use a narrower unsigned type for `rsizes` and `secidx` to shave some stack space in the general case, otherwise I suspect some judicious use of `restrict` would be needed to allow the compiler to optimally overlay the used stack items.

apoelstra:
@jgriffiths i actually have a massive branch (which will need to be rebased now since i reused bits and pieces for this) whose goal is to dramatically reduce the stack space of the rangeproof code. Unfortunately it required a near rewrite of all the logic, pulling the borromean sig stuff out of its own functions and inlining it in the rangeproof/surjectionproof code.

Essentially the way it works is by directly using the `proof` buffer to store data so that we don't need all these arrays of intermediate values.

I'm quite proud of it but I don't think we have the review bandwidth to do anything with it :/ and I'm not sure that I finished it.

apoelstra:
https://github.com/ElementsProject/secp256k1-zkp/pull/160 is the start of it

jgriffiths:
> I don't think we have the review bandwidth to do anything with it

The eternal problem lol.

At this stage things are fine for us; its only if the hardware wallets have issues that we may need to look at lowering resource usage. That time is creeping forward for Jade at least, since the qrcode handling has been improved to the point where we should be able to take psbts from the camera to sign. But until we are ready to resource that we can't commit anyone to review or test either.

In the meantime though I'd remove the redundant memset just because its, well, redundant :)

ChristopherA:
> since the qrcode handling has been improved to the point where we should be able to take psbts from the camera to sign. 

Have you looked at how we do animated QR PSBTs? A number of both hardware & software bitcoin wallets and services now support it. It is quite mature, has been ported to multiple languages, and has other advantages.  

https://github.com/BlockchainCommons/Research/blob/master/papers/bcr-2020-006-urtypes.md#partially-signed-bitcoin-transaction-psbt-crypto-psbt

jgriffiths:
Hi @ChristopherA - yes, I believe we are targeting that format, however this PR and repo aren't the place to discuss it. Please follow https://github.com/Blockstream/Jade where the PR adding support will land after internal review, or raise an issue there if you have questions/comments, thanks.

real-or-random:
> In the meantime though I'd remove the redundant memset just because its, well, redundant :)

Please leave it the memset in it's there for a reason. It's to remove secrets from the stack. Now of course everyone knows that this won't work because the compiler will remove dead stores. But for some reasons we have the plain memsets in the code because this is how the code evolved. There's [an old PR upstream](https://github.com/bitcoin-core/secp256k1/pull/636) that replaces all the memsets by proper methods to overwrite secret data. (Yeah I should rebase this...) Once this lands in upstream, we should do the same changes here.

So in sense, the dead memsets in the code base currently act as a marker for where we should do proper overwriting in the future, and so we should keep them.

real-or-random:
I don't know how other reviewers feel but I think I'd need some refresher on how the surjection proofs work before I'm a position to review this. Or maybe @apoelstra can walk one of us through the changes.

apoelstra:
@real-or-random the surjection proofs work by taking a single output commitment, a variable number of input commitments, and then doing a ring signature over (output - input_i) for all i. The premise is that the user will only be able to construct such a ring signature if the output is merely a reblinding of one of the inputs.

Everything else in this code is just mechanical stuff to make the commitment structure and proof serialization match the deployed surjection proofs in Elements.

apoelstra:
Oh -- and to add, a "ring signature" is just a Schnorr signature in both cases, since we're proving a single value so there is only one key to check.

apoelstra:
@real-or-random I'm going to make another PR on top of this which pulls the asset ID/entropy logic from Elements Core into secp-zkp, so the sooner we can merge this the simpler my life will be.

apoelstra:
Actually, scrap that. On further inspection I think the API separation is already fine.

real-or-random:
Ok I'll try to have a look tomorrow. :)

src/modules/rangeproof/tests_impl.h
```diff
@@ -547,8 +547,95 @@ static void test_rangeproof(void) {
...
+        &commit,
+        proof, plen,
+        NULL, 0,
+        secp256k1_generator_h
+    ));
+    CHECK(val_out == val);
+    CHECK(min_val_out == val);
+    CHECK(max_val_out == val);
+    CHECK(m_len_out == 0);
+    CHECK(memcmp(blind, blind_out, 32) == 0);
```
real-or-random, line 678: `secp256k1_memcmp_var` (multiple times)

src/modules/surjection/main_impl.h
```diff
@@ -393,4 +393,74 @@ int secp256k1_surjectionproof_verify(const secp256k1_context* ctx, const secp256
...
+
+    secp256k1_generator_load(&inputp, input_tag);
+    secp256k1_generator_load(&outputp, output_tag);
+    secp256k1_ge_neg(&inputp, &inputp);
+    secp256k1_gej_set_ge(&xj, &inputp);
+    secp256k1_gej_add_ge(&xj, &xj, &outputp);
+
+    /* Now we just have a Schnorr signature in (e, s) form. The verification
+     * equation is e == H(sG - eX || proof params), where X is the difference
+     * between the output and input. */
```
real-or-random, line 431: Would it make sense to extract this code into a single function used by rangeproofs and surjectionproofs?

src/modules/surjection/main_impl.h
```diff
@@ -393,4 +393,74 @@ int secp256k1_surjectionproof_verify(const secp256k1_context* ctx, const secp256
...
+    /* Validate and decode surjectionproof data */
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(proof != NULL);
+    ARG_CHECK(input_tag != NULL);
+    ARG_CHECK(output_tag != NULL);
+#ifdef VERIFY
+    CHECK(proof->initialized == 1);
+#endif
+
+    if (proof->n_inputs != 1 || proof->used_inputs[0] != 1) {
```
real-or-random, line 419: Do we need to check the other bytes in `proof->used_inputs`?

src/modules/surjection/main_impl.h
```diff
@@ -393,4 +393,74 @@ int secp256k1_surjectionproof_verify(const secp256k1_context* ctx, const secp256
...
+    secp256k1_gej_set_ge(&xj, &inputp);
+    secp256k1_gej_add_ge(&xj, &xj, &outputp);
+
+    /* Now we just have a Schnorr signature in (e, s) form. The verification
+     * equation is e == H(sG - eX || proof params), where X is the difference
+     * between the output and input. */
+
+    /* 1. Compute slow/overwrought commitment to proof params */
+    secp256k1_surjection_genmessage(pp_comm, input_tag, 1, output_tag);
+    /* (past this point the code is identical to rangeproof_verify_value) */
```
real-or-random, line 435: Would it make sense to extract this code into a single function used by rangeproofs and surjectionproofs?

src/modules/rangeproof/tests_impl.h
```diff
@@ -225,6 +225,53 @@ static void test_rangeproof_api(const secp256k1_context *none, const secp256k1_c
         CHECK(secp256k1_rangeproof_rewind(both, blind_out, &value_out, NULL, 0, commit.data, &min_value, &max_value, &commit, proof, len, NULL, 0, NULL) == 0);
         CHECK(*ecount == 29);
     }
+
+    {
+        *ecount = 0;
+        len = sizeof(proof);
+        CHECK(secp256k1_rangeproof_create_value(none, proof, &len, val, blind, &commit, secp256k1_generator_h) == 1);
+        CHECK(secp256k1_rangeproof_create_value(vrfy, proof, &len, val, blind, &commit, secp256k1_generator_h) == 1);
+        CHECK(secp256k1_rangeproof_create_value(sign, proof, &len, val, blind, &commit, secp256k1_generator_h) == 1);
```
real-or-random, line 234: maybe just use "none" in new code, the other contexts anyway don't exist any longer

src/modules/rangeproof/main_impl.h
```diff
@@ -397,4 +397,105 @@ int secp256k1_rangeproof_verify_value(const secp256k1_context* ctx, const unsign
...
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(secp256k1_ecmult_gen_context_is_built(&ctx->ecmult_gen_ctx));
+    ARG_CHECK(proof != NULL);
+    ARG_CHECK(plen != NULL);
+    ARG_CHECK(blind != NULL);
+    ARG_CHECK(commit != NULL);
+    ARG_CHECK(gen != NULL);
+
+    if (*plen < 73 || (value == 0 && *plen < 65)) {
+        return 0;
```
real-or-random, line 422:
```
if (*plen < 65 || (value != 0 && *plen < 73)) {
```

or more readable (I think):
```
if (*plen < (value == 0) ? 65 : 73) {
```

(This implies that we should have also a test for `value == 0`.)

real-or-random (commented):
Just for my the understanding: The benefit here is simply efficiency, right? 

In general, I think we could add some more tests that test consistency with the ordinary functions, maybe also with bitflips (because we have many and "bitmaps"/"flags").

src/modules/rangeproof/tests_impl.h
```diff
@@ -547,8 +547,95 @@ static void test_rangeproof(void) {
...
+        &min_val_out, &max_val_out,
+        &commit,
+        proof, plen,
+        NULL, 0,
+        secp256k1_generator_h
+    ) == 1);
+    CHECK(min_val_out == val);
+    CHECK(max_val_out == val);
+
+    memset(message_out, 0, sizeof(message_out));
```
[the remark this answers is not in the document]
apoelstra: I don't think we make any promises about what `rewind` will do to the outpointer. I would expect it to overwrite with all zeros but honestly I'm not sure. This memset is probably just there to silence gcc warnings.

src/modules/rangeproof/main_impl.h
```diff
@@ -304,4 +304,97 @@ int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof
...
+    /* Now we just have a Schnorr signature in (e, s) form. The verification
+     * equation is e == H(sG - eX || proof params) */
+
+    /* 1. Compute slow/overwrought commitment to proof params */
+    secp256k1_sha256_initialize(&sha2);
+    secp256k1_rangeproof_serialize_point(tmpch, &commitp);
+    secp256k1_sha256_write(&sha2, tmpch, 33);
+    secp256k1_rangeproof_serialize_point(tmpch, &genp);
+    secp256k1_sha256_write(&sha2, tmpch, 33);
+    secp256k1_sha256_write(&sha2, proof, offset); /* lol we commit to one extra byte here */
```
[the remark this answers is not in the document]
apoelstra: I'll drop the comment. I think I miscounted -- I believed that we were actually commiting to one byte *past* proof[0] and the minimum value.

src/modules/rangeproof/main_impl.h
```diff
@@ -304,4 +304,97 @@ int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof
...
+    /* 0x80 must be unset for any rangeproof; 0x40 indicates "has nonzero range"
+     * so must also be unset for single-value proofs */
+    if ((proof[0] & 0xC0) != 0x00) {
+        return 0;
+    }
+
+    secp256k1_pedersen_commitment_load(&commitp, commit);
+    secp256k1_generator_load(&genp, gen);
+    /* Verify that value in the header is what we expect; 0x20 is "has nonzero min-value" */
+    if ((proof[0] & 0x20) == 0x00) {
```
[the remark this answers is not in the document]
apoelstra: No we don't need to check the other bits. You are welcome to set them to crap even in the normal rangeproof verification code.

Using `getheader_impl` would make the code much much harder to read for somebody trying to implement these functions independently, and writing a reference implementation was one goal of this PR.

src/modules/surjection/main_impl.h
apoelstra: It differs in just enough places that I don't think it's practical.

src/modules/surjection/main_impl.h
apoelstra: No, because if `proof->n_inputs` is 1 then we know only one bit of `proof->used_inputs` is used.

src/modules/surjection/main_impl.h
apoelstra: Oh, I see, between rangeproofs and surjection proofs ... yes, that might make sense. I thought you meant between proving and verifying.

src/modules/rangeproof/main_impl.h
apoelstra: BTW there is an explicit test for the size being correct when `value == 0`. You can search for the comment `DIfferent proof sizes are unfortunate`.

apoelstra:
@real-or-random I have pushed fixup commits which address all your nits except the code-sharing ones. If you'd like I can refactor the code, but I don't think the savings would be super big and it'd undermine the "make these functions as self-contained as possible" goal of writing this.

src/modules/rangeproof/main_impl.h
```diff
@@ -304,4 +304,198 @@ int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof
...
+    secp256k1_eckey_pubkey_serialize(&rp, tmpch, &sz, 1);
+
+    /* 2. Compute e = H(R || proof params) */
+    secp256k1_sha256_initialize(&sha2);
+    secp256k1_sha256_write(&sha2, tmpch, sz);
+    secp256k1_sha256_write(&sha2, pp_comm, sizeof(pp_comm));
+    secp256k1_sha256_finalize(&sha2, tmpch);
+
+    /* 3. Check computed e against original e */
+    return !memcmp(tmpch, &proof[offset], 32);
```
real-or-random, line 397: There are more `memcmp`s, and those could actually matter because they're not in the real code, not in the tests.

src/modules/rangeproof/main_impl.h
```diff
@@ -304,4 +304,198 @@ int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof
...
+
+    /* Now we have to make a Schnorr signature in (e, s) form. */
+
+    /* 1. Compute slow/overwrought commitment to proof params */
+    secp256k1_sha256_initialize(&sha2);
+    secp256k1_rangeproof_serialize_point(tmpch, &commitp);
+    secp256k1_sha256_write(&sha2, tmpch, 33);
+    secp256k1_rangeproof_serialize_point(tmpch, &genp);
+    secp256k1_sha256_write(&sha2, tmpch, 33);
+    secp256k1_sha256_write(&sha2, proof, offset); /* lol we commit to one extra byte here */
```
real-or-random, line 453: Here's a duplicate of the removed comment.

src/modules/surjection/main_impl.h
real-or-random: Oh, I said the same thing twice... Let's track this in the other review comment.

src/modules/surjection/main_impl.h
real-or-random: Yeah, I think it would make sense (e.g., it would have prevented the issue with the duplicate comment, and it will make you change `memcmp` only once. ;)) But your goal of making them self-consistent is also good. Whatever you prefer.

src/modules/rangeproof/tests_impl.h
```diff
@@ -696,6 +944,11 @@ void test_pedersen_commitment_fixed_vector(void) {
 void run_rangeproof_tests(void) {
     int i;
     test_api();
+
+    test_single_value_proof(0);
+    test_single_value_proof(12345678);
+    test_single_value_proof(UINT64_MAX);
```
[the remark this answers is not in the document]
real-or-random: maybe add a small loop?

include/secp256k1_rangeproof.h
real-or-random: You gave a ":+1:" but I think this hasn't been fixed.

src/modules/rangeproof/main_impl.h
real-or-random: Ah what I meant here is: If we had a (unit) test that calls this function with  `value == 0` and `65 <= size < 73`, it would have caught this bug.

real-or-random:
> Just for my the understanding: The benefit here is simply efficiency, right?
> 
> In general, I think we could add some more tests that test consistency with the ordinary functions, maybe also with bitflips (because we have many and "bitmaps"/"flags").

You didn't answer these. I guess then an another goal is to have a simple readable "reference-like" implementation?

And for the tests: I still think that consistency tests could be useful. (Not sure about the bitflips anymore, I think you've convinced me that we know what we're doing with the header bits and "used input" bitmap.)

apoelstra:
* Regarding the purpose of these functions, there are two goals: to be a reference implementation for somebody who is working on implementing them on a Ledger with bare crypto primitives; and efficiency.
* Regarding the comment, revisiting this I think the "how to create these" comment does make sense on the `verify` function and less so on the `create` function.

I'll add some more consistency tests, and a test for the size thing.

src/modules/rangeproof/main_impl.h
apoelstra: Lol, there's an even worse bug here -- we never actually set `plen`. It looks like I did basically no tests on `create_value`.

apoelstra:
One last thing -- @real-or-random should I rename these functions from `create_value` and `verify_value` to `create_single` and `verify_single`?

real-or-random:
> One last thing -- @real-or-random should I rename these functions from `create_value` and `verify_value` to `create_single` and `verify_single`?

Ah yeah, I haven't really thought about the names. Now that you mention it, I think `create_single` and `verify_single` are slightly better. I'd suggest `create_exact` and `verify_exact`. This is even more precise if you ask me.

apoelstra:
@real-or-random I think one of the API tests tripped the "providing `plen` between 65 and 73 doesn't work" bug, and I was asserting that it failed (copy/paste error because almost all the API tests end in `== 0`..). So when I fixed it it broke the API test.

I'll start running `git rebase -x 'make && ./tests 1'` before pushing from now on.

apoelstra:
...ok, I might not be able to fix this today, but there's another bug in `create_value` where I use `&proof[offset]` before this value is filled in ... the fact that the tests were passing before seems to be a very surprising coincidence related to my calling `create_value` several times in a row before calling `verify_value`. There is a good chance I'll have to rewrite much of the function.

apoelstra:
Ok @real-or-random one more question ... should you be able to "rewind" single-value proofs? I'm going to vote no, it will make my life much harder (I will need to recreate the insane hmac-sha2-based RNG and run a ton of crap through it) and I don't see any real value to it.

apoelstra:
Ok, this one should work :sweat_smile:

real-or-random:
> Ok @real-or-random one more question ... should you be able to "rewind" single-value proofs? I'm going to vote no, it will make my life much harder (I will need to recreate the insane hmac-sha2-based RNG and run a ton of crap through it) and I don't see any real value to it.

Yep, if you don't think we need tell. (I don't know, I don't use these methods :D).

apoelstra:
Rebase/squash only. Will add one more fixup which changes how `k` is computed.

apoelstra:
Ok, rebased and added one more commit with @real-or-random's suggestion to compute `k` differently. I can squash that commit if you'd like though I don't really think it's necessary.

real-or-random (approved):
ACK a20c1e950e258d4e5b478f4f20baee589240ee02

apoelstra:
Ok to merge?

PR #196, opened 2022-07-26
surjectionproof: fail to generate proofs when an input equals the output

apoelstra:
If any ephemeral input tag equals the ephemeral output tag (i.e. an input asset is exactly equal to the output asset), verification will fail due to an unexpected interaction between our surjectionproof logic and the underlying borromean ring siganture logic. However, our generation code still allows creating proofs like this, "succeeding" in creating bad proofs.

Since we cannot fix the verification side without hardforking Liquid, fix the generation side to fail in this situation.

### DIFF
--- a/src/modules/surjection/main_impl.h
+++ b/src/modules/surjection/main_impl.h
@@ -312,7 +312,7 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
      * it is deployed on Liquid and cannot be fixed without a hardfork. We should review
      * this at the same time that we relax the max-256-inputs rule. */
     for (i = 0; i < n_ephemeral_input_tags; i++) {
-        if (memcmp(ephemeral_input_tags[i].data, ephemeral_output_tag->data, sizeof(ephemeral_output_tag->data)) == 0) {
+        if (secp256k1_memcmp_var(ephemeral_input_tags[i].data, ephemeral_output_tag->data, sizeof(ephemeral_output_tag->data)) == 0) {
             return 0;
         }
     }
--- a/src/modules/surjection/main_impl.h
+++ b/src/modules/surjection/main_impl.h
@@ -307,10 +307,14 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     if (overflow) {
         return 0;
     }
-    /* The only time the input may equal the output is if neither one was blinded in the first place,
-     * i.e. both blinding keys are zero. Otherwise this is a privacy leak. */
-    if (secp256k1_scalar_eq(&tmps, &blinding_key) && !secp256k1_scalar_is_zero(&blinding_key)) {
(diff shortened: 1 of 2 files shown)


jgriffiths:
utack bf18ff5a8c6295cb7db6e2989aefd6a78df7720f LGTM

src/modules/surjection/main_impl.h
```diff
@@ -307,10 +307,14 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
...
-    /* The only time the input may equal the output is if neither one was blinded in the first place,
-     * i.e. both blinding keys are zero. Otherwise this is a privacy leak. */
-    if (secp256k1_scalar_eq(&tmps, &blinding_key) && !secp256k1_scalar_is_zero(&blinding_key)) {
-        return 0;
+    /* If any input tag is equal to an output tag, verification will fail, because our ring
+     * signature logic would receive a zero-key, which is illegal. This is unfortunate but
+     * it is deployed on Liquid and cannot be fixed without a hardfork. We should review
+     * this at the same time that we relax the max-256-inputs rule. */
+    for (i = 0; i < n_ephemeral_input_tags; i++) {
+        if (memcmp(ephemeral_input_tags[i].data, ephemeral_output_tag->data, sizeof(ephemeral_output_tag->data)) == 0) {
```
real-or-random, line 315:
```suggestion
        if (secp256k1_memcmp_var(ephemeral_input_tags[i].data, ephemeral_output_tag->data, sizeof(ephemeral_output_tag->data)) == 0) {
```

src/modules/surjection/main_impl.h
```diff
@@ -307,10 +307,14 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
...
         return 0;
     }
-    /* The only time the input may equal the output is if neither one was blinded in the first place,
-     * i.e. both blinding keys are zero. Otherwise this is a privacy leak. */
-    if (secp256k1_scalar_eq(&tmps, &blinding_key) && !secp256k1_scalar_is_zero(&blinding_key)) {
-        return 0;
+    /* If any input tag is equal to an output tag, verification will fail, because our ring
+     * signature logic would receive a zero-key, which is illegal. This is unfortunate but
+     * it is deployed on Liquid and cannot be fixed without a hardfork. We should review
+     * this at the same time that we relax the max-256-inputs rule. */
```
jonasnick, line 313: I doubt that this note will have an effect. Tthere's perhaps a better place to collect backwards-incompatible changes under consideration and if not we may want to create one (a la "hardfork wishlist").

jonasnick (commented):
ACK 4ff6e4274d49cb95ab246b599b274104baf83f9f

real-or-random (approved):
utACK d1175d265d514bd0c22faaf262d7df362f33af89

PR #198, opened 2022-08-05
rangeproof: add a test for all-zero blinding factors

apoelstra:
I was curious about under what conditions you can create a rangeproof on an "unblinded" commitment which has a zero blinding factor. Apparently the answer is "when you are proving at least 3-bits". In this case rewinding words and you can encode 32 bytes of data. (In fact I believe you can encode up to 128 but I haven't tested that.)

### DIFF
--- a/src/modules/rangeproof/tests_impl.h
+++ b/src/modules/rangeproof/tests_impl.h
@@ -196,7 +196,7 @@ static void test_rangeproof_api(const secp256k1_context *none, const secp256k1_c
         CHECK(max_value >= val);
         CHECK(value_out == val);
         CHECK(message_len == sizeof(message_out));
-        CHECK(memcmp(message, message_out, sizeof(message_out)) == 0);
+        CHECK(secp256k1_memcmp_var(message, message_out, sizeof(message_out)) == 0);
 
         CHECK(secp256k1_rangeproof_rewind(both, NULL, &value_out, message_out, &message_len, commit.data, &min_value, &max_value, &commit, proof, len, ext_commit, ext_commit_len, secp256k1_generator_h) != 0);
         CHECK(*ecount == 21);  /* blindout may be NULL */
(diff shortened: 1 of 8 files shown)


real-or-random:
> Apparently the answer is "when you are proving at least 3-bits".

Hm, interesting... Do we also want to document this somewhere where it's user-visible?

apoelstra:
Yeah, I think so. I was going back-and-forth but I think this is likely to be generally useful -- sometimes when balancing CT transactions you wind up being forced to set a blinding factor to 0 and it's good to know that you can still create a rangeproof so that the transaction will be accepted.

real-or-random:
Valgrind failure: https://cirrus-ci.com/task/6618538872930304?logs=cat_tests_log#L11

apoelstra:
Oops -- I forgot to initialize my `msg` array before encoding it.

src/modules/rangeproof/tests_impl.h
```diff
@@ -547,6 +547,58 @@ static void test_rangeproof(void) {
...
+
+        len = 1000;
+        secp256k1_testrand256(msg);
+        secp256k1_testrand256(&msg[32]);
+        secp256k1_testrand256(&msg[64]);
+        secp256k1_testrand256(&msg[96]);
+        CHECK(secp256k1_rangeproof_sign(ctx, proof, &len, v, &commit, blind, commit.data, 0, 3, v, msg, sizeof(msg), NULL, 0, secp256k1_generator_h));
+        CHECK(secp256k1_rangeproof_rewind(ctx, blind_out, &value_out, msg_out, &msg_len, commit.data, &minv, &maxv, &commit, proof, len, NULL, 0, secp256k1_generator_h) != 0);
+        CHECK(memcmp(blind, blind_out, sizeof(blind)) == 0);
+        CHECK(memcmp(msg, msg_out, sizeof(msg)) == 0);
```
real-or-random, line 595: we prefer `secp256k1_memcmp_var` here (but feel free to ignore me -- I feel stupid pointing this out in every single PR... we should either check in CI or not bother...)

real-or-random (commented):
utACK but maybe squash this. The test changes are now spread over two commits. (You could also cleanly separate them but having this in a single commit is totally fine or even better, I think).

src/modules/rangeproof/tests_impl.h
apoelstra: Well, with enough constant reminders eventually I'll learn :)

apoelstra:
Squashed the two commits in this PR; added another which fixes all the `memcmp` instances in the secp-zkp modules.

real-or-random (approved):
utACK 5a40f3d99bbd879391a3fb3c038a6d49ec01bc03

real-or-random:
tACK 5a40f3d99bbd879391a3fb3c038a6d49ec01bc03

PR #199, opened 2022-08-09
surjectionproof: make sure that n_used_pubkeys > 0 in generate

jonasnick:
If the proof was generated with surjectionproof_initialize (as mandated by the
API docs), then n_used_pubkeys can never be 0. Without this commit, compilers
will (rightfully) warn that borromean_s[ring_input_index] is not initialized in
surjectionproof_generate. Therefore, this commit makes sure that n_used_pubkeys
is greater than 0 which ensures that the array is initialized at
ring_input_index.

### DIFF
--- a/src/modules/surjection/main_impl.h
+++ b/src/modules/surjection/main_impl.h
@@ -298,6 +298,10 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     CHECK(proof->initialized == 1);
 #endif
 
+    n_used_pubkeys = secp256k1_surjectionproof_n_used_inputs(ctx, proof);
+    /* This must be true if the proof was created with surjectionproof_initialize */
+    ARG_CHECK(n_used_pubkeys > 0);
+
     /* Compute secret key */
     secp256k1_scalar_set_b32(&tmps, input_blinding_key, &overflow);
     if (overflow) {
@@ -321,7 +325,7 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
 
     /* Compute public keys */
     n_total_pubkeys = secp256k1_surjectionproof_n_total_inputs(ctx, proof);
-    n_used_pubkeys = secp256k1_surjectionproof_n_used_inputs(ctx, proof);
+
     if (n_used_pubkeys > n_total_pubkeys || n_total_pubkeys != n_ephemeral_input_tags) {
         return 0;
     }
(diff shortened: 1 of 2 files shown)


apoelstra (approved):
ACK 9e58040be038670d62771e5a0dd4b79c932fffe0

apoelstra:
Thanks for resetting `ecount` to 0 rather than incrementing all the values and leaving it to the next person to do it again :)

src/modules/surjection/main_impl.h
```diff
@@ -322,6 +322,10 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     /* Compute public keys */
     n_total_pubkeys = secp256k1_surjectionproof_n_total_inputs(ctx, proof);
     n_used_pubkeys = secp256k1_surjectionproof_n_used_inputs(ctx, proof);
+
+    /* This must be true if the proof was created with surjectionproof_initialize */
+    ARG_CHECK(n_used_pubkeys > 0);
+
     if (n_used_pubkeys > n_total_pubkeys || n_total_pubkeys != n_ephemeral_input_tags) {
         return 0;
     }
```
real-or-random, line 331: Is it inconsistent to return different errors here?

Should the ARG_CHECK be moved up to the other ARG_CHECKs?

jonasnick: `ARG_CHECK` just seems like the right thing to do (since it's an API violation) and I didn't want to touch existing code (which doesn't have tests afaics).

It can't be moved up because we only set n_used_pubkeys in the line right above this new ARG_CHECK.

real-or-random: > `ARG_CHECK` just seems like the right thing to do (since it's an API violation) and I didn't want to touch existing code (which doesn't have tests afaics).

Hm yeah, I assumed this may contradict some some unwritten rule that ARG_CHECKs checks things that should be enforceable by normal type system. But in this case, rust-secp256k1-zkp anyway always calls `initialize` and `create` in a row. (This makes me wonder why these are separate functions in C.)

real-or-random: > It can't be moved up because we only set n_used_pubkeys in the line right above this new ARG_CHECK.

Ok, I meant it could be moved together together with the other code, because it looked wrong on first glance to have `ARG_CHECK`s, then `return 0` a few times , and then an `ARG_CHECK` again. 

But ok, it's not wrong. I still think it will be cleaner to have all the `ARG_CHECK` first.

apoelstra: We should look at merging `initialize` and `create`. I think this was my doing and I don't remember what my motivation was.

jonasnick: @real-or-random 

> Ok, I meant it could be moved together together with the other code,

Ah sorry, got it. Didn't realize I can just move the other code to the top. Updated the commit.

apoelstra (approved):
utACK 5ac8fb035e8f0ad012d9f54b49e77b5d1f5e713a

PR #200, opened 2022-08-09
build: automatically enable module dependencies

jonasnick:


configure.ac
```diff
@@ -385,6 +385,10 @@ SECP_CFLAGS="$SECP_CFLAGS $WERROR_CFLAGS"
 ### Handle module options
 ###
 
+# Besides testing whether modules are enabled, the following code also enables
+# module dependencies. The order of the tests matters: the dependency must be
+# tested first.
+
```
real-or-random, line 391: nit: is the comment worth adding to upstream?

real-or-random (commented):
Concept ACK

Though what I don't like too much about this approach that it silently enables options, even if explicitly disabled, e.g., in 
```
./configure --enable-experimental --enable-module-surjectionproof --disable-module-rangeproof
```
It may be better to add a notice "Automatically enabling xy" then. 
But yeah upstream does the same, so if anything, we should fix it there. (I'm fine either way)

---

The code further below that checks the errors can (should?) also be removed then.
 
---

Not this PR: 
The option listing in the end is incomplete. It lacks some modules, e.g., whitelist. Those should be added, and when we do this, we could also remove the (then redundant) list from the experimental warning. We should just keep the warning, but maybe move it further down, at the moment it's interleaved with some other configure output... I think this PR would be a good opportunity to fix this but feel free to ignore, we can also do it in a separate PR.

jonasnick:
@real-or-random 

> Though what I don't like too much about this approach that it silently enables options, even if explicitly disabled

Oh, I missed that. I don't like that either. And yeah, if we keep it that way we can remove the error checking.

jonasnick:
I don't see an straightforward way to error out if dependencies are explicitly disabled. If we fix the configure output with respect to the enabled modules maybe that's not so much of a problem.

configure.ac
jonasnick: no imo, as there are comments already (what I removed here)

jonasnick:
I removed the error checking and added two commits that improve the output.

real-or-random:
> I don't see an straightforward way to error out if dependencies are explicitly disabled. If we fix the configure output with respect to the enabled modules maybe that's not so much of a problem.

Agreed, and PR is an improvement in the end, so we shouldn't hold it up.

real-or-random (approved):
utACK 9e58040be038670d62771e5a0dd4b79c932fffe0

jonasnick:
@real-or-random  this seems to be a commit id belonging to a different PR :)

real-or-random (approved):
utACK 171b294a1c7a736c1b93fa194e3af90b625259fa

PR #202, opened 2022-08-14
shut up spurious gcc warning

apoelstra:
gcc 12.1.1 has a new warning:
```
In file included from src/secp256k1.c:846,
                 from src/tests_exhaustive.c:20:
src/modules/surjection/main_impl.h: In function ‘secp256k1_surjectionproof_generate’:
src/modules/surjection/main_impl.h:343:24: warning: ‘borromean_s’ may be used uninitialized [-Wmaybe-uninitialized]
  343 |     nonce = borromean_s[ring_input_index];
      |             ~~~~~~~~~~~^~~~~~~~~~~~~~~~~~
```

The warning is spurious but add some code to shut it up anyway.


jonasnick:
This is also fixed by  #199 (which is better I think).

PR #204, opened 2022-09-11
rangeproof: add a "net blinding factor" API for Elements

apoelstra:
Our original API for Confidential Assets transaction balancing was the single function `secp256k1_pedersen_blind_generator_blind_sum` which attempts to take a complete list of vbfs and abfs and modifies a single abf at the end. However this API has a number of shortcomings:

   * it is really confusing
   * it assumes that the user has all the abfs and vbfs in convenient arrays, requiring marshalling on the C++ side
   * it does not support partial computations, as are needed by PSET
   * there is no easy/sensible way to extend this API to allow more interesting of transaction balancing (e.g. by blinding only an asset, leaving the value explicit)

The hope is that by exposing the arithmetic at a more fine-grained level, these issues will be fixed. These methods can be abused to do arithmetic on arbitrary scalars, but this is already possible (in an ugly manner) by using secp256k1_seckey_tweak_add and explicit 0-checks.


apoelstra:
I may add more commits to this but I think it's ready for review as-is.

PR #205, opened 2022-10-27
Bulletproofs++: Norm argument

sanket1729:
- Implements the norm argument for bulletproofs++. Companion code can be found at https://github.com/sanket1729/rust-bulletproofs-pp. 

Future PRs will include the range proof reciprocal argument. Based on #108


sanket1729:
This is ready for review. Supercedes #108 .

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -0,0 +1,51 @@
...
+ * Assumes both points' coordinates have been normalized.
+ */
+static void secp256k1_bulletproofs_serialize_points(unsigned char *output, const secp256k1_ge *lpt, const secp256k1_ge *rpt) {
+    output[0] = (secp256k1_fe_is_odd(&lpt->y) << 1) + secp256k1_fe_is_odd(&rpt->y);
+    secp256k1_fe_get_b32(&output[1], &lpt->x);
+    secp256k1_fe_get_b32(&output[33], &rpt->x);
+}
+
+/* Initializes SHA256 with fixed midstate. This midstate was computed by applying
+ * SHA256 to SHA256("Bulletproofs/commitment")||SHA256("Bulletproofs/commitment"). */
```
apoelstra, line 24: In e9491da1ae699531ba9f9d01c185732ee5ad11e8:

I kinda think we should change this value, since I'm pretty sure the grin people are using this old code in their consensus logic. (I dunno if grin is even alive right now, but still good to avoid collisions if we can.)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -77,4 +77,66 @@ static void secp256k1_bulletproofs_powers_of_r(secp256k1_scalar *powers, const s
...
+    const secp256k1_scalar* q
+) {
+    secp256k1_scalar v, l_c;
+    /* First n_vec_len generators are Gs, rest are Hs*/
+    if (g_vec->n != (n_vec_len + l_vec_len) || l_vec_len != c_vec_len) {
+        return 0;
+    }
+
+    if (!secp256k1_check_power_of_two(n_vec_len) ||  !secp256k1_check_power_of_two(c_vec_len)) {
+        return 0;
```
apoelstra, line 120: In e85e3d10729bd5cdd133334adb0f9b29ec946209:

I kinda think we should `CHECK` these rather than returning 0, since this is internal-only code. Or we could do both.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -77,4 +77,66 @@ static void secp256k1_bulletproofs_powers_of_r(secp256k1_scalar *powers, const s
...
+    const secp256k1_scalar* l_vec,
+    size_t l_vec_len,
+    const secp256k1_scalar* c_vec,
+    size_t c_vec_len,
+    const secp256k1_scalar* q
+) {
+    secp256k1_scalar v, l_c;
+    /* First n_vec_len generators are Gs, rest are Hs*/
+    if (g_vec->n != (n_vec_len + l_vec_len) || l_vec_len != c_vec_len) {
+        return 0;
```
apoelstra, line 116: In e85e3d10729bd5cdd133334adb0f9b29ec946209:

Ditto here -- let's just assert.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
apoelstra: We also may want to add a comment here indicating that the code could be made to work with non-powers-of-2 ... we just don't support this now. --liam

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -139,4 +139,251 @@ static int secp256k1_bulletproofs_commit(
...
+    r_cb_data.l1 = ls;
+    r_cb_data.G_GENS_LEN = G_GENS_LEN;
+    secp256k1_scalar_sqr(&q, &r);
+
+
+    secp256k1_sha256_initialize(&sha256);
+    secp256k1_sha256_write(&sha256, transcript_hash32, 32);
+    secp256k1_fe_normalize_var(&comm.x);
+    secp256k1_fe_normalize_var(&comm.y);
+    ser_commit[0] = 0x02 | secp256k1_fe_is_odd(&comm.y);
```
apoelstra, line 282: In a726d5e62ef188f0549ce2ba1a6b59d01f897255:

I'm curious why you take pains here to encode this as a SECG point, but don't use the existing point serialization methods. I have some old BP code that does stuff like this but only if:
* I'm not able to modify the point I'm serializing (but you can, since you're normalizing its coordinates in place)
* I need to do the serialization in constant time (but we don't need to, since this is a public point .... and in any case you're not, because you're calling `normalize_var`)

apoelstra: Also Liam points out that this LOC is not present in your other branch `bpp_final`....is this PR out of date?

apoelstra: Ah, this code is removed in aec736da964be3db5dd137b824ed0b4cb612db51 (commit 15 of #207). Ok.

apoelstra:
From Liam:

* We should commit to the set of generators (since for Elements rangeproofs this will always be the same we can precompute it, but it's good to commit as defense-in-depth)
* We should definitely try to construct a proof of length 0 and make sure that it doesn't validate (or should it? does such a thing make sense? I'm not sure)
* Need to add failing tests (Liam says he brought this up with you)

Other than that, and the nits I brought up above, this looks great to me! I am running local tests (will take a while on 12 commits) and want to squint at the verification a bit more. but otherwise ACK from me.

sanket1729:
@apoelstra and Liam, thanks for the review.

> We should commit to the set of generators (since for Elements range proofs this will always be the same we can precompute it, but it's good to commit as defense-in-depth)

If we add something to the transcript, we must use it in verifier validation. We would then need another API that allows this pre-computed transcript. 

Points 2) and 3), I will address in the range-proof PR where I can easily add failing test cases for the exposed API.

sanket1729:
> We should commit to the set of generators (since for Elements range proofs this will always be the same we can precompute it, but it's good to commit as defense-in-depth)

This would be done in the rangeproof PR. The norm-argument is just a sub-protocol. The initial hashing should take place before any of the code here is invoked. 

Overall, I agree with all the three points from Liam, but those are related to range-proof PR and are best addressed there.

sanket1729:
Force pushed an update. But looks like the CI here is broken :( .

apoelstra (approved):
ACK a978d16d84b2364adde92324b45e6b8b5a680a9f

apoelstra:
@sanket1729 can you open a PR to backport c0ae48c9950a908b637bff27791fabbe2833c4a5 from https://github.com/bitcoin-core/secp256k1/pull/1152 ?

Or I can try to do it later today.

sanket1729:
@apoelstra opened #213

jonasnick (commented):
I created a branch that adds the bulletproofs module to the CI tests because that seemed like a very low hanging fruit. Feel free to cherry-pick the three commits [here](https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn).

The CI jobs that have address sanitizer enabled do not succeed, but unfortunately without a sanitizer error. Instead, a `CHECK` is failing:
```
src/modules/bulletproofs/tests_impl.h:200: test condition failed: res == 1
```
I haven't tried to debug this further so far.

sanket1729:
This is interesting and I don't (yet) know what to make of it. I tried reproducing it locally, but everything is succeeding. 

All of the randomnesses in used in tests is deterministic and I am testing at the commit you posted 7229b85f22be2529830cf2a8efbecfb32b643d27. 
```
sanket1729:~/secp256k1-zkp$ ./tests 32 84326c02370fcc6962a79b9f4d36a80f
test count = 32
random seed = 84326c02370fcc6962a79b9f4d36a80f
Skipping test_ecmult_constants_2bit (iteration count too low)
random run = 93105842ab632cf398e7ef53d8538d1a
no problems found
```

Is there a way to update the CI file so that the same randomness is across targets? My suspicion is that it is something compiler version, target machine etc.

sanket1729:
My best possible guess is that those CI machine configs don't have enough memory and hence the malloc is failing. I added some print statements but seems like those are not displayed in the logs. Is this the best way to debug the CI here? 

What would be the best way to confirm this?

src/modules/bulletproofs/bulletproofs_util.h
piqjen-1xofFi-sikxyh: 🚀

jonasnick:
Sorry, I should have mentioned that I can reproduce this locally on my laptop! For example by running configure like so:

```
./configure --enable-experimental --enable-module-bulletproofs CFLAGS="-fsanitize=address -g" --without-asm 
```

> I added some print statements but seems like those are not displayed in the logs.

By the way, I can see the printfs when I expand "Run cat_tests_log" right below "Run tests" on [cirrus](https://cirrus-ci.com/task/6293890643263488) (which doesn't indicate that malloc failed).

src/modules/bulletproofs/tests_impl.h
```diff
@@ -0,0 +1,225 @@
...
+    }
+
+    for (i = 0; i < m; i++) {
+        random_scalar_order(&l_vec[i]);
+        random_scalar_order(&c_vec[i]);
+    }
+
+    res = secp256k1_bulletproofs_commit(ctx, scratch, &commit, gs, n_vec, n, l_vec, m, c_vec, m, &q);
+    CHECK(res == 1);
+    res = secp256k1_bulletproofs_pp_rangeproof_norm_product_prove(ctx, scratch, proof, &plen, transcript32, &r, gs, n_vec, n, l_vec, m, c_vec, m, &commit);
```
jonasnick, line 199: Ok I found it: `plen` is not initialized but read at the start of the `norm_product_prove` function. When I initialize plen as `sizeof(proof)` the tests succeed.

src/modules/bulletproofs/tests_impl.h
```diff
@@ -0,0 +1,225 @@
...
+}
+
+void norm_arg_test(unsigned int n, unsigned int m) {
+    secp256k1_scalar n_vec[64], l_vec[64], c_vec[64];
+    secp256k1_scalar r, q;
+    secp256k1_ge commit;
+    size_t i, plen;
+    int res;
+    secp256k1_bulletproofs_generators *gs = secp256k1_bulletproofs_generators_create(ctx, n + m);
+    secp256k1_scratch *scratch = secp256k1_scratch_space_create(ctx, 1000*1000); /* shouldn't need much */
```
jonasnick, line 180: These objects need to be destroyed at the end of the test.

sanket1729:
Force pushed and rebase with the following changes:
- Cleanup generators and scratch space objects in tests
- Corrected the documentation of generators' serialize/parse API. This doc was accidentally included in this PR, but the code in updated in the range-proof PR
-  Fixed usage of uninitialized variable `plen` in tests which correctly caused CI for fail on some machines.  
- Included fixup commits from @jonasnick.
- Added support for CI, again thanks to Jonas.

apoelstra:
Can you rebase on #213 now that this is merged?

Liam-Eagen:
I don't think the norm argument is adding the final scalar openings for the reduced l_vec and n_vec to the transcript, and I think it probably should. Not sure if it actually affects security, but for completeness should add since they are part of the transcript.

apoelstra (approved):
ACK ce7b9ada79f40812f7208a047026f494aa82232c

src/modules/bulletproofs/main_impl.h
```diff
@@ -7,6 +7,115 @@
...
+    if (ret->gens == NULL) {
+        free(ret);
+        return NULL;
+    }
+    ret->n = n;
+
+    secp256k1_fe_get_b32(&seed[0], &secp256k1_ge_const_g.x);
+    secp256k1_fe_get_b32(&seed[32], &secp256k1_ge_const_g.y);
+
+    secp256k1_rfc6979_hmac_sha256_initialize(&rng, seed, 64);
```
real-or-random, line 47: Huh, RFC6979 is terribly slow. I think that should just be a tagged hash. Or even use a tagged hash to create a ChaCha20 seed and then use ChaCha20 to expand the seed. (I think we'll need many many generators, so this could make a difference? I mean we could hardcode them in a further PR but maybe not every implementation wants to do this.)

include/secp256k1_bulletproofs.h
```diff
@@ -9,7 +9,62 @@ extern "C" {
 
 #include <stdint.h>
 
-/* TODO */
+/** Opaque structure representing a large number of NUMS generators */
+typedef struct secp256k1_bulletproofs_generators secp256k1_bulletproofs_generators;
+
+/** Allocates and initializes a list of NUMS generators.
+ *  Returns a list of generators, or NULL if allocation failed.
```
real-or-random, line 16: This should probably rather say that it calls the error callback when allocation failed. And since the error callback is not supposed to return, documenting the return value here is not necessary. (The function should still return NULL if the error callback returns unexpectedly.) Same below.

include/secp256k1_bulletproofs.h
```diff
@@ -9,7 +9,62 @@ extern "C" {
...
+ *  Args:        ctx: pointer to a context object
+ *               gen: pointer to the generator set to be serialized
+ *  Out:        data: pointer to buffer into which the generators will be serialized
+ *  In/Out: data_len: the length of the `data` buffer. Should be at least
+ *                    k = 33 * num_gens. Will be set to k on successful return
+ *
+ * TODO: For ease of review, this setting G = H0 is not included in this commit. We will
+ * add it in the follow-up rangeproof PR.
+ */
+SECP256K1_API int secp256k1_bulletproofs_generators_serialize(
```
real-or-random, line 52: Does the user need this function at all?

I guess my question is more general: How is this data structure supposed to be used? Do you imagine the user serializing and storing the set of generators to avoid recomputing it? How does this compare to accepting arrays of generators (then we'd get rid of the allocations in our code.)

src/modules/bulletproofs/main_impl.h
```diff
@@ -7,6 +7,115 @@
...
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(gens != NULL);
+    ARG_CHECK(data != NULL);
+    ARG_CHECK(data_len != NULL);
+
+    memset(data, 0, *data_len);
+    if (*data_len < 33 * gens->n) {
+        return 0;
+    }
```
real-or-random, line 104: I think that should be an ARG_CHECK.

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -0,0 +1,52 @@
...
+    sha->s[0] = 0x52fc8185ul;
+    sha->s[1] = 0x0e7debf0ul;
+    sha->s[2] = 0xb0967270ul;
+    sha->s[3] = 0x6f5abfe1ul;
+    sha->s[4] = 0x822bdec0ul;
+    sha->s[5] = 0x36db8beful;
+    sha->s[6] = 0x03d9e1f1ul;
+    sha->s[7] = 0x8a5cef6ful;
+
+    sha->bytes = 64;
```
real-or-random, line 37: It will be good to add a test that compares the output of `sha256_finalize` on a SHA object initialized using this function to the output of `sha256_finalize` on a plain SHA object with the proper prefix written. This will ensure that the constants here are correct.

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -49,4 +49,17 @@ static void secp256k1_bulletproofs_le64(unsigned char *output, const uint64_t n)
     output[7] = n >> 56;
 }
 
+/* Check if n is power of two*/
+static int secp256k1_check_power_of_two(size_t n) {
+    return n > 0 && (n & (n - 1)) == 0;
```
real-or-random, line 54: nit: maybe call this `_is_power_of_two`  ... (`check` reminds me of `CHECK` in this library...)

real-or-random (commented):
Some comments on the first 7 commits. More soon. :)

src/modules/bulletproofs/main_impl.h
apoelstra: In practice, at least for rangeproofs we'll have a hardcoded generator list. The use of RFC6979 is just so that people can reproduce our work to generate the NUMS points.

Agreed that chacha would be better, but I don't think it's a big deal.

include/secp256k1_bulletproofs.h
apoelstra: If we accept arrays of generators then how does the user produce that array? By manually defining a serialization format?

I'm fine removing this code but we'll just have to recreate it in Elements.

Liam-Eagen:
I still think we should add failing tests, but this can be done in the range proof PR. Also think we should maybe re-considering how the transcript and proof are encoded into memory. I don't understand the benefit of combining the sign bits from the response points versus just combining all the sign bits when encoding the final proof. Otherwise looks good to me.

Liam-Eagen:
ACK

apoelstra:
@Liam-Eagen if the transcript and proof are encoded identically this makes the verifier's job easier because it can just feed bytes from the proof directly into the transcript.

It also reduces the likelihood of "malleability" bugs where somebody can change the encoded proof without changing its validity.

src/modules/bulletproofs/main_impl.h
real-or-random: > In practice, at least for rangeproofs we'll have a hardcoded generator list.

Is this list in https://github.com/ElementsProject/secp256k1-zkp/pull/207? I can't find it, but I've only skimmed that PR. I think I don't get the entire picture of the API yet.

apoelstra: No, it will be hardcoded in Elements. This library tries to be agnostic about the set of generators.

include/secp256k1_bulletproofs.h
real-or-random: Ok, yes, when I wrote this comment I hadn't understood yet that the goal is to accept generators (see below).

real-or-random:
Here's a more high-level comment on the API: 

I think the generators used by the Bulletproofs++ protocol are an implementation detail. That means the user shouldn't need to bother with them at all. Instead of letting the user specify them, we could just use the same generators for every application (like secp256k1's generator is predefined for all application).

In the library, we could generate generators on the fly, or hardcode them. We could also hardcode a some compilation-time constant number of them, and generate generators which are not hardcoded on the fly.

This will have the following advantages:
 - The generators won't appear in the API at all, and the caller doesn't need to bother with generators at all. No need to create, serialize, store, them. 
 - We can change the implementation here (hardcode vs on-the-fly) as we wish without changing the API.
 - We get rid of dynamic memory management for the generator code.
 - The user can't accidentally pass a generator which is not NUMS. (Okay, this point is weak: Even with the current PR, the user would need to serialize his points manually to be able to do this...)  

Alternatively, we could have some kind of generator object that manages precomputation of the generators. This is essentially what the current code does, but then I suggest getting rid of the serialization and deserialization functions. No need to add them in the caller code.  The user will just run `secp256k1_bulletproofs_generators_create(ctx, n)` for some `n` in the beginning. This adds some startup overhead, but I don't think that's a problem. (And even if this is a problem, we can again hardcode some number of generators that `secp256k1_bulletproofs_generators_create` won't have to generate them.)

apoelstra:
@real-or-random I like all these ideas. Though they should definitely be a followup PR rather than being in this one.

We should also include some benchmarks for the generator code. At first I thought "if we use chacha20 rather than sha2_hmac_rng" then we can random-access-generate them and won't need to store them. But now that I'm looking at the `shallue_van_de_woestijne` code .. there's three field sqrts and a modular inverse .. and we call this function twice to generate a point. So it may be very expensive.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+    secp256k1_bulletproofs_powers_of_r(r_inv_pows, &r_inv, log_n);
+
+    secp256k1_sha256_initialize(&sha256);
+    secp256k1_sha256_write(&sha256, transcript_hash32, 32);
+    secp256k1_fe_normalize_var(&comm.x);
+    secp256k1_fe_normalize_var(&comm.y);
+    ser_commit[0] = 0x02 | secp256k1_fe_is_odd(&comm.y);
+    secp256k1_fe_get_b32(&ser_commit[1], &comm.x);
+    secp256k1_sha256_write(&sha256, ser_commit, 33);
+    secp256k1_sha256_finalize(&sha256, transcript_hash32);
```
jonasnick, line 506: I was quickchecking the rust implementation and noticed that X and R can become the point at infinity. @Liam-Eagen and I found multiple conditions in which this can happen, e.g. when l and n don't have the same length and the longer one is the zero vector. I added a test case to this C code to demonstrate this [here](https://github.com/jonasnick/secp256k1-zkp/commit/2cbfe52571d9f200097158f341cd05465197cef6#diff-fd6906ded5f921ba870d059568d0d8a2ca9bc495d09e0738557cce6008603d3fR256).

As the added test also demonstrates, if X or R are the point at infinity then proving succeeds but verifying fails. That's because the prover algorithm serializes some invalid X-coordinate into the proof when X or R are the point at infinity. The verifier calls `eckey_parse` on the proof data which fails on invalid X-coordinates.

If we want to avoid that, I see two options:
1. make the prover fail when X or R are zero and add a note to the function documentation that either says that the function may fail when the inputs are not uniformly random or state the exact conditions when it happens
2. properly parse and serialize the point at infinity (the musig PR has some utilities for doing that)

I'd prefer option 2 because it doesn't add an error branch and doesn't complicate the doc. I've also updated the rust code to implement option 2.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,80 @@
...
+        secp256k1_scalar_mul(&term, &a_vec[a_offset + step*i], &b_vec[b_offset + step*i]);
+        secp256k1_scalar_add(res, res, &term);
+    }
+    return 1;
+}
+
+/* Computes the q-weighted inner product of two vectors of scalars
+ * for elements starting from offset a and offset b respectively with the
+ * given step.
+ * Returns: Sum_i(a[offset_a + step*i] * b[offset_b2 + step*i]*q^(i+1)) */
```
real-or-random, line 44: This should probably document also `len`.

src/modules/bulletproofs/main_impl.h
```diff
@@ -11,11 +11,13 @@
 struct secp256k1_bulletproofs_generators {
     size_t n;
     /* n total generators; set n = 2*k to get G_i and H_i values for i in [1..k] */
+    /* For BP++, the generators are G_i from [1..(n - 6)]* and the last values are generators for H */
```
real-or-random, line 14: Nit: Is the `*` a typo here? Or what does `[1..(n - 6)]*` mean?

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -77,4 +77,67 @@ static void secp256k1_bulletproofs_powers_of_r(secp256k1_scalar *powers, const s
...
+    secp256k1_scratch_space* scratch,
+    secp256k1_ge* commit,
+    const secp256k1_bulletproofs_generators* g_vec,
+    const secp256k1_scalar* n_vec,
+    size_t n_vec_len,
+    const secp256k1_scalar* l_vec,
+    size_t l_vec_len,
+    const secp256k1_scalar* c_vec,
+    size_t c_vec_len,
+    const secp256k1_scalar* q
```
real-or-random, line 111: The variables in the comment don't match the function arguments. (I think it's a good idea to have this `_vec` suffix, by the way.)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -77,4 +77,67 @@ static void secp256k1_bulletproofs_powers_of_r(secp256k1_scalar *powers, const s
         secp256k1_scalar_sqr(&powers[i], &powers[i - 1]);
     }
 }
+
+typedef struct ec_mult_bp_commit_cb_data {
```
real-or-random, line 81: This is a really a micro nit: We usually spell this `ecmult` in code/identifiers (instead of `ec_mult`)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -77,4 +77,67 @@ static void secp256k1_bulletproofs_powers_of_r(secp256k1_scalar *powers, const s
...
+        secp256k1_gej commitj;
+        data.g = g_vec->gens;
+        data.n = n_vec;
+        data.l = l_vec;
+        data.g_len = n_vec_len;
+
+        if (!secp256k1_ecmult_multi_var(&ctx->error_callback, scratch, &commitj, &v, ec_mult_bp_commit_cb, (void*) &data, n_vec_len + l_vec_len)) {
+            return 0;
+        }
+        secp256k1_ge_set_gej_var(commit, &commitj);
```
real-or-random, line 139: Wouldn't it make sense to keep this a `gej`? It's an internal function anyway.

real-or-random (commented):
Would it make sense to call the module `bppp` instead of `bulletproofs`?

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
apoelstra: I also prefer option 2. We've had problems in past crypto (surjection proofs) where we discovered "legitimate" uses of intermediate infinity values but weren't able to use them ... and more annoyingly, being able to "successfully" create proofs for them that wouldn't verify.

real-or-random (commented):
There should also be a corresponding header `bulletproofs_pp_norm_product.h` that documents the internal interface (and the docs should be moved there.)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+    secp256k1_scratch_space* scratch,
+    unsigned char* proof,
+    size_t proof_len,
+    unsigned char *transcript_hash32,
+    const secp256k1_scalar* r_ch,
+    secp256k1_bulletproofs_generators* g_vec,
+    size_t g_len,
+    secp256k1_scalar* c_vec,
+    size_t c_vec_len,
+    const secp256k1_ge* commit
```
jonasnick, line 463: This argument is removed in the follow-up PR and it's unnecessary. Above `norm_produce_verify` assumes that `r` and `c_vec` (as @Liam-Eagen pointed out) are already included in `transcript_hash`, so it could also just assume that `commit` is in the transcript hash.

It'd be really great if we could clearly document the assumptions we put on the argument of the function with respect to the transcript because this is a common source of vulnerabilities. It will also make it easier to review the follow-up PR.

Liam-Eagen:
General comment: After discussion with @real-or-random think it is important to note that in the code all lengths should be powers of 2, so that all the places where division by two occurs should be valid. This is verified, but may not be obvious at the divisions by 2. Discussed possibility of switching to explicit powers of 2 with @apoelstra as well and think it is probably better to keep in length form. Would likely still need to do range checks for the log form, and keeping length form might make it easier to support non-powers of 2 in the future.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+        }
+    }
+    {
+        ec_mult_verify_cb_data2 data;
+        data.g_vec = g_vec->gens;
+        data.g_vec_len = g_len;
+        data.s_g = s_g;
+        data.s_h = s_h;
+
+        if (!secp256k1_ecmult_multi_var(&ctx->error_callback, scratch, &res2, &v, ec_mult_verify_cb2, &data, g_len + h_len)) {
```
jonasnick, line 577: Did you split it into two separate `ecmult_mult_multi`s because it's easier to write and read? If we would have just one large ecmult_multi it would be faster.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -388,4 +388,202 @@ int secp256k1_bulletproofs_pp_rangeproof_norm_product_prove(
...
+        *pt = *data->commit;
+        secp256k1_scalar_set_int(sc, 1);
+        return 1;
+    }
+    idx -= 1;
+    if (idx % 2 == 0) {
+        unsigned char pk_buf[33];
+        idx /= 2;
+        *sc = data->challenges[idx];
+        pk_buf[0] = 2 | (data->proof[65*idx] >> 1);
```
real-or-random, line 410: If only that "then" branch is executed, then the verifier ignores the LSB of `data->proof[65*idx]`. Is that a problem? I believe it's not a problem for third-party malleability (because the points get hashed into the challenges)? I believe it could be a problem if we care about "uniqueness" of the norm argument, i.e., even the prover can create just a single valid proof for any statement. Should we care about this? I don't think it will matter, but then good cryptographic practice says we should care.  

Can it happen that only the "then" branch is executed? I think no, because we only accept vectors of size a power of two. But it seems still somewhat fragile.

Summary:
So I *think* this is somewhat fragile if we ever extend this to handle vectors which are not a power of two.

(All of this may be wrong, I haven't thought it through entirely.)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+        data.proof = proof;
+        data.commit = commit;
+        data.challenges = es;
+
+        secp256k1_gej_set_ge(&temp2, commit);
+        secp256k1_scalar_set_int(&one, 1);
+        secp256k1_ecmult(&temp1, &temp2, &one, NULL);
+
+        if (!secp256k1_ecmult_multi_var(&ctx->error_callback, scratch, &res1, NULL, ec_mult_verify_cb1, &data, 2*n_rounds + 1)) {
+            return 0;
```
jonasnick, line 567: I think it's important here to apply the checkpoint in case of failure, otherwise the scratch space is never reclaimed (which together with not checking `r_inv_pow != NULL` above would allow crashing the verifier).

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+    secp256k1_scalar_set_b32(&l, &proof[n_rounds*65 + 32], &overflow); /* l */
+    if (overflow) return 0;
+    if (secp256k1_scalar_is_zero(r_ch)) return 0;
+
+    /* Compute powers of q_inv. Later used in g_factor computations*/
+    r = *r_ch;
+    secp256k1_scalar_inverse_var(&r_inv, &r);
+    scratch_checkpoint = secp256k1_scratch_checkpoint(&ctx->error_callback, scratch);
+
+    r_inv_pows = (secp256k1_scalar*)secp256k1_scratch_alloc(&ctx->error_callback, scratch, log_n * sizeof(secp256k1_scalar));
```
jonasnick, line 496: We should check that `r_inv_pows` is not `NULL`.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+/* Verify the proof */
+int secp256k1_bulletproofs_pp_rangeproof_norm_product_verify(
+    const secp256k1_context* ctx,
+    secp256k1_scratch_space* scratch,
+    unsigned char* proof,
+    size_t proof_len,
+    unsigned char *transcript_hash32,
+    const secp256k1_scalar* r_ch,
+    secp256k1_bulletproofs_generators* g_vec,
+    size_t g_len,
```
jonasnick, line 460: Is the g_len argument necessary? It seems to be implied by `c_vec_len` and `gens -> n`.

sanket1729 (commented):
Went through the review. Will address the changes soon

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
sanket1729: > It'd be really great if we could clearly document the assumptions we put on the argument of the function with respect to the transcript because this is a common source of vulnerabilities. It will also make it easier to review the follow-up PR.

My initial goal was to create a stand-alone norm argument protocol and compose it with range proofs API. However, as I have slowly realized through this review, these are so easily composable for efficiency reasons. If we want to standalone norm argument, we should add the commitment `C` to transcript-hash because that is the initial input to the norm argument.

But if we use it in a parent protocol, the prover needs to invest additional resources to compute the final commitment only to commit it in the transcript. This is unnecessary as the final commit is a deterministic function of already committed points and is not shared in prover verifier interaction during the norm argument. 

To elaborate, there are three protocols:

1) `norm_arg_prove_internal` :  This does not commit to initial commitment points and assumes that the parent protocol has committed to it. We **want** `secp256k1_bulletproofs_pp_rangeproof_norm_product_prove` to do this.
2) `norm_arg_prove`: This should be a standalone function that we can use for norm argument. This would be the wrapper around `norm_arg_prove_internal` that adds commitment points, generators, etc to the transcript. The current version of `secp256k1_bulletproofs_pp_rangeproof_norm_product_prove` does this instead.
3) `rangeproof_prove`: This should use the (2) as a sub-protocol, not (1). The second PR changes this function to do (2) instead of (1). 

I think I will change the PR to 
- Change the secp256k1_bulletproofs_pp_rangeproof_norm_product_prove to follow (1)
- add the `norm_arg_prove`, `norm_arg_verify` and `norm_arg_commit` according to protocol (2) in tests as wrappers around (1)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
real-or-random: Sounds good to me!

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -140,4 +140,252 @@ static int secp256k1_bulletproofs_commit(
...
     return 1;
 }
+
+typedef struct ec_mult_x_cb_data {
+    const secp256k1_scalar *n;
+    const secp256k1_ge *g;
+    const secp256k1_scalar *l;
+    const secp256k1_scalar *r;
+    const secp256k1_scalar *r_inv;
+    size_t G_GENS_LEN; /* Figure out initialization syntax so that this can also be const */
```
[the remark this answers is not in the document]
real-or-random: Ok, makes sense. Then the comment should be removed (or changed to explain this reasoning.)

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
real-or-random: > I _think_ I understand what you are trying to say, but am not completely sure. I think you are suggesting that there is an unused bit that can be changed.

Yes!

> The prover can create different proofs by choosing different nonces. So, perhaps this point is not possible in general. I believe this is also the case with schnorr signatures.

It depends. The norm argument doesn't have a nonce, so the norm argument on its own could perhaps be made "unique". (I don't know if that's useful. Maybe it is useful if the norm argument will be exposed as a separate API in the future. My point is that it's cheap to make sure the unused bit can't be changed, and it certainly won't hurt.)

> The way we are planning to support non-powers of two is by extending the vectors to powers of two by adding dummy zeros. And then use some optimizations to avoid computations because we know the last values would be zero.

Ok, this would probably avoid the issue anyway.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,589 @@
...
+    }
+    *pt = data->g_vec[idx];
+    return 1;
+}
+
+/* Verify the proof */
+int secp256k1_bulletproofs_pp_rangeproof_norm_product_verify(
+    const secp256k1_context* ctx,
+    secp256k1_scratch_space* scratch,
+    unsigned char* proof,
```
jonasnick, line 455:
```suggestion
    const unsigned char* proof,
```

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
sanket1729: It is not necessary. But it seemed easy to have it than not. I can also remove it and infer it from as `gens->n - c_vec_len` as you mention

sanket1729:
Comments not addressed here: 
1) generators API: compile time hard-coded and removed the high level API. https://github.com/ElementsProject/secp256k1-zkp/pull/205#issuecomment-1404865622.
2) Allowing proofs to have serialized points at infinity.

sanket1729 (commented):
Force pushed a new version. It has a cleaner transcript API which was used in the second PR.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
jonasnick: > Change the secp256k1_bulletproofs_pp_rangeproof_norm_product_prove to follow (1)

I agree.

> add the norm_arg_prove, norm_arg_verify and norm_arg_commit according to protocol (2) in tests as wrappers around (1)

I like this idea.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
jonasnick: Fine for me to do in a separate PR.

src/modules/bulletproofs/tests_impl.h
```diff
@@ -0,0 +1,369 @@
...
+    secp256k1_fe_normalize(&comm.y);
+    secp256k1_bulletproofs_serialize_pt(&ser_commit[0], &comm);
+    secp256k1_sha256_write(&transcript, ser_commit, 33);
+    for (i = 0; i < g_vec->n; i++) {
+        secp256k1_fe_normalize(&g_vec->gens[i].x);
+        secp256k1_fe_normalize(&g_vec->gens[i].y);
+        secp256k1_bulletproofs_serialize_pt(&ser_commit[0], &g_vec->gens[i]);
+        secp256k1_sha256_write(&transcript, ser_commit, 33);
+    }
+    res = secp256k1_bulletproofs_pp_rangeproof_norm_product_verify(
```
jonasnick, line 295: I think we're missing `r_ch` and `c_vec` in the transcript.

src/modules/bulletproofs/tests_impl.h
```diff
@@ -0,0 +1,369 @@
...
+    memcpy(ls, l_vec, h_len * sizeof(secp256k1_scalar));
+    memcpy(cs, c_vec, h_len * sizeof(secp256k1_scalar));
+    memcpy(gs, g_vec->gens, (g_len + h_len) * sizeof(secp256k1_ge));
+
+    /* Commit to the initial public values */
+    secp256k1_bulletproofs_pp_sha256_tagged_commitment_init(&transcript);
+    secp256k1_fe_normalize(&comm.x);
+    secp256k1_fe_normalize(&comm.y);
+    secp256k1_bulletproofs_serialize_pt(&ser_commit[0], &comm);
+    secp256k1_sha256_write(&transcript, ser_commit, 33);
```
jonasnick, line 237: I think just using sha256_write to append to the transcript is dangerous. For example, if the transcript `T` consists of two vectors of scalars `l` and `r` such that:
```
l = [1, 2]; r = [3]; T = [l, r]
```
and similarly for transcript T' 
```
l' = [1]; r' = [2,3]; T' = [l', r']
```
Then `T != T'` but the challenge we compute is the same because input to SHA256 is just `[1,2,3]`. So it would be much better to have a transcript object that can be written to like sha256, but internally prepends the length (in bytes) of the input before calling sha256.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
jonasnick: Fine for me. I've updated the tests that exercise this branch accordingly.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
jonasnick: This branch is now exercised in the static verification test vectors in my branch.

src/modules/bulletproofs/tests_impl.h
sanket1729: Good catch. We should certainly prepend the length. I take it that you are suggesting that we serialize all the vector and include the total byte size as an argument.

I think it might be more logical to add a method to add another method to `transcript_consume_vec` that would add the len too. Obviously, this is footgunny because the user can write the for loop (just like I did above). But the approach where we push the entire vector as a blob feels odd. It also does not solve the above issue you mention where the user can still call items in a loop as individual elements.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
sanket1729: This PR now uses the new FS scheme.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -140,4 +140,252 @@ static int secp256k1_bulletproofs_commit(
...
+    ec_mult_r_cb_data r_cb_data;
+    size_t g_len = n_vec_len, h_len = l_vec_len;
+    const size_t G_GENS_LEN = g_len;
+    /* The initial pointers to the callback function remain the same. We mutate
+     * the values and len in loop*/
+    secp256k1_scalar *ns, *ls, *cs;
+    secp256k1_ge *gs, comm = *commit;
+    size_t scratch_checkpoint;
+    size_t log_n = secp256k1_bulletproofs_pp_log2(g_len), log_m = secp256k1_bulletproofs_pp_log2(h_len);
+    size_t n_rounds = log_n > log_m ? log_n : log_m;
```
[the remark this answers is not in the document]
sanket1729: done

jonasnick:
I added more verification test vectors to my branch and simplified them a little bit (feel free to cherry-pick).

src/modules/bulletproofs/main_impl.h
```diff
@@ -0,0 +1,123 @@
+/**********************************************************************
+ * Copyright (c) 2020 Andrew Poelstra                                 *
+ * Distributed under the MIT software license, see the accompanying   *
+ * file COPYING or http://www.opensource.org/licenses/mit-license.php.*
+ **********************************************************************/
+
+#ifndef _SECP256K1_MODULE_BULLETPROOFS_MAIN_
+#define _SECP256K1_MODULE_BULLETPROOFS_MAIN_
+
+/* this type must be completed before any of the modules/bulletproofs includes */
```
jonasnick, line 10: fixed that in my branch

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -38,4 +38,17 @@ static void secp256k1_bulletproofs_le64(unsigned char *output, const uint64_t n)
...
+}
+
+/* Compute the log2 of n. If n is not a power of two, it returns the largest
+ * `k` such that 2^k <= n. Assumes n < 2^64. In Bulletproofs, this is bounded
+ * by len of input vectors which can be safely assumed to be less than 2^64.
+*/
+static size_t secp256k1_bulletproofs_pp_log2(size_t n) {
+    return 64 - 1 - secp256k1_clz64_var((uint64_t)n);
+}
```
real-or-random, line 52: Is there a reason not to take `uint64_t` as an argument directly?

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,540 @@
...
+) {
+    secp256k1_scalar v, l_c;
+    /* First n_vec_len generators are Gs, rest are Hs*/
+    CHECK(g_vec->n == (n_vec_len + l_vec_len));
+    CHECK(l_vec_len == c_vec_len);
+
+    /* It is possible to extend to support n_vec and c_vec to not be power of
+    two. For the initial iterations of the code, we stick to powers of two for simplicity.*/
+    CHECK(secp256k1_is_power_of_two(n_vec_len));
+    CHECK(secp256k1_is_power_of_two(c_vec_len));
```
jonasnick, line 125:
```suggestion
    VERIFY_CHECK(g_vec->n == (n_vec_len + l_vec_len));
    VERIFY_CHECK(l_vec_len == c_vec_len);

    /* It is possible to extend to support n_vec and c_vec to not be power of
    two. For the initial iterations of the code, we stick to powers of two for simplicity.*/
    VERIFY_CHECK(secp256k1_is_power_of_two(n_vec_len));
    VERIFY_CHECK(secp256k1_is_power_of_two(c_vec_len));
```

`CHECK` are intended only for tests.

jonasnick:
> Would it make sense to call the module `bppp` instead of `bulletproofs`?

Would be fine by me. Less typing.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
real-or-random: agreed

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
```diff
@@ -0,0 +1,540 @@
...
+    }
+    /* s_g[0] = n * r^(n - 1) = n * (r)^n * r_inv = n * q * r_inv */
+    secp256k1_scalar_mul(&s_g[0], &n, r);
+    secp256k1_scalar_mul(&s_g[0], &s_g[0], &r_inv);
+    for (i = 1; i < g_len; i++) {
+        size_t log_i = secp256k1_bulletproofs_pp_log2(i);
+        size_t nearest_pow_of_two = (size_t)1 << log_i;
+        secp256k1_scalar_mul(&s_g[i], &s_g[i - nearest_pow_of_two], &es[log_i]);
+        secp256k1_scalar_mul(&s_g[i], &s_g[i], &r_inv_pows[log_i]);
+    }
```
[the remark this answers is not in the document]
jonasnick: Ok, the rust code does require a second loop. @Liam-Eagen explained to me how your code can this save one loop. Cool! But would be good to document this better.

src/modules/generator/main_impl.h
```diff
@@ -249,6 +249,7 @@ static void secp256k1_pedersen_commitment_load(secp256k1_ge* ge, const secp256k1
     if (commit->data[0] & 1) {
         secp256k1_ge_neg(ge, ge);
     }
+    secp256k1_fe_normalize(&ge->y);
```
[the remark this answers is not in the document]
sanket1729: cc @apoelstra. Sorry, I did not think about this at all. This was just cherry-picked from @apoelstra's old bulletproofs PR.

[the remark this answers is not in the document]
apoelstra: Heh, I have long since forgotten.

src/modules/bulletproofs/bulletproofs_util.h
sanket1729: Don't know how the conventions are here. Everywhere in the code, this is called with `size_t` argument. Perhaps, casting at the calling sites is better?

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -0,0 +1,41 @@
...
+
+/* Outputs a serialized point in compressed form. Assumes point is normalized */
+static void secp256k1_bulletproofs_serialize_pt(unsigned char *output, const secp256k1_ge *lpt) {
+    output[0] = (1 << 1) + secp256k1_fe_is_odd(&lpt->y);
+    secp256k1_fe_get_b32(&output[1], &lpt->x);
+}
+
+/* little-endian encodes a uint64 */
+static void secp256k1_bulletproofs_le64(unsigned char *output, const uint64_t n) {
```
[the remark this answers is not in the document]
sanket1729: We will need the 64 bits in the next PR to commit the `min_value` in the transcript. Would be good to use the same function here.

src/modules/bulletproofs/bulletproofs_util.h
apoelstra: I think that we should use `size_t` for lengths, unless we seriously expect to encounter lengths in excess of 2^32 and we want to support that.

apoelstra: Though we'd then encounter problems with array indexing etc

sanket1729:
Force pushed with the latest requested changes. I renamed `bulletproofs_pp*` to `bppp*` in the last commit. It would have been annoying to do that across all commits.

src/modules/bulletproofs/bulletproofs_util.h
sanket1729: Keeping this as `size_t` for now.

src/modules/bulletproofs/bulletproofs_util.h
```diff
@@ -0,0 +1,41 @@
...
+ * Assumes both points' coordinates have been normalized.
+ */
+static void secp256k1_bulletproofs_serialize_points(unsigned char *output, const secp256k1_ge *lpt, const secp256k1_ge *rpt) {
+    output[0] = (secp256k1_fe_is_odd(&lpt->y) << 1) + secp256k1_fe_is_odd(&rpt->y);
+    secp256k1_fe_get_b32(&output[1], &lpt->x);
+    secp256k1_fe_get_b32(&output[33], &rpt->x);
+}
+
+/* Outputs a serialized point in compressed form. Assumes point is normalized */
+static void secp256k1_bulletproofs_serialize_pt(unsigned char *output, const secp256k1_ge *lpt) {
```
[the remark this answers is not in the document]
sanket1729: Done. Also cleanly allows us to deal with point at infinity case.

include/secp256k1_bulletproofs.h
```diff
@@ -9,7 +9,63 @@ extern "C" {
...
+ * in a separate commit to make review easier.
+ */
+SECP256K1_API secp256k1_bulletproofs_generators *secp256k1_bulletproofs_generators_create(
+    const secp256k1_context* ctx,
+    size_t n
+) SECP256K1_ARG_NONNULL(1);
+
+/** Allocates a list of generators from a static array
+ *  Returns a list of generators, or calls the error callback if the allocation
+ *  or parsing fails.
```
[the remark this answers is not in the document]
sanket1729: Done

src/modules/bulletproofs/test_vectors/verify.h
```diff
@@ -0,0 +1,65 @@
...
+static secp256k1_scalar verify_vector_2_c_vec[1];
+static const unsigned char verify_vector_2_r32[32] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x36 };
+static const unsigned char verify_vector_2_proof[] = { 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x00, 0x05, 0x78, 0x14, 0x7C, 0xF5, 0xDE, 0xFF, 0xE6, 0x5A, 0xA1, 0xD1, 0xDC, 0x9E, 0x7D, 0x87, 0x42, 0x22, 0x45, 0xCC, 0x43, 0xBE, 0x9E, 0x45, 0xFA, 0x2C, 0xEB, 0x9D, 0xC7, 0x66, 0x2F, 0xF4, 0xB2, 0x5D, 0xAF, 0x80, 0xA9, 0x72, 0xCB, 0xC0, 0x21, 0x3B, 0x45, 0xD2, 0xD3, 0x98, 0x89, 0x71, 0x22, 0xF2, 0x38, 0x56, 0x2C, 0x73, 0xAE, 0x1A, 0x8A, 0xC0, 0x45, 0x00, 0xAC, 0x23, 0x0C, 0xD6, 0x63, 0x6F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x3E };
```
jonasnick, line 21: For some reason, these two test vector lines are different from the vectors in my branch. According to the rust code that produces them they should be:

```suggestion
static const unsigned char verify_vector_2_r32[32] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x3A };
static const unsigned char verify_vector_2_proof[] = { 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x3F, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFE, 0xBA, 0xAE, 0xDC, 0xE6, 0xAF, 0x48, 0xA0, 0x3B, 0xBF, 0xD2, 0x5E, 0x8C, 0xD0, 0x36, 0x41, 0x41 };
```

src/modules/bulletproofs/tests_impl.h
```diff
@@ -0,0 +1,545 @@
...
+    secp256k1_scalar_sqr(&q, &r);
+
+    /* l and n are the zero vector */
+    {
+        unsigned int n_vec_len = 1;
+        unsigned int c_vec_len = 1;
+        secp256k1_bulletproofs_generators *gs = secp256k1_bulletproofs_generators_create(ctx, n_vec_len + c_vec_len);
+        size_t plen = sizeof(proof);
+        for (i = 0; i < n_vec_len; i++) {
+            secp256k1_scalar_set_int(&n_vec[i], 1);
```
jonasnick, line 372: Is there a reason you changed this from 0 to 1? The test description says that `n_vec[i]` should be 0.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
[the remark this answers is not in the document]
jonasnick: You added a comment a few lines, but I still couldn't make sense of the original comment: n is used both to refer to the reduced norm vector as well as the length of the original norm vector. `q` is in the equation but not used in the code below. `r` doesn't refer to the `r` argument provided to the function.

I created a new branch https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn-2 that contains an attempt to improve this and also adds a few other variable renames for clarification (e.g., `log_n` in this PR isn't the logarithm of `n` but the logarithm of the length of the norm vector).

I also made the `r` argument constant because that seemed much less error prone for the caller. But maybe this was intententional and the caller should actually get `r_f` from the function? In PR #207 it doesn't look like the `r_f` output is being used. Same for the prover, so could we make it constant for the prover as well?

src/modules/bulletproofs/tests_impl.h
sanket1729: Yes, because the bulletproofs commitment was the point at infinity, because all `n`, `l` and `v` are zero. The new serialization code as per @real-or-random 's suggestion `ec_pubkey_serialize` correctly fails to serialize this. I had to update to test to reflect this. 

My understanding was this test should only check that this test only produces two scalars which are input `n[0]` and `l[0]` directly. Them being zero really does not change anything.

src/modules/bulletproofs/test_vectors/verify.h
sanket1729: Interesting, perhaps some rebase messup on my end. I will re-run the rust code on my end use those test cases.

src/modules/bulletproofs/tests_impl.h
sanket1729: I thought I added in the commit message description, but seems like not :( .

sanket1729:
CI was failing because of spurious timeouts. https://cirrus-ci.com/task/6603390013145088. Force pushed to restart the CI

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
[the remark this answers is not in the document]
sanket1729: Sorry, I missed this comment and forgot to address this. Overall, I like your fixups. Will add those.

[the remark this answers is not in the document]
sanket1729: > I also made the r argument constant because that seemed much less error prone for the caller. But maybe this was intentional and the caller should actually get r_f

I only did this because it was an internally used function and wanted to avoid the extra copy. In an earlier implementation(1~ week ago) I had everything as const `n_vec`, `l_vec`, `c_vec` as consts. 

But for avoiding the mallocs inside the `secp256k1_bpp_norm_arg_prove_impl`, I changed those to non-const versions. While doing that, I also changed r. Don't mind keeping the r as `const` if it helps to understand.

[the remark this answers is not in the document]
jonasnick: I think there's a difference between copying O(1) (`r`) and mallocing+copying O(n) (`n_vec`, `l_vec`, `c_vec`). I think the former is definitely fine, the latter is debatable. So if that was the only reason for making `r` not const, then I think it'd be better if we'd make it `const` again.

src/modules/bulletproofs/tests_impl.h
jonasnick: The intention of the test was actually to test what happens if you pass zero vectors. The new test is somewhat redundant as we already have a static test vector. So I think it would be preferable to reflect the `ec_pubkey_serialize` change in the tests by checking that the functions fail as expected.

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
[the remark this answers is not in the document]
sanket1729: Force pushed to fix this.

src/modules/generator/main_impl.h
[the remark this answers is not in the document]
sanket1729: I am removing this

src/modules/bulletproofs/bulletproofs_pp_norm_product_impl.h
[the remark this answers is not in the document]
sanket1729: Done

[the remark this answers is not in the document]
sanket1729: Done

jonasnick:
This branch should fix CI: https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn-3.

src/modules/bulletproofs/tests_impl.h
jonasnick: I added a commit to https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn-3 that allows the commitment to be the point at infinity again. The reason is that we want the norm argument to be usable by a variety of protocols which don't necessarily forbid serializing the point at infinity into the transcript. In particular, for verify and prove we shouldn't rely on the outer protocol to serialize points in a certain way. Therefore, it's better to test verify and prove for the zero vector (good to test the edges anyway) which required that `commit()` succeeds even if its output is the point at infinity.

sanket1729:
> This branch should fix CI: [jonasnick/secp256k1-zkp@b%2B%2B-jn-3 (commits)](https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn-3).

Done.

src/modules/bppp/bppp_norm_product_impl.h
```diff
@@ -0,0 +1,548 @@
...
+        *sc = data->s_h[idx - data->g_vec_len];
+    }
+    *pt = data->g_vec[idx];
+    return 1;
+}
+
+/* Verify the proof. This function modifies the generators, c_vec and the challenge r. The
+   caller should make sure to back them up if they need to be reused.
+*/
+int secp256k1_bppp_rangeproof_norm_product_verify(
```
jonasnick, line 425: missing static

jonasnick (commented):
The updates look good to me! I checked some of the potential problems with integer conversions and I don't think there are any. But I noticed that bppp_log2 can be called with 0 which results in return value of SIZE_MAX. This can lead to funky things when someone is persuaded to call prove/vector with any of the vectors having length 0. I added a commit to my branch at
https://github.com/jonasnick/secp256k1-zkp/commits/b%2B%2B-jn-4
that explicitly prevents this and adds a test.

jonasnick (commented):
ACK d7fb25c8ca5bda0e969ce94ccabedfd7b8432769

PR #211, opened 2023-01-11
Update musig module to BIP MuSig2 v1.0.0-rc.3

jonasnick:
Version 1.0.0-rc.3 of BIP MuSig2 can be found [here](https://github.com/jonasnick/bips/pull/75). This PR does _not_ implement the following optional features that have been added to BIP MuSig2:

- variable length messages
- deterministic signing
- identifiable aborts

The PR also does _not_ yet change the `secnonce` structure to also contain the signer's public key (which would also imply changing the seckey argument in `sign` to a keypair). Additionally, we may want to rename some things in the future to be more consistent with the BIP (e.g. keyagg_cache vs. keyagg_ctx, applytweak vs. tweak_add).


instagibbs:
fwiw: interface changes make sense to me and will result in fewer loc in my usage

real-or-random:
ci says you need to update the cttime tests:
https://cirrus-ci.com/task/6684150934011904?logs=test#L177

src/modules/musig/session_impl.h
```diff
@@ -226,42 +226,54 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
...
+    secp256k1_sha256_write(&sha, rand, sizeof(rand));
+    secp256k1_nonce_function_musig_helper(&sha, 1, agg_pk32);
+    secp256k1_nonce_function_musig_helper(&sha, 1, msg32);
+    secp256k1_nonce_function_musig_helper(&sha, 4, extra_input32);
 
     for (i = 0; i < 2; i++) {
         unsigned char buf[32];
-        secp256k1_sha256_initialize(&sha);
-        secp256k1_sha256_write(&sha, seed, 32);
-        secp256k1_sha256_write(&sha, &i, sizeof(i));
```
real-or-random, line 263: so this was wrong even before this PR?

src/modules/musig/keyagg_impl.h
```diff
@@ -47,13 +47,30 @@ static void secp256k1_point_load(secp256k1_ge *ge, const unsigned char *data) {
...
+    if (secp256k1_ge_is_infinity(ge)) {
+        memset(data, 0, 64);
+    } else {
+        secp256k1_point_save(data, ge);
+    }
+}
+
+static void secp256k1_point_load_ext(secp256k1_ge *ge, const unsigned char *data) {
+    unsigned char zeros[64] = { 0 };
+    if (memcmp(data, zeros, sizeof(zeros)) == 0) {
```
real-or-random, line 60:
```suggestion
    if (secp256k1_memcmp_var(data, zeros, sizeof(zeros)) == 0) {
```

src/modules/extrakeys/tests_impl.h
```diff
@@ -643,103 +645,143 @@ void permute(size_t *arr, size_t n) {
...
+    sorted[3] = &pubkeys[1];
+    sorted[4] = &pubkeys[2];
+
+    for (i = 0; i < N_PUBKEYS; i++) {
+        CHECK(secp256k1_ec_pubkey_parse(ctx, &pubkeys[i], pk_ser[i], sizeof(pk_ser[i])));
+        pks_ptr[i] = &pubkeys[i];
+    }
+    CHECK(secp256k1_pubkey_sort(ctx, pks_ptr, N_PUBKEYS) == 1);
+    for (i = 0; i < N_PUBKEYS; i++) {
+        CHECK(memcmp(pks_ptr[i], sorted[i], sizeof(secp256k1_pubkey)) == 0);
```
real-or-random, line 765:
```suggestion
        CHECK(secp256k1_memcmp_var(pks_ptr[i], sorted[i], sizeof(secp256k1_pubkey)) == 0);
```

src/modules/musig/session_impl.h
```diff
@@ -74,12 +74,24 @@ static int secp256k1_musig_pubnonce_load(const secp256k1_context* ctx, secp256k1
...
 }
 
+static const unsigned char secp256k1_musig_aggnonce_magic[4] = { 0xa8, 0xb7, 0xe4, 0x67 };
+
 static void secp256k1_musig_aggnonce_save(secp256k1_musig_aggnonce* nonce, secp256k1_ge* ge) {
-    secp256k1_musig_pubnonce_save((secp256k1_musig_pubnonce *) nonce, ge);
+    int i;
+    memcpy(&nonce->data[0], secp256k1_musig_aggnonce_magic, 4);
+    for (i = 0; i < 2; i++) {
+        secp256k1_point_save_ext(&nonce->data[4+64*i], &ge[i]);
```
real-or-random, line 83:
```suggestion
        secp256k1_point_save_ext(&nonce->data[4 + 64*i], &ge[i]);
```
(just to be consistent with formatting in the function below)

src/modules/musig/session_impl.h
```diff
@@ -180,17 +192,69 @@ int secp256k1_musig_pubnonce_parse(const secp256k1_context* ctx, secp256k1_musig
...
         }
     }
-    /* The group elements can not be infinity because they were just parsed */
     secp256k1_musig_pubnonce_save(nonce, ge);
     return 1;
 }
 
+/* Outputs 33 zero bytes if the given group element is the point at infinity and
+ * otherwise outputs the compressed serialization */
+void secp256k1_ge_serialize_ext(unsigned char *out33, secp256k1_ge* ge) {
```
real-or-random, line 201:
```suggestion
static void secp256k1_ge_serialize_ext(unsigned char *out33, secp256k1_ge* ge) {
```
Well I think this should be done for every function not exposed in the public API but maybe it's better to postpone this to a PR that tidies the entire codebase... I suspect it's wrong in other modules too.

src/modules/musig/session_impl.h
```diff
@@ -418,13 +493,20 @@ static int secp256k1_musig_nonce_process_internal(int *fin_nonce_parity, unsigne
...
     secp256k1_scalar_set_b32(b, noncehash, NULL);
-    secp256k1_ecmult(&fin_nonce_ptj, &aggnoncej[1], b, NULL);
-    secp256k1_gej_add_ge(&fin_nonce_ptj, &fin_nonce_ptj, &aggnonce[0]);
+    secp256k1_gej_set_infinity(&fin_nonce_ptj);
+    if (!secp256k1_gej_is_infinity(&aggnoncej[1])) {
+        secp256k1_ecmult(&fin_nonce_ptj, &aggnoncej[1], b, NULL);
+    }
+    if (!secp256k1_gej_is_infinity(&aggnoncej[0])) {
+        secp256k1_gej_add_ge(&fin_nonce_ptj, &fin_nonce_ptj, &aggnonce[0]);
+    }
```
real-or-random, line 495: I would be surprised if `secp256k1_ecmult` or `secp256k1_gej_add_ge` can't handle infinity as an input. Or is this just supposed to be an optimization? If yes, I suggest to drop it because it optimizes the malicious case.

src/modules/musig/session_impl.h
```diff
@@ -333,6 +332,7 @@ static void secp256k1_nonce_function_musig(secp256k1_scalar *k, const unsigned c
     secp256k1_sha256 sha;
     unsigned char rand[32];
     unsigned char i;
+    unsigned char present;
```
real-or-random, line 315: maybe rename to `m_present`

real-or-random:
for reference:
- 0.2.0: https://github.com/jonasnick/bips/commit/aab260c37112b2010266fd8f0d3d56ea1111a2cc 
- 0.3.0: https://github.com/jonasnick/bips/commit/a3686a4bcce1879f68e975e5bd000ca2bd04cee5
- 0.4.0: https://github.com/jonasnick/bips/commit/20ba03106d245d42375e65f739c58dd9074daf6e
- 0.5.1: https://github.com/jonasnick/bips/commit/ea8ad301344b5f20de04b38c934a2532ff1618fa
- 0.6.0: https://github.com/jonasnick/bips/commit/298c7ce99bab3d07790d7d7a4f06b248871da3e8
  ("Change order of arguments (...) in the ''NonceGen'' hash function": is this a bug in the changelog? The order was not changed as far as I can tell)
- 0.7.0: https://github.com/jonasnick/bips/commit/60b806e18467759cf3f41e3ff94381751fe2eda5

real-or-random (commented):
reviewed up to 0.7.0 inclusive

src/modules/musig/session_impl.h
jonasnick: I don't think it was wrong. The code just predated the specification.

src/modules/musig/keyagg_impl.h
jonasnick: fixed

src/modules/musig/session_impl.h
jonasnick: fixed

src/modules/extrakeys/tests_impl.h
jonasnick: fixed

src/modules/musig/session_impl.h
jonasnick: I think that plus another function were the only offenders in the module. Fixed.

src/modules/musig/session_impl.h
jonasnick: done

src/modules/musig/session_impl.h
jonasnick: Ah thanks, I wanted to revisit this. `secp256k1_gej_add_ge` can't deal with infinity, but `secp256k1_gej_add_ge_var` can. Fixed.

include/secp256k1_extrakeys.h
```diff
@@ -64,21 +64,6 @@ SECP256K1_API int secp256k1_xonly_pubkey_serialize(
...
- *  Args: ctx:      a secp256k1 context object.
- *  In:   pubkey1:  first public key to compare
- *        pubkey2:  second public key to compare
- */
-SECP256K1_API int secp256k1_xonly_pubkey_cmp(
-    const secp256k1_context* ctx,
-    const secp256k1_xonly_pubkey* pk1,
-    const secp256k1_xonly_pubkey* pk2
-) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
-
```
real-or-random, line 81: I think we shouldn't remove stuff from extrakeys, which is an upstream module (this function is present upstream.).  Adding the sort function is already ugly enough -- maybe that can be cleaned up in the future.

src/modules/extrakeys/tests_impl.h
```diff
@@ -643,84 +645,82 @@ void permute(size_t *arr, size_t n) {
...
-    secp256k1_xonly_pubkey pk[5];
-    unsigned char pk_ser[5][32];
+void test_sort(void) {
+    secp256k1_pubkey pk[5];
+    unsigned char pk_ser[5][33] = {
+        { 0x02, 0x08 },
+        { 0x02, 0x0b },
+        { 0x02, 0x0c },
+        { 0x03, 0x05 },
+        { 0x03, 0x0a },
```
real-or-random, line 690: I know this is not this PR but I think It would really be nice to have some more variety here in the test cases here. For example, if you had forgotten to replace 32 by 33 in the comparison function in this PR, the tests wouldn't have caught this so maybe add at least a test case where the keys differ only the last byte. 

edit: ok now that I see the next commit, maybe we should add such a case to the test vectors... This will be better but this is certainly not in the scope of this PR. (So feel free to address this in test_sort or ignore it here and leave it leave entirely to an improved test vector.)

include/secp256k1_musig.h
```diff
@@ -190,7 +184,7 @@ SECP256K1_API int secp256k1_musig_partial_sig_parse(
  *
  *  Different orders of `pubkeys` result in different `agg_pk`s.
  *
- *  The pubkeys can be sorted before combining with `secp256k1_xonly_sort` which
+ *  The pubkeys can be sorted before combining with `secp256k1_pubkey_sort` which
```
real-or-random, line 187: nit:
```suggestion
 *  Before combining, the pubkeys can be sorted with `secp256k1_pubkey_sort`, which
```
My brain parsed this as "(The pubkeys can be sorted (before combining with `secp256k1_xonly_sort`)) which ..."

src/modules/musig/keyagg_impl.h
```diff
@@ -140,52 +159,58 @@ static void secp256k1_musig_keyaggcoef_sha256(secp256k1_sha256 *sha) {
...
+ * second_pk is the point at infinity in case there is no second_pk. Assumes
+ * that the coordinates of pk and second_pk are normalized. */
+static void secp256k1_musig_keyaggcoef_internal(secp256k1_scalar *r, const unsigned char *pk_hash, secp256k1_ge *pk, const secp256k1_ge *second_pk) {
     secp256k1_sha256 sha;
-    unsigned char buf[32];
 
-    if (secp256k1_fe_cmp_var(x, second_pk_x) == 0) {
+    if (!secp256k1_ge_is_infinity(second_pk)
+          && secp256k1_fe_equal(&pk->x, &second_pk->x)
+          && secp256k1_fe_is_odd(&pk->y) == secp256k1_fe_is_odd(&second_pk->y)) {
```
real-or-random, line 170: (not this PR:)
We could suggest a `ge_equal` function to upstream, that won't hurt... 

Maybe I'll do it after this PR has been merged.

src/modules/musig/keyagg_impl.h
```diff
@@ -140,52 +159,58 @@ static void secp256k1_musig_keyaggcoef_sha256(secp256k1_sha256 *sha) {
...
+        size_t buflen = sizeof(buf);
+        int ret;
         secp256k1_musig_keyaggcoef_sha256(&sha);
         secp256k1_sha256_write(&sha, pk_hash, 32);
-        secp256k1_fe_get_b32(buf, x);
-        secp256k1_sha256_write(&sha, buf, 32);
+        ret = secp256k1_eckey_pubkey_serialize(pk, buf, &buflen, 1);
+        /* Since the pubkeys were just loaded, it is impossible for a point to
+         * be at infinity. */
+        VERIFY_CHECK(ret && buflen == sizeof(buf));
```
real-or-random, line 181: True. Would it make sense add this precondition to the function docs?

real-or-random:
> The PR also does _not_ yet change the `secnonce` structure to also contain the signer's public key (which would also imply changing the seckey argument in `sign` to a keypair)

yes let's think about this separately after this PR.

real-or-random (commented):
changes that address my previous comments are all good

reviewed up to 1.0.0-rc2

jonasnick:
@real-or-random 

> 0.6.0: https://github.com/jonasnick/bips/commit/298c7ce99bab3d07790d7d7a4f06b248871da3e8
("Change order of arguments (...) in the ''NonceGen'' hash function": is this a bug in the changelog? The order was not changed as far as I can tell)

Yes this is a bug. Maybe happened because we initially hashed the message last to avoid a length prefix. Fixed [here](https://github.com/jonasnick/bips/pull/75).

include/secp256k1_extrakeys.h
jonasnick: Oh, I had forgotten that this is in upstream. Changed the commit to keep this function.

src/modules/extrakeys/tests_impl.h
jonasnick: Good point. I added a test vector to https://github.com/jonasnick/bips/pull/75 and also added it here.

include/secp256k1_musig.h
jonasnick: Done (and replaced "combine" with "aggregate")

src/modules/musig/keyagg_impl.h
jonasnick: Fwiw there's a `ge_equals_ge` for the tests.

src/modules/musig/keyagg_impl.h
jonasnick: Good idea. Also removed nonsensical note that the pubkey was just loaded.

src/modules/musig/tests_impl.h
```diff
@@ -921,6 +923,367 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+
+    create_key_agg_vector(&vector);
+
+    for (i = 0; i < (int)(sizeof(vector.valid_case)/sizeof(vector.valid_case[0])); i++) {
+        secp256k1_musig_keyagg_cache keyagg_cache;
+        unsigned char agg_pk[32];
+        struct musig_key_agg_valid_test_case *c = &vector.valid_case[i];
+
+        CHECK(musig_vectors_keyagg_and_tweak(&keyagg_cache, agg_pk, vector.pubkeys, vector.tweaks, c->key_indices_len, c->key_indices, 0, NULL, NULL));
+        CHECK(memcmp(agg_pk, c->expected, sizeof(agg_pk)) == 0);
```
real-or-random, line 992: secp256k1_memcmp_var (also in a few other places)

src/modules/musig/tests_impl.h
```diff
@@ -921,6 +923,367 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+
+    for (i = 0; i < (int)key_indices_len; i++) {
+        if (!secp256k1_ec_pubkey_parse(ctx, &pubkeys[i], pubkeys33[key_indices[i]], 33)) {
+            return 0;
+        }
+        pk_ptr[i] = &pubkeys[i];
+
+    }
+    if (!secp256k1_musig_pubkey_agg(ctx, NULL, NULL, keyagg_cache, pk_ptr, key_indices_len)) {
+        return 0;
```
real-or-random, line 949: is there a reason why you return 0 in this function (instead of CHECK)? CHECK makes it easier to spot the failing line in case of failure

real-or-random (commented):
Changes are good. H

Here are some nits. I'll try to have a closer look at the test vector commit later today

src/modules/musig/tests_impl.h
jonasnick: Yes, there is a reason. For some test vectors I want the function to fail (`CHECK(!musig_vectors_keyagg_and_tweak(...)`) and therefore do not want to abort the program.

src/modules/musig/tests_impl.h
jonasnick: Would it make sense to allow regular `memcmp` in the tests?

src/modules/musig/keyagg_impl.h
```diff
@@ -47,13 +47,30 @@ static void secp256k1_point_load(secp256k1_ge *ge, const unsigned char *data) {
...
+}
+
 static const unsigned char secp256k1_musig_keyagg_cache_magic[4] = { 0xf4, 0xad, 0xbb, 0xdf };
 
 /* A keyagg cache consists of
  * - 4 byte magic set during initialization to allow detecting an uninitialized
  *   object.
  * - 64 byte aggregate (and potentially tweaked) public key
- * - 32 byte X-coordinate of "second" public key (0 if not present)
+ * - 65 byte "second" public key (0 if not present)
```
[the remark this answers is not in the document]
jonasnick: Thanks! fixed.

src/modules/musig/tests_impl.h
jonasnick: I changed the function such that also error enum is set to distinguish the various error cases in keyagg_and_tweak.

real-or-random (commented):
Crazy script! :)

So far, I've only checked that the generated `vectors.h` matches the JSON.

Here's what I found:
 - `sign_verify_vectors`: I think the check in https://github.com/jonasnick/bips/blob/90133b00affd7d10389fbde42ada68ff08563e1e/bip-musig2/reference.py#L613 is missing in C code for the valid vectors? `nonce_indices` is missing entirely for those. Is this intentional? If yes, it should be documented (probably in the Python script and/or in its output). 
 - `sigagg_vectors`: Those also check nonce aggregation. Did you omit it on purpose? If yes, this should be documented. 
 - `keysort_test_vectors`: Those are missing. I think this is because they've been copied in the regular tests, but this should be documented somewhere.

Nits in `vectors.h`:
 - Maybe prefix `ERR_...` with `MUSIG_` because identifiers which "begin with E and a digit or E and an uppercase letter" are reserved for the standard library (https://port70.net/~nsz/c/c99/n1256.html#7.26.3)
 - Is there a reason why the `vectors.h` has functions instead of just static constants? (If the answer is that this makes the python script easier, then that's a good answer, just checking if you did this on purpose.)
 - (If we stick with functions,) is there a reason you use `memcpy` instead of struct assignment at the end of each function in `vectors.h`?
 - (If we stick with functions,) then the script should create `static` functions after https://github.com/bitcoin-core/secp256k1/pull/1190.

src/modules/musig/tests_impl.h
real-or-random: Maybe, but that's something that should be discussed upstream. I think if we believe that `memcmp` risky, then we should avoid it everywhere. But I know, it's very annoying, and I feel bad because I point this out in some many reviews. I guess we should have a check in CI...

jonasnick: I do think it makes a significant difference whether memcmp is used in tests or not. The only risk I can see with a memcmp like the one above is that miscompilation may hide an actual bug. But the miscompilation issue has been fixed since in GCC and at least you and I and apoelstra don't use the buggy GCC versions. Also CI runs almost all of the tests with also with clang.

jonasnick:
> So far, I've only checked that the generated vectors.h matches the JSON.

Thanks, good idea. I didn't really do that.

> sign_verify_vectors: I think the check in https://github.com/jonasnick/bips/blob/90133b00affd7d10389fbde42ada68ff08563e1e/bip-musig2/reference.py#L613 is missing in C code for the valid vectors? nonce_indices is missing entirely for those. Is this intentional? If yes, it should be documented (probably in the Python script and/or in its output).

Yes this is intentional. The `sign_verify` vectors are not intended to test nonce aggregation. We only added the individual nonces because some implementations may not allow for nonce aggregation outside of signing or verifying. Therefore, they need the individual nonces. This was requested [here](https://github.com/jonasnick/bips/issues/14). The check in the reference implementation just makes sure that the individual nonces are correct. Unfortunately, we can't (easily) add comments to the json.

> sigagg_vectors: Those also check nonce aggregation. Did you omit it on purpose? If yes, this should be documented.

Ommitted on purpose for the same reason as above. Where do you suggest this being documented? In the python reference code?

> keysort_test_vectors: Those are missing. I think this is because they've been copied in the regular tests, but this should be documented somewhere.

Ah, yes, done.

> Maybe prefix ERR_... with MUSIG_ because identifiers which "begin with E and a digit or E and an uppercase letter" are reserved for the standard library (https://port70.net/~nsz/c/c99/n1256.html#7.26.3)

Not sure about "reserved for the standard library". errno may define it, but we don't use errno, so it should be fine. And even if we did, we'd get a warning. `ERR_` looks nicer but I changed it nonetheless.

> Is there a reason why the vectors.h has functions instead of just static constants?

I hadn't put any thought into this. Static constants are indeed better. Added a fixup commit. Thanks.

real-or-random:
> Unfortunately, we can't (easily) add comments to the json.

> Where do you suggest this being documented? In the python reference code?

Ah no, sorry, what I had in mind is documenting that in the `musig2-vectors.py` (like you have a comment there explaining that we skip the vectors for messages which are not 32 bytes). I think it will be nice if the script is "complete" in the sense that for every piece of information in the JSON files, it either translates it or has a comment explaining why it's ignored. 


> > Maybe prefix ERR_... with MUSIG_ because identifiers which "begin with E and a digit or E and an uppercase letter" are reserved for the standard library ([port70.net/~nsz/c/c99/n1256.html#7.26.3](https://port70.net/~nsz/c/c99/n1256.html#7.26.3))
> 
> Not sure about "reserved for the standard library". errno may define it, but we don't use errno, so it should be fine. And even if we did, we'd get a warning. `ERR_` looks nicer but I changed it nonetheless.

Ok yes, I don't think it will ever be a problem, but I think we should try to stick to the rules in new code. See https://port70.net/~nsz/c/c99/n1256.html#7.26p1 and https://port70.net/~nsz/c/c99/n1256.html#7.1.3 for a definitive reference that these are reserved.

src/modules/musig/tests_impl.h
real-or-random: Yep, valid points. I'm not going to insist on this obviously, so either is fine here for now, but let's maybe restart this discussing in https://github.com/bitcoin-core/secp256k1/issues/823?

real-or-random: Yep, valid points. I'm not going to insist on this obviously, so either is fine here for now, but let's maybe restart this discussing in https://github.com/bitcoin-core/secp256k1/issues/823?

real-or-random: Yep, valid points. I'm not going to insist on this obviously, so either is fine here for now, but let's maybe restart this discussing in https://github.com/bitcoin-core/secp256k1/issues/823?

jonasnick:
rebased to fix CI

include/secp256k1_musig.h
```diff
@@ -430,6 +426,18 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_musig_nonce_process(
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
 
 /** Produces a partial signature
+ *
+ *  WARNING: It is crucial that the secret signing key corresponds to one of the
+ *  aggregated public keys. keys. Otherwise, the secret key can be compromised
```
real-or-random, line 431: keys. keys.

src/modules/musig/tests_impl.h
jonasnick: changed memcmp to secp256k1_memcmp_var

jonasnick:
@real-or-random 

> what I had in mind is documenting that in the musig2-vectors.py

done

src/modules/musig/tests_impl.h
```diff
@@ -921,6 +923,379 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+        {
+            /* In the last test vector we sign with an invalid secnonce, which
+             * triggers an illegal_callback. Hence, we need to use a custom
+             * context that does not abort in this case. */
+            secp256k1_context *ctx_tmp = secp256k1_context_create(SECP256K1_CONTEXT_NONE);
+            int ecount;
+            secp256k1_context_set_error_callback(ctx_tmp, counting_illegal_callback_fn, &ecount);
+            secp256k1_context_set_illegal_callback(ctx_tmp, counting_illegal_callback_fn, &ecount);
+            expected = c->error != MUSIG_SECNONCE;
+            CHECK(expected == secp256k1_musig_partial_sign(ctx_tmp, &partial_sig, &secnonce, &keypair, &keyagg_cache, &session));
```
real-or-random, line 1159: Maybe CHECK that `ecount` is indeed increased when `c->error != MUSIG_SECNONCE`.

nit: `ecount` should be `int32_t`.

src/modules/musig/tests_impl.h
```diff
@@ -921,6 +923,379 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+                                   const unsigned char pubkeys33[][33],
+                                   const unsigned char tweaks32[][32],
+                                   size_t key_indices_len,
+                                   const size_t *key_indices,
+                                   size_t tweak_indices_len,
+                                   const size_t *tweak_indices,
+                                   const int *is_xonly) {
+    secp256k1_pubkey pubkeys[MUSIG_VECTORS_MAX_PUBKEYS];
+    const secp256k1_pubkey *pk_ptr[MUSIG_VECTORS_MAX_PUBKEYS];
+    int i;
```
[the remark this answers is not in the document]
jonasnick: fixed

src/modules/musig/tests_impl.h
```diff
@@ -921,6 +923,379 @@ void musig_tweak_test(secp256k1_scratch_space *scratch) {
...
+    }
+    for (i = 0; i < (int)(sizeof(vector->error_case)/sizeof(vector->error_case[0])); i++) {
+        const struct musig_nonce_agg_test_case *c = &vector->error_case[i];
+        secp256k1_musig_pubnonce pubnonce[2];
+        for (j = 0; j < 2; j++) {
+            if (j == c->invalid_nonce_idx) {
+                CHECK(!secp256k1_musig_pubnonce_parse(ctx, &pubnonce[j], vector->pnonces[c->pnonce_indices[j]]));
+            } else {
+                CHECK(secp256k1_musig_pubnonce_parse(ctx, &pubnonce[j], vector->pnonces[c->pnonce_indices[j]]));
+            }
```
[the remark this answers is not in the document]

src/modules/musig/tests_impl.h

real-or-random (approved):
ACK a8dc6bee07c0b50ad98bbfd186c243ac541fa754 code review + tested every commit with valgrind

src/modules/musig/keyagg_impl.h
```diff
@@ -140,52 +159,59 @@ static void secp256k1_musig_keyaggcoef_sha256(secp256k1_sha256 *sha) {
...
+ * otherwise tagged_hash(pk_hash, x) where pk_hash is the hash of public keys.
+ * second_pk is the point at infinity in case there is no second_pk. Assumes
+ * that pk is not the point at infinity and that the coordinates of pk and
+ * second_pk are normalized. */
+static void secp256k1_musig_keyaggcoef_internal(secp256k1_scalar *r, const unsigned char *pk_hash, secp256k1_ge *pk, const secp256k1_ge *second_pk) {
     secp256k1_sha256 sha;
-    unsigned char buf[32];
 
-    if (secp256k1_fe_cmp_var(x, second_pk_x) == 0) {
+    if (!secp256k1_ge_is_infinity(second_pk)
```
ariard, line 169: If this an implementation of `KeyAggCoeffInternal`, I'm not sure the special-case on the second pubkey being the point at infinity is reflected in the latest state of the specification.

ariard (commented):
Just catching up with musig2 for some LN support, minor comments.

src/modules/musig/session_impl.h
```diff
@@ -226,53 +290,71 @@ static int secp256k1_xonly_ge_serialize(unsigned char *output32, secp256k1_ge *g
...
+    } else {
+        len = 0;
+        secp256k1_sha256_write(sha, &len, 1);
+    }
+}
+
+static void secp256k1_nonce_function_musig(secp256k1_scalar *k, const unsigned char *session_id, const unsigned char *msg32, const unsigned char *seckey32, const unsigned char *pk33, const unsigned char *agg_pk32, const unsigned char *extra_input32) {
     secp256k1_sha256 sha;
-    unsigned char seed[32];
+    unsigned char rand[32];
```
ariard, line 313: Can be called `masked_key` (as in schnorr module) to make it more evident it's for side-channel resistance.

include/secp256k1_extrakeys.h
```diff
@@ -256,6 +242,35 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_keypair_xonly_tweak_add
...
+ *  In:   pubkey1:  first public key to compare
+ *        pubkey2:  second public key to compare
+ */
+SECP256K1_API int secp256k1_pubkey_cmp(
+    const secp256k1_context* ctx,
+    const secp256k1_pubkey* pk1,
+    const secp256k1_pubkey* pk2
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Sorts public keys according to secp256k1_pubkey_cmp
```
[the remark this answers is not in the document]
jonasnick: Clarified the doc (we promise lexicographical order now).

src/modules/musig/keyagg_impl.h
jonasnick: It's not a special case. These three lines correspond to the line `If pk' = pk2:` in the BIP. In the code we can only compare x and y coordinate if the second pk is not the point at infinity.

src/modules/musig/keyagg_impl.h
```diff
@@ -47,13 +47,30 @@ static void secp256k1_point_load(secp256k1_ge *ge, const unsigned char *data) {
...
+}
+
 static const unsigned char secp256k1_musig_keyagg_cache_magic[4] = { 0xf4, 0xad, 0xbb, 0xdf };
 
 /* A keyagg cache consists of
  * - 4 byte magic set during initialization to allow detecting an uninitialized
  *   object.
  * - 64 byte aggregate (and potentially tweaked) public key
- * - 32 byte X-coordinate of "second" public key (0 if not present)
+ * - 64 byte "second" public key
```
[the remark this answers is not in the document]
jonasnick: done

src/modules/musig/keyagg_impl.h
```diff
@@ -140,52 +159,59 @@ static void secp256k1_musig_keyaggcoef_sha256(secp256k1_sha256 *sha) {
 }
 
 /* Compute KeyAgg coefficient which is constant 1 for the second pubkey and
- * tagged_hash(pk_hash, x) where pk_hash is the hash of public keys otherwise.
- * second_pk_x can be 0 in case there is no second_pk. Assumes both field
- * elements x and second_pk_x are normalized. */
-static void secp256k1_musig_keyaggcoef_internal(secp256k1_scalar *r, const unsigned char *pk_hash, const secp256k1_fe *x, const secp256k1_fe *second_pk_x) {
+ * otherwise tagged_hash(pk_hash, x) where pk_hash is the hash of public keys.
+ * second_pk is the point at infinity in case there is no second_pk. Assumes
+ * that pk is not the point at infinity and that the coordinates of pk and
```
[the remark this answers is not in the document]
jonasnick: A public key cannot be the point at infinity by definition (BIP 340). This is reflected in libsecp256k1 by `pubkey_load(secp256k1_ge group_elemnt, secp256k1_pubkey *pk)`, whose output `group_element` cannot be the point at infinity.

src/modules/musig/musig.md
```diff
@@ -23,16 +23,16 @@ Therefore, users of the musig module must take great care to make sure of the fo
 # Key Aggregation and (Taproot) Tweaking
 
 Given a set of public keys, the aggregate public key is computed with `secp256k1_musig_pubkey_agg`.
-A (Taproot) tweak can be added to the resulting public key with `secp256k1_xonly_pubkey_tweak_add` and an ordinary tweak can be added with `secp256k1_ec_pubkey_tweak_add`.
+A (Taproot) tweak can be added to the resulting public key with `secp256k1_xonly_pubkey_tweak_add` and an plain tweak can be added with `secp256k1_ec_pubkey_tweak_add`.
```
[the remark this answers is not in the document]
jonasnick: fixed

src/modules/musig/session_impl.h
```diff
@@ -74,12 +74,24 @@ static int secp256k1_musig_pubnonce_load(const secp256k1_context* ctx, secp256k1
     return 1;
 }
 
+static const unsigned char secp256k1_musig_aggnonce_magic[4] = { 0xa8, 0xb7, 0xe4, 0x67 };
```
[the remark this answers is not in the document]
jonasnick: `SECP256K1_SCHNORRSIG_EXTRAPARAMS_MAGIC` is special because the user of the API needs to know it. `secp256k1_musig_aggnonce_magic` on the other hand is internal.

src/modules/musig/session_impl.h
jonasnick: It's called `rand` to match the variable `rand` in the musig spec. `masked_key` isn't really precise because if no seckey is provided, then the variable doesn't contain the masked key.

jonasnick:
Thanks @ariard for your comments.

src/modules/musig/keyagg_impl.h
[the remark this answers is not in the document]
ariard: Correct, should have said a curve point. Understand `pubkey_load()` must be called before `secp256k1_musig_keyaggcoef_internal()`.

ariard (approved):
Light Code Review ACK b43dd83b, mostly looks on how the user API will make sense for Lightning, thanks for the answers!

For LN, for each channel/state/HTLCs they're are a bunch of unique identifiers such as channel_id, per_commitment_point, htlc_id that we've already well-defined in the protocol. I don't know how they could be used safely as additional contextual data that changes at each `secp256k1_musig_nonce_gen()`. They're shared with the counterparty and deterministic and could be batched to save on round-trips at the channel-level I think. Maybe to save signing latency.

jonasnick:
@ariard The main scenario where it makes sense to add data as`extra_input` to `musig_nonce_gen` is when it is likely that the data is different in each `musig_nonce_gen` call. If that's the case then there's protection against accidentally generating the same nonce if the `session_id32` is repeated in different `musig_nonce_gen` calls. This defense-in-depth measure works even if the `extra_input32` is public.

ariard:
@jonasnick Thanks - Well with channel the `htlc_id` is different at each states updates so I think the data is different with each `musig_nonce_gen()` call if you like to use the `extra_input`.

include/secp256k1_extrakeys.h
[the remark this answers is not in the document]
real-or-random: It says "lexicographic order" now, but should probably say "lexicographic order of their compressed 33-byte serialization" or something like this. But this can be fixed in a separate PR, lets' get this merged.

real-or-random (approved):
ACK b43dd83b43eac0ca8ad9ee1f557e9126c9e08d9e

open nits can be addressed separately

PR #218, opened 2023-03-01
Norm arg: allow X and R to be point at infinity

jonasnick:
This PR allows X and R to be the point at infinity in norm argument proving and verification. With that change, the verification algorithm should be equivalent to the [rust-bulletproofs-pp](https://github.com/sanket1729/rust-bulletproofs-pp) implementation.

This PR also adds other static test vectors for verify edge cases (corresponding to rust-bulletproofs-pp master)

- [x] The PR is a work-in-progress because it uses `ge_serialize_ext` and `ge_parse_ext` that are introduced in open PR #211


jonasnick:
rebased

PR #224, opened 2023-04-11
Backport of "ct: Use volatile "trick" in all fe/scalar cmov implementations"

real-or-random:
I tried to use our sync script, but it tries to merge the tree, so it's not appropriate for backports...


jonasnick (commented):
ACK 96f48538503ff40bf0017652c4b4f3a42cd3fa94

PR #225, opened 2023-04-11
sync-upstream: Use --autostash to handle uncommitted changes

real-or-random:
This makes it possible to use sync-upstream with uncommitted changes. (This is in particular helpful when working on the script itself.)

Without this commit, git pull will fail due to the uncommitted changes.


apoelstra:
utACK 13c438cdeed358a20b1f0324ee36a6cadfaf0016

PR #226, opened 2023-04-19
bppp: align terminology with paper

jonasnick:
Solves #221


real-or-random (approved):
utACK  2c63d17c1e3b89950af9b6e51b14f025a2251c95

jonasnick:
I think that'll solve #221

PR #227, opened 2023-04-21
Use relative #include paths and tidy header guards (as in upstream)

### LINKED ISSUE #208: Fix backport of bitcoin-core/secp256k1#925 (drop 'include/' prefix from -I)
I have a few questions about the compilation of the project...

After running autogen and configure with the custom flags, at that point is it possible to compile the project without using the newly generated makefile (aka use directly the `gcc` command) or something like that.

I'm trying also to compile the project from xcode because I have the need to use this code in a swift library... the main problems encountered with the compilation from xcode are wrong includes in the files. I'm not an expert on AutoMake, but from what I have understood, all the includes are managed by automake, is there a way to do it manually?

real-or-random:
Fixes #208.

This will make it much easier to sync with upstream.

After `./configure --enable-dev-mode`, you can compile manually with `clang -c src/secp256k1.c src/precomputed_*.c -DHAVE_CONFIG_H` to check that the paths are good, at least for the library. I manually inspected `git grep '#include "'` for the rest...

Unfortunately, just removing the `-I` arguments from the Makefile (as in https://github.com/bitcoin-core/secp256k1/pull/1116/commits/40a3473a9d44dc409412e94f70ad0f09bd9da3ac) is not enough to test this because automake adds some standards includes that we can't get rid of. Once we sync with upstream, running a CMake build is a direct way to test this.

### DIFF
--- a/include/secp256k1_bppp.h
+++ b/include/secp256k1_bppp.h
@@ -1,5 +1,5 @@
-#ifndef _SECP256K1_BPPP_
-# define _SECP256K1_BPPP_
+#ifndef SECP256K1_BPPP_H
+# define SECP256K1_BPPP_H
 
 # include "secp256k1.h"
 
--- a/include/secp256k1_generator.h
+++ b/include/secp256k1_generator.h
@@ -1,5 +1,5 @@
-#ifndef _SECP256K1_GENERATOR_
-# define _SECP256K1_GENERATOR_
+#ifndef SECP256K1_GENERATOR_H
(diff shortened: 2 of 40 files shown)


apoelstra (approved):
ACK e444d24bcad57091746784fcea6d07e95d058cd3

jonasnick (commented):
ACK e444d24bcad57091746784fcea6d07e95d058cd3

PR #230, opened 2023-04-25
norm arg: add prove test vectors

jonasnick:
Based on #218. The test vectors were generated with https://github.com/sanket1729/rust-bulletproofs-pp.


Liam-Eagen:
ACK f3126fd

PR #232, opened 2023-05-14
Backports from libsecp256k1 v0.3.2

jonasnick:


real-or-random (approved):
ACK 39407c3f5999aa10e1470bc9eae8f63800a63e51 I also verified that the ct time tests pass on GCC 13.1.1 and Clang 15.0.7.

PR #242, opened 2023-07-17
sync-upstream.sh: Fix position of "-b" option in reproduce command

real-or-random:


jonasnick (approved):
utACK 9b6a1c384d00cc61bf9160e5df858ca6c52e6af3

PR #248, opened 2023-07-20
bppp: Fix and improve test for invalid sign byte

real-or-random:
The test is supposed to create an invalid sign byte. Before this PR, the generated sign byte could in fact be valid due to an overflow.

### DIFF
--- a/src/modules/bppp/tests_impl.h
+++ b/src/modules/bppp/tests_impl.h
@@ -257,7 +257,11 @@ void test_serialize_two_points(void) {
         random_group_element_test(&X);
         random_group_element_test(&R);
         secp256k1_bppp_serialize_points(buf, &X, &R);
-        buf[0] |= 4 + (unsigned char)secp256k1_testrandi64(4, 255);
+
+        buf[0] = 4 + (unsigned char)secp256k1_testrandi64(0, 253);
+        /* Assert that buf[0] is actually invalid. */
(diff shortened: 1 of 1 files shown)


src/modules/bppp/tests_impl.h
```diff
@@ -254,10 +254,14 @@ void test_serialize_two_points(void) {
...
         unsigned char buf[65];
+        unsigned char invalid_sign;
         random_group_element_test(&X);
         random_group_element_test(&R);
         secp256k1_bppp_serialize_points(buf, &X, &R);
-        buf[0] |= 4 + (unsigned char)secp256k1_testrandi64(4, 255);
+        do {
+            invalid_sign = (unsigned char)secp256k1_testrand256();
+        } while(invalid_sign == 0x02 || invalid_sign == 0x03);
+        buf[0] = invalid_sign;
```
jonasnick, line 264: why not just `buf[0] = 4 + (unsigned char)secp256k1_testrandi64(0, 253);`?

real-or-random: Okay yes, this has the same effect...

real-or-random: done

jonasnick (approved):
ACK 3372993060bfe914d828e13499fcd2763ef83dd5

PR #249, opened 2023-07-20
Upstream PRs 1160, 1193, 1169, 1190, 1192, 1194, 1196, 1195, 1170, 1172, 1200, 1199, 1203, 1201, 1206, 1078, 1209, 979, 1212, 1218, 1217, 1221, 1222

jonasnick:
[bitcoin-core/secp256k1#1160]: Makefile: add `-I$(top_srcdir)/{include,src}` to `CPPFLAGS` for precomputed
[bitcoin-core/secp256k1#1193]: Add `noverify_tests` to `.gitignore`
[bitcoin-core/secp256k1#1169]: Add support for msan instead of valgrind (for memcheck and ctime test)
[bitcoin-core/secp256k1#1190]: Make all non-API functions (except main) static
[bitcoin-core/secp256k1#1192]: Switch to exhaustive groups with small B coefficient
[bitcoin-core/secp256k1#1194]: Ensure safety of ctz_debruijn implementation.
[bitcoin-core/secp256k1#1196]: Drop no longer used variables from the build system
[bitcoin-core/secp256k1#1195]: ctime_tests: improve output when CHECKMEM_RUNNING is not defined
[bitcoin-core/secp256k1#1170]: contexts: Forbid destroying, cloning and randomizing the static context
[bitcoin-core/secp256k1#1172]: benchmarks: fix bench_scalar_split
[bitcoin-core/secp256k1#1200]: Drop no longer used Autoheader macros
[bitcoin-core/secp256k1#1199]: ci: Minor improvements inspired by Bitcoin Core
[bitcoin-core/secp256k1#1203]: Do not link `bench` and `ctime_tests` to `COMMON_LIB`
[bitcoin-core/secp256k1#1201]: ci: Do not set git's `user.{email,name}` config options
[bitcoin-core/secp256k1#1206]: build: Add -Wreserved-identifier supported by clang
[bitcoin-core/secp256k1#1078]: group: Save a normalize_to_zero in gej_add_ge
[bitcoin-core/secp256k1#1209]: build: Add SECP256K1_API_VAR to fix importing variables from DLLs
[bitcoin-core/secp256k1#979]: Native jacobi symbol algorithm
[bitcoin-core/secp256k1#1212]: Prevent dead-store elimination when clearing secrets in examples
[bitcoin-core/secp256k1#1218]: Update overflow check
[bitcoin-core/secp256k1#1217]: Add secp256k1_fe_add_int function
[bitcoin-core/secp256k1#1221]: Update Changelog
[...]

PR #250, opened 2023-07-20
ci: Always define EXPERIMENTAL variable

real-or-random:
See the failure here:

https://github.com/BlockstreamResearch/secp256k1-zkp/pull/249/checks?check_run_id=15209841852

I suggest merging this PR and then rebasing sync-upstream on master (and then #249 on sync-upstream). This will give us a bit more meaningful CI output for now when working on sync-upstream, even though the Linux tasks still don't work.


jonasnick (approved):
ACK 7e9193666f840eaec024d1dad968361b79a14753

PR #255, opened 2023-07-22
musig: change test vector generation code shebang from python to python3

jonasnick:
The linter included in the Bitcoin Core and Elements test framework requires python3.


real-or-random (approved):
utACK ea478beec666dc38729f4927a005875f079fa914

PR #258, opened 2023-07-27
lint: fix python open() to use encoding="utf8" in musig2-vectors.py

delta1:
In [elements#1248](ElementsProject/elements#1248) we are updating to the current latest `secp256k1-zkp`. 

The linter from upstream `lint-python-utf8-encoding.sh` requires that calls to `open(...)` explicitly specify the encoding, which is failing for the file `musig2-vectors.py`

### DIFF
--- a/contrib/musig2-vectors.py
+++ b/contrib/musig2-vectors.py
@@ -94,7 +94,7 @@ enum MUSIG_ERROR {
 """
 
 # key agg vectors
-with open(sys.argv[1] + "/key_agg_vectors.json", "r") as f:
+with open(sys.argv[1] + "/key_agg_vectors.json", "r", encoding="utf8") as f:
     data = json.load(f)
 
     max_key_indices = max(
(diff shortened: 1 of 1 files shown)


delta1:
CI error seems unrelated?

real-or-random:
> In [elements#1248](ElementsProject/elements#1248) we are updating to the current latest `secp256k1-zkp`.
> 
> The linter from upstream `lint-python-utf8-encoding.sh` requires that calls to `open(...)` explicitly specify the encoding, which is failing for the file `musig2-vectors.py`

The JSON files are pure ASCII, so encoding shouldn't matter. But happy to make that change if it makes the linter happy. 

Yep, please ignore CI. We're actively working on updating master to incorporate tons of changes from upstream secp256k1. That will fix CI. And perhaps it's worth updating to current secp256k1-zkp after we're done with this in a few days, or just wait a few days (but then the diff will be even larger, so it may make sense to do it in two steps).

real-or-random (approved):
utACK 10bde4734b1cb5b032e0d0e5dcd5a5bc3900b7bd

delta1:
Thanks @real-or-random

jonasnick:
@delta1 I wonder if it'd make more sense to exclude the file from the linter. It's just a script that translates the JSON test vectors into C code. It's not part of the build process or the tests. It seems a bit arbitrary that we need to conform to the (partially non-sensical) requirements of a linter that's not even part of secp-zkp CI.

Is this the last linter complaint?

PR #130, opened 2021-05-04
Add Schnorrsig half aggregation

jonasnick:
Result of a recorded pair-programming session with @real-or-random to demonstrate how to write libsecp(-zkp) code. Therefore the code is still in a very early stage.

PR #261, opened 2023-07-30
Schnorr (Incremental) Half Aggregation 

b-wagn:
Revisited PR #130 by jonasnick.
I am happy to hear your thoughts.

**Summary of changes compared to #130:** 

- Address comments from rustyrussell 
- Use tagged hash
- Compute hashes with common prefix by copying midstate
- Allow Incremental Aggregation and make code consistent with the [draft spec](https://github.com/BlockstreamResearch/cross-input-aggregation/blob/master/half-aggregation.mediawiki)

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -184,6 +184,10 @@ AC_ARG_ENABLE(module_schnorrsig,
     AS_HELP_STRING([--enable-module-schnorrsig],[enable schnorrsig module [default=yes]]), [],
     [SECP_SET_DEFAULT([enable_module_schnorrsig], [yes], [yes])])
 
+AC_ARG_ENABLE(module_schnorrsig_halfagg,
+    AS_HELP_STRING([--enable-module-schnorrsig-halfagg],[enable schnorrsig half-aggregation module (experimental) [default=no]]), [],
+    [SECP_SET_DEFAULT([enable_module_schnorrsig_halfagg], [no], [yes])])
+
 AC_ARG_ENABLE(module_ellswift,
     AS_HELP_STRING([--enable-module-ellswift],[enable ElligatorSwift module [default=yes]]), [],
     [SECP_SET_DEFAULT([enable_module_ellswift], [yes], [yes])])
@@ -445,6 +449,11 @@ SECP_CFLAGS="$SECP_CFLAGS $WERROR_CFLAGS"
 
 # Processing must be done in a reverse topological sorting of the dependency graph
 # (dependent module first).
+if test x"$enable_module_schnorrsig_halfagg" = x"yes"; then
+  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_SCHNORRSIG_HALFAGG=1"
+  enable_module_schnorrsig=yes
+fi
+
 if test x"$enable_module_bppp" = x"yes"; then
   if test x"$enable_module_generator" = x"no"; then
     AC_MSG_ERROR([Module dependency error: You have disabled the generator module explicitly, but it is required by the bppp module.])
@@ -497,7 +506,6 @@ if test x"$enable_module_generator" = x"yes"; then
   SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_GENERATOR=1"
 fi
 
-
 if test x"$enable_module_ellswift" = x"yes"; then
   SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_ELLSWIFT=1"
 fi
@@ -544,6 +552,9 @@ else
   # module (which automatically enables the module dependencies) we want to
   # print an error for the dependent module, not the module dependency. Hence,
   # we first test dependent modules.
+  if test x"$enable_module_schnorrsig_halfagg" = x"yes"; then
+    AC_MSG_ERROR([Schnorrsig Half-Aggregation module is experimental. Use --enable-experimental to allow.])
+  fi
   if test x"$enable_module_bppp" = x"yes"; then
     AC_MSG_ERROR([Bulletproofs++ module is experimental. Use --enable-experimental to allow.])
   fi
@@ -599,6 +610,7 @@ AM_CONDITIONAL([ENABLE_MODULE_MUSIG], [test x"$enable_module_musig" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_ECDSA_S2C], [test x"$enable_module_ecdsa_s2c" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_ECDSA_ADAPTOR], [test x"$enable_module_ecdsa_adaptor" = x"yes"])
 AM_CONDITIONAL([ENABLE_MODULE_BPPP], [test x"$enable_module_bppp" = x"yes"])
+AM_CONDITIONAL([ENABLE_MODULE_SCHNORRSIG_HALFAGG], [test x"$enable_module_schnorrsig_halfagg" = x"yes"])
 AM_CONDITIONAL([USE_REDUCED_SURJECTION_PROOF_SIZE], [test x"$use_reduced_surjection_proof_size" = x"yes"])
 AM_CONDITIONAL([USE_EXTERNAL_ASM], [test x"$enable_external_asm" = x"yes"])
 AM_CONDITIONAL([USE_ASM_ARM], [test x"$set_asm" = x"arm32"])
@@ -638,18 +650,19 @@ echo "  module musig            = $enable_module_musig"
 echo "  module ecdsa-s2c        = $enable_module_ecdsa_s2c"
 echo "  module ecdsa-adaptor    = $enable_module_ecdsa_adaptor"
 echo "  module bppp             = $enable_module_bppp"
+echo "  module schnorrsig-halfagg = $enable_module_schnorrsig_halfagg"
 echo
-echo "  asm                     = $set_asm"
-echo "  ecmult window size      = $set_ecmult_window"
-echo "  ecmult gen prec. bits   = $set_ecmult_gen_precision"
+echo "  asm                       = $set_asm"
+echo "  ecmult window size        = $set_ecmult_window"
+echo "  ecmult gen prec. bits     = $set_ecmult_gen_precision"
 # Hide test-only options unless they're used.
 if test x"$set_widemul" != xauto; then
-echo "  wide multiplication     = $set_widemul"
+echo "  wide multiplication       = $set_widemul"
 fi
 echo
-echo "  valgrind                = $enable_valgrind"
-echo "  CC                      = $CC"
-echo "  CPPFLAGS                = $CPPFLAGS"
-echo "  SECP_CFLAGS             = $SECP_CFLAGS"
-echo "  CFLAGS                  = $CFLAGS"
-echo "  LDFLAGS                 = $LDFLAGS"
+echo "  valgrind                  = $enable_valgrind"
+echo "  CC                        = $CC"
+echo "  CPPFLAGS                  = $CPPFLAGS"
+echo "  SECP_CFLAGS               = $SECP_CFLAGS"
+echo "  CFLAGS                    = $CFLAGS"
+echo "  LDFLAGS                   = $LDFLAGS"
--- a/include/secp256k1_schnorrsig_halfagg.h
+++ b/include/secp256k1_schnorrsig_halfagg.h
@@ -1,0 +1,107 @@
+#ifndef SECP256K1_SCHNORRSIG_HALFAGG_H
+#define SECP256K1_SCHNORRSIG_HALFAGG_H
+
+#include "secp256k1.h"
+#include "secp256k1_extrakeys.h"
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+/** Incrementally (Half-)Aggregate a sequence of Schnorr
+ *  signatures to an existing half-aggregate signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In/Out:      aggsig: pointer to the serialized aggregate signature
+ *                       that is input. The first 32*(n_before+1) of this
+ *                       array should hold the input aggsig. It will be
+ *                       overwritten by the new serialized aggregate signature.
+ *                       It should be large enough for that, see aggsig_len.
+ *           aggsig_len: size of aggsig array in bytes.
+ *                       Should be large enough to hold the new
+ *                       serialized aggregate signature, i.e.,
+ *                       should satisfy aggsig_size >= 32*(n_before+n_new+1).
+ *                       It will be overwritten to be the exact size of the
+ *                       resulting aggsig.
+ *  In:     all_pubkeys: Array of (n_before + n_new) many x-only public keys,
+ *                       including both the ones for the already aggregated signature
+ *                       and the ones for the signatures that should be added.
+ *                       Can only be NULL if n_before + n_new is 0.
+ *           all_msgs32: Array of (n_before + n_new) many 32-byte messages,
+ *                       including both the ones for the already aggregated signature
+ *                       and the ones for the signatures that should be added.
+ *                       Can only be NULL if n_before + n_new is 0.
+ *           new_sigs64: Array of n_new many 64-byte signatures, containing the new
+ *                       signatures that should be added. Can only be NULL if n_new is 0.
+ *             n_before: Number of signatures that have already been aggregated
+ *                       in the input aggregate signature.
+ *                n_new: Number of signatures that should now be added
+ *                       to the aggregate signature.
+ */
+SECP256K1_API int secp256k1_schnorrsig_inc_aggregate(
+    const secp256k1_context *ctx,
+    unsigned char *aggsig,
+    size_t *aggsig_len,
+    const secp256k1_xonly_pubkey* all_pubkeys,
+    const unsigned char *all_msgs32,
+    const unsigned char *new_sigs64,
+    size_t n_before,
+    size_t n_new
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** (Half-)Aggregate a sequence of Schnorr signatures.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:           ctx: a secp256k1 context object.
+ *  Out:         aggsig: pointer to an array of aggsig_len many bytes to
+ *                       store the serialized aggregate signature.
+ *  In/Out:  aggsig_len: size of the aggsig array that is passed in bytes;
+ *                       will be overwritten to be the exact size of aggsig.
+ *  In:         pubkeys: Array of n many x-only public keys.
+ *                       Can only be NULL if n is 0.
+ *               msgs32: Array of n many 32-byte messages.
+ *                       Can only be NULL if n is 0.
+ *               sigs64: Array of n many 64-byte signatures.
+ *                       Can only be NULL if n is 0.
+ *                    n: number of signatures to be aggregated.
+ */
+SECP256K1_API int secp256k1_schnorrsig_aggregate(
+    const secp256k1_context *ctx,
+    unsigned char *aggsig,
+    size_t *aggsig_len,
+    const secp256k1_xonly_pubkey *pubkeys,
+    const unsigned char *msgs32,
+    const unsigned char *sigs64,
+    size_t n
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Verify a (Half-)aggregate Schnorr signature.
+ *
+ *  Returns:          1: correct signature.
+ *                    0: incorrect signature.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In:         pubkeys: Array of n many x-only public keys. Can only be NULL if n is 0.
+ *               msgs32: Array of n many 32-byte messages. Can only be NULL if n is 0.
+ *                    n: number of signatures to that have been aggregated.
+ *               aggsig: Pointer to an array of aggsig_size many bytes
+ *                       containing the serialized aggregate
+ *                       signature to be verified.
+ *           aggsig_len: Size of the aggregate signature in bytes.
+ *                       Should be aggsig_len = 32*(n+1)
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_aggverify(
+    const secp256k1_context *ctx,
+    const secp256k1_xonly_pubkey *pubkeys,
+    const unsigned char *msgs32,
+    size_t n,
+    const unsigned char *aggsig,
+    size_t aggsig_len
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(5);
+
+#ifdef __cplusplus
+}
+#endif
+
+#endif /* SECP256K1_SCHNORRSIG_HALFAGG_H */
--- a/src/modules/schnorrsig_halfagg/Makefile.am.include
+++ b/src/modules/schnorrsig_halfagg/Makefile.am.include
@@ -1,0 +1,3 @@
+include_HEADERS += include/secp256k1_schnorrsig_halfagg.h
+noinst_HEADERS += src/modules/schnorrsig_halfagg/main_impl.h
+noinst_HEADERS += src/modules/schnorrsig_halfagg/tests_impl.h
--- a/src/modules/schnorrsig_halfagg/tests_impl.h
+++ b/src/modules/schnorrsig_halfagg/tests_impl.h
@@ -1,0 +1,338 @@
+#ifndef SECP256K1_MODULE_SCHNORRSIG_HALFAGG_TESTS_H
+#define SECP256K1_MODULE_SCHNORRSIG_HALFAGG_TESTS_H
+
+#include "../../../include/secp256k1_schnorrsig_halfagg.h"
+
+#define N_MAX 50
+
+/* We test that the hash initialized by secp256k1_schnorrsig_sha256_tagged_aggregate
+ * has the expected state. */
+void test_schnorrsig_sha256_tagged_aggregate(void) {
+    unsigned char tag[18] = "HalfAgg/randomizer";
+    secp256k1_sha256 sha;
+    secp256k1_sha256 sha_optimized;
+
+    secp256k1_sha256_initialize_tagged(&sha, (unsigned char *) tag, sizeof(tag));
+    secp256k1_schnorrsig_sha256_tagged_aggregation(&sha_optimized);
+    test_sha256_eq(&sha, &sha_optimized);
+}
+
+/* Create n many x-only pubkeys and sigs for random messages */
+void test_schnorrsig_aggregate_input_helper(secp256k1_xonly_pubkey *pubkeys, unsigned char *msgs32, unsigned char *sigs64, size_t n) {
+    size_t i;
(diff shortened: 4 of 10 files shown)


jonasnick (commented):
Cool, thanks @b-wagn.

Can you remove the added vscode stuff from the commits? This would allow to getting rid of the two "remove vscode settings" commits.

In order to verify that this matches the spec, it would help to add the preliminary [test vectors](https://github.com/BlockstreamResearch/cross-input-aggregation/blob/master/hacspec-halfagg/tests/tests.rs#L78) (they don't seem to cover incremental aggregation yet).

include/secp256k1_schnorrsig.h
```diff
@@ -175,6 +175,81 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_verify(
...
+
+
+/** Verify a (Half-)aggregate Schnorr signature.
+ *
+ *  Returns:          1: correct signature
+ *                    2: incorrect signature.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In:         pubkeys: Array of x-only public keys. Assume to contain n_sigs many public keys.
+ *               msgs32: Array of 32-byte messages. Assumed to contain n_sigs many messages.
+ *               n_sigs: number of signatures to that have been aggregated.
```
real-or-random, line 239: nit:
```suggestion
 *               n: number of public keys and messages
```

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,291 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+static int compute_aggregation_coefficients(
+    secp256k1_scalar* zs,
+    const unsigned char* pubkeys_ser, 
+    const unsigned char** rs,
+    const unsigned char* msgs32, 
+    size_t starting_from,
+    size_t n_sigs) {
+
+
+    uint32_t i;
```
real-or-random, line 303: `size_t`?

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,291 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+    /* zi = Hash( pre || r_{n'} || pk_{n'} || mes_{n'} || ... || r_{i} || pk_{i} || mes_{i} ) */
+    for (i = starting_from; i < n_sigs; ++i) {
+        unsigned char hashoutput[32];
+        secp256k1_sha256 hashcopy; 
+        /* Write into hash r_i, pk_i, mes_i, r_i */
+        secp256k1_sha256_write(&hash, rs[i], 32);
+        secp256k1_sha256_write(&hash, &pubkeys_ser[i*32], 32);
+        secp256k1_sha256_write(&hash, &msgs32[i*32], 32);
+        /* Copy the hash */
+        memcpy(&hashcopy,&hash,sizeof(hash));
```
real-or-random, line 332:
```suggestion
        hashcopy = hash;
```

structs can be assigned in C ;)

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,291 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+    for (i = 0; i < n_sigs; ++i) {
+        if (!secp256k1_xonly_pubkey_serialize(ctx, &pubkeys_ser[i*32], &all_pubkeys[i])) {
+            free(pubkeys_ser);
+            return 0;
+        }
+    }
+
+    /* Compute new aggregation coefficients zi*/
+    zs = (secp256k1_scalar*) checked_malloc(&ctx->error_callback, n_sigs_new*sizeof(secp256k1_scalar));
+    rs = (const unsigned char**) checked_malloc(&ctx->error_callback, n_sigs*sizeof(unsigned char*));
```
real-or-random, line 376: Oh, hm, so it's a policy (like in secp256k1) that we don't call malloc/free (except for context creation, but this can also be done in a block of memory allocated by the caller, see the `secp256k1_prealloc.h` header). The reasons are safety and support for embedded systems.

I haven't thought this through entirely, but I think all algorithms in the spec can be implemented without dynamic allocation (i.e., in O(1) stack memory), by rearranging the loops. 

At least this is true as long as we don't multi-multiplications. For multi-multiplications, we'll necessarily need some form of memory management. In upstream secp256k1, we have a special data type, but it's unused there, and we may change this in the future, because we're not particularly happy with it. See https://github.com/bitcoin-core/secp256k1/issues/1302.

So I tend to suggest rewriting the loops (if possible) and ignoring multi-multiplications for now.  @jonasnick What do you think?

real-or-random (commented):
Here are some comments but I haven't had a closer look yet.

This also needs trivial rebase on master as we made quite a lot of changes there (synced with upstream secp256k1).

b-wagn:
Thanks for your feedback.
I will provide an updated version asap.

.gitignore
```diff
@@ -23,6 +23,8 @@ schnorr_example
 *.trs
 *.sage.py
 
+.vscode/
+
```
real-or-random, line 27: nit: I think this is something you should set on your local machine, either just for this repo, or for all repos, see https://docs.github.com/en/get-started/getting-started-with-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer for instructions for both variants.

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(aggsig != NULL);
+    ARG_CHECK(aggsig_size != NULL);
+    ARG_CHECK(all_pubkeys != NULL);
+    ARG_CHECK(all_msgs32 != NULL);
+    ARG_CHECK(new_sigs64 != NULL);
+
+    /* Check that aggsig_size is large enough, i.e. aggsig_size >= 32*(n+1) */
+    n = n_before + n_new;
```
real-or-random, line 300: I think this should `return 0` also when the addition here overflows. (I mean, you could say then anyway, something is wrong somewhere else, because the user would think they pass an array of size larger than `size_t`...) I just think it's more obviously correct (and thus easier to review with the overflow check.)


On a second thought, I wonder whether this should be an `ARG_CHECK` instead of `return 0`. Not sure, we'll need to think about it.

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+    if ((*aggsig_size >> 5) <= 0 || ((*aggsig_size >> 5)-1) < n) {
```
real-or-random, line 301: nit: I think this is a bit more natural to read (but yeah, it's maybe just taste in the end):
```suggestion
    if ((*aggsig_size / 32) == 0 || ((*aggsig_size / 32)-1) < n) {
```

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+        /* serialize pk_i */
+        unsigned char pk_ser[32];
+        if (!secp256k1_xonly_pubkey_serialize(ctx, pk_ser, &all_pubkeys[i])) {
+            return 0;
+        }
+        /* write r_i */
+        secp256k1_sha256_write(&hash, &aggsig[i*32], 32);
+        /* write pk_i */
+        secp256k1_sha256_write(&hash, pk_ser, 32);
+        /* write mes_i*/
```
real-or-random, line 319:
```suggestion
        /* write m_i*/
```

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+        secp256k1_sha256_write(&hash, pk_ser, 32);
+        secp256k1_sha256_write(&hash, &all_msgs32[i*32], 32);
+        /* 1.b) Copy the hash */
+        hashcopy = hash;
+        /* 1.c) Finalize the copy to get zi*/
+        secp256k1_sha256_finalize(&hashcopy, hashoutput);
+        secp256k1_scalar_set_b32(&zi, hashoutput, NULL);
+
+        /* Step2: s := s + zi*si */
+        secp256k1_scalar_set_b32(&si, &new_sigs64[(i-n_before)*64+32], NULL);
```
real-or-random, line 356: I wonder if this should really ignore the overflow. Overflow should never happen here for a valid input sig, so it's probably stricter to bail out. (I see that this will be a change to the spec. cc @jonasnick )

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+    secp256k1_sha256 hash;
+    int overflow;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(pubkeys != NULL);
+    ARG_CHECK(msgs32 != NULL);
+    ARG_CHECK(aggsig != NULL);
+
+    /* Check that aggsig_size is correct, i.e. aggsig_size = 32*(n+1) */
+    if ((aggsig_size >> 5) <= 0 || ((aggsig_size >> 5)-1) != n) {
```
real-or-random, line 391: same as above

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+        /* Step 1: z_i = TaggedHash(...) */
+        /* 1.a) Write into hash r_i, pk_i, mes_i, r_i */
+        secp256k1_sha256_write(&hash, &aggsig[i*32], 32);
+        secp256k1_sha256_write(&hash, pk_ser, 32);
+        secp256k1_sha256_write(&hash, &msgs32[i*32], 32);
+        /* 1.b) Copy the hash */
+        hashcopy = hash;
+        /* 1.c) Finalize the copy to get zi*/
+        secp256k1_sha256_finalize(&hashcopy, hashoutput);
+        secp256k1_scalar_set_b32(&zi, hashoutput, NULL);
```
real-or-random, line 429: Have you considered extracting this to a separate helper function? It's also used in signing. (I am not entirely sure if it will be more readable in the end.)

src/modules/schnorrsig/tests_impl.h
```diff
@@ -1000,6 +1000,163 @@ static void test_schnorrsig_taproot(void) {
     CHECK(secp256k1_xonly_pubkey_tweak_add_check(CTX, output_pk_bytes, pk_parity, &internal_pk, tweak) == 1);
 }
 
+
+
```
real-or-random, line 1004: nit: We usually have just a single empty line. (Check also in the other files you touched.)

include/secp256k1_schnorrsig.h
```diff
@@ -183,6 +183,81 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_verify(
     const secp256k1_xonly_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
 
+
+/** Incrementally (Half-)Aggregate a sequence of Schnorr signatures to an existing half-aggregate signature.
```
real-or-random, line 187: nit: We try to keep doc comments in public headers to 80 chars per line.

real-or-random (commented):
This is very clean code already, I honestly the quality of the code.

A somewhat annoying point is that this should probably moved to a separate module. If you want, you can give it a try, but this will involve modifying the build system, too. (Just the autotools build system with configure.ac and Makefile.am. You can ignore the CMake, we don't support it currently in this repo, it's just there because it's in upstream secp256k1 and we would like to keep the diff small.) Happy to help if you want to give it a try; otherwise Jonas or I could just take of it.

src/modules/schnorrsig/main_impl.h
b-wagn: I think it will be less readable in the end. The function would not only have to output the zi, but also update hash.

src/modules/schnorrsig/main_impl.h
```diff
@@ -264,4 +264,206 @@ int secp256k1_schnorrsig_verify(const secp256k1_context* ctx, const unsigned cha
...
+    }
+
+    /* Compute s = s_old + sum_{i = n_before}^{n} z_i*s_i */
+    /* where s_old = 0 if n_before = 0 */
+    secp256k1_scalar_set_int(&s, 0);
+    if (n_before > 0) {
+        secp256k1_scalar_set_b32(&s, &aggsig[n_before*32], &overflow);
+        if (overflow) {
+            return 0;
+        }
```
jonasnick, line 330: I don't think this is spec compliant at the moment. The spec ignores overflow:

> int(aggsig[(v⋅32:(v+1)⋅32])

Maybe we shouldn't do that in the spec though.

src/modules/schnorrsig/main_impl.h
jonasnick: i think when i wrote this, i liked the idea of having no error paths. currently it returns 1 always (if the overflow of the previous s above is fixed to be consistent with the spec). however, failing early also has it's merit. i opened a pr for the spec change: https://github.com/blockstreamresearch/cross-input-aggregation/pull/10.

src/modules/schnorrsig/main_impl.h
jonasnick: PR for the spec change:  https://github.com/blockstreamresearch/cross-input-aggregation/pull/10.

include/secp256k1_schnorrsig.h
```diff
@@ -183,6 +183,94 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_schnorrsig_verify(
     const secp256k1_xonly_pubkey *pubkey
 ) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(5);
 
+/** Incrementally (Half-)Aggregate a sequence of Schnorr 
+ *  signatures to an existing half-aggregate signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In/Out:      aggsig: pointer to the serialized aggregate signature
```
jonasnick, line 191: There are a lot of trailing whitespaces in your code (e.g., here).

jonasnick (commented):
Here's a (maybe somewhat outdated) template for creating an independent module: https://github.com/BlockstreamResearch/secp256k1-zkp/commit/048f9f8642297578a4e7975fa1e9837a58fc1c66

src/modules/schnorrsig/main_impl.h
jonasnick: PR has been merged

jonasnick:
We've just merged #270 which means that this PR can be rebased to fix CI.

configure.ac
```diff
@@ -490,6 +494,11 @@ if test x"$enable_module_schnorrsig" = x"yes"; then
   enable_module_extrakeys=yes
 fi
 
+if test x"$enable_module_schnorrsig_halfagg" = x"yes"; then
+  SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_SCHNORRSIG_HALFAGG=1"
+  enable_module_schnorrsig=yes
+fi
```
jonasnick, line 500: You must do this before testing for schnorrsig because the setting `enable_module_schnorrsig=yes` happens _after_ checking for `enable_module_schnorrsig=yes`. There's a comment above mentioning this (sorry, hard to find):
```
# Besides testing whether modules are enabled, the following code also enables
# module dependencies. The order of the tests matters: the dependency must be
# tested first.
```

src/modules/schnorrsig_halfagg/Makefile.am.include
```diff
@@ -0,0 +1,3 @@
+include_HEADERS += include/secp256k1_schnorrsig_halfagg.h
+noinst_HEADERS += src/modules/schnorrsig_halfagg/main_impl.h
+noinst_HEADERS += src/modules/schnorrsig_halfagg/tests_impl.h
```
real-or-random, line 3: nit: file should end with a line break

real-or-random (commented):
@b-wagn I think the main thing to do here from your side is to organize this PR in proper commits. Normally, we would probably prefer small atomic commits (one logical change at a time) but, as this adds an entirely new module instead of changing existing code, all the changes could just be squashed into a single commit. Or you could have one commit for the code, one commit for the tests, and one commit for the build system + CI or something like this. Please don't hesitate to force-push to your branch.

I think then we can do a final review, and I don't except further significant changes then. If you ask me), it's ready for merge then as an experimental module. 

We should perhaps add a version number (or some revision identifier) to the BIP draft, as long as the BIP is unstable. We could then say here which version the current code implements. Since it's an experimental module, we don't guarantee API stability, so it's not a big deal if we change the BIP in an  incompatible way and need to update the code accordingly.

include/secp256k1_schnorrsig_halfagg.h
```diff
@@ -0,0 +1,107 @@
...
+extern "C" {
+#endif
+
+
+/** Incrementally (Half-)Aggregate a sequence of Schnorr
+ *  signatures to an existing half-aggregate signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In/Out:      aggsig: pointer to the serialized aggregate signature
```
real-or-random, line 17: Perhaps we should clarify that the first `32*(n_before+1)` of this array should hold the input aggsig

include/secp256k1_schnorrsig_halfagg.h
```diff
@@ -0,0 +1,107 @@
...
+ *                       that is input. Will be overwritten by the new
+ *                       serialized aggregate signature.
+ *          aggsig_size: size of the memory allocated in aggsig.
+ *                       Should be large enough to hold the new
+ *                       serialized aggregate signature.
+ *                       I.e., should be a multiple of 32 and should
```
real-or-random, line 23: (no multiple of 32, see other comment)

include/secp256k1_schnorrsig_halfagg.h
```diff
@@ -0,0 +1,107 @@
...
+ *                       satisfy aggsig_size >= 32*(n_before+n_new+1)
+ *  In:     all_pubkeys: Array of x-only public keys, including both
```
real-or-random, line 25: nit: This way of writing things is a bit more concise (and no need to define n, which is then unused) but this is a probably a matter of taste. Feel free to pick up or disregard.
```suggestion
 *  In:     all_pubkeys: Array of (n_before + n_new) x-only public keys, including both
```

If you like this, then I suggest doing it also for the other parameters here and for the other functions.

real-or-random (commented):
I still need to look at the tests and the build system/CI stuff again.

include/secp256k1_schnorrsig_halfagg.h
```diff
@@ -0,0 +1,107 @@
...
+
+/** Incrementally (Half-)Aggregate a sequence of Schnorr
+ *  signatures to an existing half-aggregate signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:           ctx: a secp256k1 context object.
+ *  In/Out:      aggsig: pointer to the serialized aggregate signature
+ *                       that is input. Will be overwritten by the new
+ *                       serialized aggregate signature.
+ *          aggsig_size: size of the memory allocated in aggsig.
```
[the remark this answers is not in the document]
b-wagn: I will wait for @jonasnick's response before I fix the comment with `_len`  vs `size`.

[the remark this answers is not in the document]
jonasnick: `_len` sounds good to me. I like it more than `_maxlen` because then we can use the same `aggsig_len` also in `_verify`.

b-wagn:
Thank you for your comments @real-or-random. 
I addressed them, and I will organize the PR into one commit as soon as your review is completed.

include/secp256k1_schnorrsig_halfagg.h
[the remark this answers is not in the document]
real-or-random: Yeah, now that you point this out, I agree. As I said above, we set this value. It would be confusing to set `_maxlen`...

real-or-random:
When reviewing the tests, I had a few comments in mind, and then I just started addressing them on my own because I thought that's more efficient in the end. But then this ended up in a larger rewrite. :D  (Also one reason why the diff seems so large is that I made the comments a bit more consistent...) 

I pushed my suggestions to https://github.com/real-or-random/secp256k1-zkp/tree/schnorr-half-agg. Please don't hesitate to cherry-pick the commit, but please also double-check that I didn't screw up anything...

b-wagn:
Thank you @real-or-random. I went through your changes and they make sense to me, so I included them.

src/modules/schnorrsig_halfagg/tests_impl.h
```diff
@@ -0,0 +1,347 @@
...
+    /* Test vector 0 */
+    {
+        size_t n = 0;
+        const unsigned char pubkeys_ser[32] = {
+            0x1b, 0x84, 0xc5, 0x56, 0x7b, 0x12, 0x64, 0x40,
+            0x99, 0x5d, 0x3e, 0xd5, 0xaa, 0xba, 0x05, 0x65,
+            0xd7, 0x1e, 0x18, 0x34, 0x60, 0x48, 0x19, 0xff,
+            0x9c, 0x17, 0xf5, 0xe9, 0xd5, 0xdd, 0x07, 0x8f
+        };
+        secp256k1_xonly_pubkey pubkeys[1]; /* C doesn't allow zero-size arrays. */
```
real-or-random, line 85: Note: This indicates that the API is a bit inconvenient for empty aggregates (i.e., n=0). 

In the schnorrsig module, we allow the caller to pass NULL for an empty message:
https://github.com/bitcoin-core/secp256k1/blob/a9db9f2d75ac9df0312dec0c329266969bcc2946/include/secp256k1_schnorrsig.h#L154

We could probably do the same here, if only for consistency. I am not sure if any caller will need this, but from the point of view of the implementation it's really just the promise not to deference the pointer (which we already keep!),  and a modified ARG_CHECK.

But let's first see if we would like to allow n=0 at all https://github.com/BlockstreamResearch/cross-input-aggregation/issues/11#issuecomment-1877084001

real-or-random:
Changes look good! 

This needs rebase again due to changes in configure.ac (hopefully the last time!), but I just noticed that we'd like to do another change to configure.ac, so let us do this first to avoid that you'll need to rebase here twice...

src/modules/schnorrsig_halfagg/tests_impl.h
real-or-random: > But let's first see if we would like to allow n=0 at all [BlockstreamResearch/cross-input-aggregation#11 (comment)](https://github.com/BlockstreamResearch/cross-input-aggregation/issues/11#issuecomment-1877084001)

I think we want to allow n=0...  @jonasnick Do you think we should document that NULL is allowed here for n=0? I tend to say yes, just because it doesn't seem to hurt.

real-or-random:
> Changes look good!
> 
> This needs rebase again due to changes in configure.ac (hopefully the last time!), but I just noticed that we'd like to do another change to configure.ac, so let us do this first to avoid that you'll need to rebase here twice...

#290 has been merged, so this is ready for rebase. (#290 adds a comment that explains in which order modules should appear in `configure.ac` and elsewhere.)

src/modules/schnorrsig_halfagg/tests_impl.h
jonasnick: Yeah, we can change the ARG_CHECKs in aggverify to
```
ARG_CHECK(pubkeys != NULL || n == 0);
ARG_CHECK(msgs32 != NULL || n == 0);
```
and change the function doc accordingly.

b-wagn: Shouldn't it be:
```
ARG_CHECK(pubkeys != NULL || (n == 0 && pubkeys == NULL));
ARG_CHECK(msgs32  != NULL || (n == 0 && msgs32  == NULL));
ARG_CHECK(aggsig  != NULL || (n == 0 && aggsig  == NULL));
```
Although in the Schnorrsig module it seems like the public key and signature must never be NULL.
I don't know if we want to allow that n = 0 but pk / msg / aggsig is not NULL.

jonasnick: I think they're logically equivalent:
```
pubkeys != NULL || (n == 0 && pubkeys == NULL)
= (pubkeys != NULL || n == 0) && (pubkeys != NULL || pubkeys == NULL)
= pubkeys != NULL || n == 0
```

I don't think we need to allow `aggsig == NULL`. For `n = 0`, aggsig is defined to be 32 bytes of 0.

b-wagn: Oh, sure. 
But if `msgs32 == NULL`, do we want to require that `n > 0`?

real-or-random: > But if `msgs32 == NULL`, do we want to require that `n > 0`?

I don't fully understand. You mean `msgs32 != NULL`?

Anyway, what @jonasnick suggested seems sane to me:
```
ARG_CHECK(pubkeys != NULL || n == 0);
ARG_CHECK(msgs32 != NULL || n == 0);
```
One can think of the inputs as (pubkey, message) pairs. (And this is how you'd implement it in other languages.) So it makes sense to treat `pubkeys` and `msgs32` equally. But the signature should always be present.

And note that is different from the Schnorr module. There, `msg32 == NULL` merely indicates a message of `""` (byte array of length 0), but the message is still present. (And we can't do this here, this would require a different interface where `msgs32` is an `unsigned char **`. Such an interface may save the caller some memcpys, but let's not care about this now.)

b-wagn: > I don't fully understand. You mean msgs32 != NULL?

yes, that's what I meant, sorry.
But I will implement @jonasnick's suggestion then.

real-or-random (approved):
ACK e1d15fd9cca27b1d6dabfdf3057a334016575904

Can you squash this into a single comment, then we can merge it.

jonasnick (commented):
Concept ACK

I did a few sanity checks and read the include file again. Looks all good. I have a few nits that can be cleaned up in a separate PR. Your detailed and helpful comments in the main code and tests is very much appreciated.

Nits:
- we may want to add a header to the include file that links to the BIP
- there's still a mention of aggsig_size
- we may want to move `aggregate` before `inc_aggregate`
- we should mention expected size of input aggsig array in`_aggregate`
- "Should be aggsig_len = 32*(n+1)" -> "Must be"

real-or-random (approved):
ACK 3a9b1d46a31e8ca93a94752e08b514c06ebc2c0c

fjahr:
@b-wagn Are you interested in opening a PR for this in secp256k1 any time soon? If not, I will give this a try myself because I want to experiment with it in core.

b-wagn:
Currently not planning to do so. Feel free to go ahead 👍

jonasnick:
Hi @Loopite,

You are correct that "full" aggregation is more efficient. However, one of its major downsides is that it requires an interactive protocol between the signers. There are applications where half aggregation is better suited, while others benefit from "full" aggregation. You may want to check out @fjahr's https://cisaresearch.org/.

I don't think there would be any opposition to a `schnorr_full_aggregate` module that is based on a reasonable specification/BIP, but that doesn't exist yet.

Issue #191, opened 2022-04-28
Schnorr Adaptor Signatures


wolfmcnally:
Does libsecp256k1-zkp offer a straightforward way to produce Schnorr adaptor signatures? I see that it has an API for ECDSA adaptor signatures, and that the Musig2 API has an "adaptor" aspect, but I'm interested in simply creating basic adaptor signatures that can be used with BIP-340, and I don't see a clear way to do this.

jonasnick:
No it does not, but it would be nice if it did. However, I don't know of anyone who is planning to implement it short to mid-term.

PR #268, opened 2023-08-19
Add an experimental schnorr signature adaptor module

ZhePang:
### Overview
This PR adds support for Schnorr Adaptor signatures. It is based on the work of Schnorr Signature by @jonasnick and @apoelstra, and the discussion #191.
### Schnorr Adaptor Signatures
This implementation is mainly based on the idea of adaptor signatures in [BIP-340 specification](https://github.com/bitcoin/bips/blob/e643d247c8bc086745f3031cdee0899803edea2f/bip-0340.mediawiki#adaptor-signatures). It refers to the design and implementation of ECDSA adaptor signatures and MuSig2 adaptor signature.

Instead of having a verification algorithm for the adaptor signature, this module uses extract_t which extracts the adaptor point from the adaptor signature. ([graph for complete atomic swap process](https://drive.google.com/file/d/1ZlJGiN8n3pHWZBcPX_nhJPt4M377nMbB/view?usp=sharing))
### Python
A [Python specification of Schnorr Adaptor signatures](https://github.com/ZhePang/Python_Specification_for_Schnorr_Adaptor) has also been implemented, which generates all test vectors for the module.


ssantos21:
Concept ACK.

It might be interesting to include a `examples/adaptor.c` file.

instagibbs:
> It might be interesting to include a examples/adaptor.c file.

this would be great :)

ZhePang:
@ssantos21 @instagibbs Thanks for the advice! I'll for sure add the example code for Schnorr Adaptor.

ZhePang:
This pull request is still under review and most likely will have some modifications. Feel free to give me more suggestions and advice!

siv2r:
The current const time test is failing because:

- invalid compressed adaptor point value
- `pre_sig` doesn’t declassify `R`, and `T` before using them is `gej_add_ge_var`

I have fixed these issues in this commit: https://github.com/siv2r/secp256k1-zkp/commit/ad1cc6e78be02f5c99cdf3dcc66c64576b2735f8. Feel free to cherry-pick it using git or simply copy the changes :)

The commit also adds const-time tests for the `adapt`, and `extract_adaptor` APIs.

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,140 @@
...
+ *           data: arbitrary data pointer that is passed through
+ *
+ *  Except for test cases, this function should compute some cryptographic hash of
+ *  the message, the key, the adaptor point, the pubkey, the algorithm description, and data.
+ */
+typedef int (*secp256k1_adaptor_nonce_function_hardened)(
+    unsigned char *nonce32,
+    const unsigned char *msg32,
+    const unsigned char *key32,
+    const unsigned char *xonly_t32,
```
siv2r, line 38:
```suggestion
    const unsigned char *t33,
```
It is better to pass the compressed adaptor pubkey for nonce creation (without ignoring the parity byte)

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,140 @@
...
+ *  Improvement Proposal 340 "Schnorr Signatures for secp256k1"
+ *  (https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
+ *
+ *  If a data pointer is passed, it is assumed to be a pointer to 32 bytes of
+ *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
+ *  the nonce derivation procedure follows BIP-340 by setting the auxiliary
+ *  random data to zero. The algo argument must be non-NULL, otherwise the
+ *  function will fail and return 0. The hash will be tagged with algo.
+ *  Therefore, to create BIP-340 compliant signatures, algo must be set to
+ *  "BIP0340/nonce" and algolen to 13.
```
siv2r, line 55: We need to use a new tag for adaptor nonce, and aux. Something like `SchnorrAdaptor/nonce`, and `SchnorrAdaptor/aux` (or a better name).

Currently, we are using the ones from the bip340 schnorr sign.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+            masked_key[i] ^= key32[i];
+        }
+    } else {
+        /* Precomputed TaggedHash("BIP0340/aux", 0x0000...00); */
+        static const unsigned char ZERO_MASK[32] = {
+              84, 241, 105, 207, 201, 226, 229, 114,
+             116, 128,  68,  31, 144, 186,  37, 196,
+             136, 244,  97, 199,  11,  94, 165, 220,
+             170, 247, 175, 105, 39,  10, 165,  20
+        };
```
siv2r, line 37: These pre-computed values will change after updating the aux tag.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+
+    return ret;
+}
+
+int secp256k1_schnorr_adaptor_presign(const secp256k1_context *ctx, unsigned char *sig65, const unsigned char *msg32, const secp256k1_keypair *keypair, const unsigned char *t33, const unsigned char *aux_rand32) {
+    /* We cast away const from the passed aux_rand32 argument since we know the default nonce function does not modify it. */
+    return secp256k1_schnorr_adaptor_presign_internal(ctx, sig65, msg32, keypair, secp256k1_adaptor_nonce_function_bip340, t33, (unsigned char*)aux_rand32);
+}
+
+int secp256k1_schnorr_adaptor_extract_t(const secp256k1_context *ctx, unsigned char *t33, const unsigned char *sig65, const unsigned char *msg32, const secp256k1_xonly_pubkey *pubkey) {
```
siv2r, line 149: We don't need this func to be const-time (can be variable time). So, no need to use a `ret` variable and store intermediate failures (like the `adaptor_presig`, and `schnorrsig_sign` do). 

We can simply `return 0` when scalar overflow, pubkey parse failure, etc. You can refer to the `schnorrsig_verify`, and `ecdsa_adaptor_verify` APIs. They are implemented as variable time too.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+        ;
+    } else if (sig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
+        secp256k1_gej_neg(&tj, &tj);
+    } else {
+        ret = 0;
+    }
+    secp256k1_ge_set_gej(&t, &tj);
+    ret &= !!secp256k1_eckey_pubkey_serialize(&t, t33, &size, 1);
+
+    secp256k1_memczero(t33, 33, !ret);
```
siv2r, line 209:
```suggestion
    memset(t33, 0, 33);
```
The function can be variable time. We probably need to perform this at the start of the function if we plan to `return 0` in case of failures.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+    size_t msglen = 32;
+    int overflow;
+    int ret = 1;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(t33 != NULL);
+    ARG_CHECK(sig65 != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(pubkey != NULL);
+
```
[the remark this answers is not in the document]
siv2r: We can add this test for the `schnorr_adaptor_adapt` API too.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+    /* T = R0 + (- R) */
+    secp256k1_gej_neg(&rj, &rj);
+    secp256k1_gej_add_ge_var(&tj, &rj, &r0, NULL);
+    if (sig65[0] == SECP256K1_TAG_PUBKEY_EVEN) {
+        ;
+    } else if (sig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
+        secp256k1_gej_neg(&tj, &tj);
+    } else {
+        ret = 0;
+    }
```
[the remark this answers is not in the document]
siv2r: Similar simplification is possible in `schnorr_adapt_adaptor`.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+    if (sig65[0] == SECP256K1_TAG_PUBKEY_EVEN) {
+        secp256k1_scalar_add(&s, &s0, &t);
+    } else if (sig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
+        secp256k1_scalar_negate(&t, &t);
+        secp256k1_scalar_add(&s, &s0, &t);
+    } else {
+        ret = 0;
+    }
+
+    memset(sig64, 0, 64);
```
siv2r, line 244: `memset` is unnecessary since we assign value to `sig64` after this anyway.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+    memcpy(sig64, &sig65[1], 32);
+    secp256k1_scalar_get_b32(&sig64[32], &s);
+
```
siv2r, line 247: We need to clear the scalars `s`, and `t`. And conditionally set `sig64` to `0` if `ret == 0`.

```c
secp256k1_memczero(sig64, 64, !ret);
secp256k1_scalar_clear(&s);
secp256k1_scalar_clear(&t);
```

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+        secp256k1_scalar_add(&t, &s, &s0);
+    } else if (sig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
+        secp256k1_scalar_negate(&s, &s);
+        secp256k1_scalar_add(&t, &s0, &s);
+    } else {
+        ret = 0;
+    }
+
+    secp256k1_scalar_get_b32(t32, &t);
+
```
siv2r, line 282: We need to clear the scalars `s` and `t`. And conditionally set `t32` to `0` if `ret == 0`.

siv2r: Without this, if the user passes an overflowed `s`, and `s0`, this API generates a valid `t`, but we need to set `t` to `0`.

ci/cirrus.sh
siv2r, line 1: - define the `SCHNORRADAPTOR` env variable in `.cirrus.yml file`
    - this is why the CI tests are failing
    - basically replace `BATCH` with `SCHNORRADAPTOR` in [this change](https://github.com/bitcoin-core/secp256k1/pull/1134/files#diff-62587956f943bb2503db7bc6dd27d0d888074a1c0ecaab3f570ad611aff0f7bb)

include/secp256k1_schnorr_adaptor.h
ZhePang: Changed in the new commit

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,140 @@
...
+    const unsigned char *xonly_t32,
+    const unsigned char *xonly_pk32,
+    const unsigned char *algo,
+    size_t algolen,
+    void *data
+);
+
+/** An implementation of the nonce generation function as defined in Bitcoin
+ *  Improvement Proposal 340 "Schnorr Signatures for secp256k1"
+ *  (https://github.com/bitcoin/bips/blob/master/bip-0340.mediawiki).
```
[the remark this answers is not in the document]
ZhePang: Changed in the new commit

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+     * because R' is not a secret.  */
+    secp256k1_declassify(ctx, &r0, sizeof(r0));
+    secp256k1_fe_normalize_var(&r0.y);
+    if (secp256k1_fe_is_odd(&r0.y)) {
+        secp256k1_scalar_negate(&k, &k);
+    }
+
+    ret &= !!secp256k1_eckey_pubkey_serialize(&r0, sig65, &size, 1);
+
+    secp256k1_schnorrsig_challenge(&e, &sig65[1], msg32, msglen, pk_buf);
```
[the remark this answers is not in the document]
ZhePang: The challenge function here should be consistent with the schnorrsig challenge function, it just hash the x coordinate of R', and that is okay since we have the parity reordered in the first half of the signature.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,286 @@
...
+    secp256k1_scalar_set_b32(&k, nonce32, NULL);
+    ret &= !secp256k1_scalar_is_zero(&k);
+    secp256k1_scalar_cmov(&k, &secp256k1_scalar_one, !ret);
+
+    /* R = k0*G */
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rj, &k);
+    secp256k1_ge_set_gej(&r, &rj);
+
+    /* T = cpoint(T) */
+    ret &= !!secp256k1_eckey_pubkey_parse(&t, t33, 33);
```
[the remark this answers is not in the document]
ZhePang: I thought this is a convention for using !! instead of nothing

src/modules/schnorr_adaptor/main_impl.h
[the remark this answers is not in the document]
ZhePang: @jonasnick discussed with me and we by that time decided to return 0 on this case. We can have further discussion on this behavior.

src/modules/schnorr_adaptor/main_impl.h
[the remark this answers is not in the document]
ZhePang: Same issue as the previous conversation

src/modules/schnorr_adaptor/main_impl.h
ZhePang: Sounds good

ZhePang: I used secp256k1_memczero here because if ret == 0, t33 must be set 0.

src/modules/schnorr_adaptor/main_impl.h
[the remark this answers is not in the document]
siv2r: The other parts of libsecp use `!!` only for `noncefp` (and not for `parse` and `serialize`). So, we can remove this here imo.

src/modules/schnorr_adaptor/main_impl.h
siv2r: Oh, okay. Then, you can use: `if(!ret) memset(t33, 0, 33)`, when you make this func variable time.

src/modules/schnorr_adaptor/main_impl.h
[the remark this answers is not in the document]
ZhePang: can be used in adapt and extract_adaptor

src/modules/schnorr_adaptor/main_impl.h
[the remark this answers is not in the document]
real-or-random: `!!` is an idiom for ensuring that the value is either `0` or `1` (and not some other non-zero value). Recall that C doesn't have a proper bool type. 

We use `!!` for valued returned by `noncefp` because `noncefp` is provided by the user, so we don't control it. For the functions in our code base, our usual convention is that the called function is responsible for returning only `0` and `1`. So yes, strictly speaking, `!!` is not necessary here, but it also won't hurt. I'd keep it.

jonasnick:
We've just merged https://github.com/BlockstreamResearch/secp256k1-zkp/pull/270 which means that this PR can be rebased to fix CI. Note that rebasing may require some additional changes that are documented in https://github.com/BlockstreamResearch/secp256k1-zkp/discussions/271. Let me know if you have any questions.

ci/ci.sh
```diff
@@ -82,6 +82,7 @@ esac
     --enable-module-rangeproof="$RANGEPROOF" --enable-module-whitelist="$WHITELIST" --enable-module-generator="$GENERATOR" \
     --enable-module-schnorrsig="$SCHNORRSIG"  --enable-module-musig="$MUSIG" --enable-module-ecdsa-adaptor="$ECDSAADAPTOR" \
     --enable-module-schnorrsig="$SCHNORRSIG" \
+    --enable-module-schnorr-adaptor="$SCHNORRADAPTOR" \
```
jonasnick, line 85: You need to add SCHNORRADAPTOR also to the `print_environment` function

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,991 @@
...
+    unsigned char t2[33];
+    unsigned char adaptor[32];
+
+    /** setup **/
+    int ecount;
+
+    secp256k1_context_set_error_callback(CTX, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(CTX, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_error_callback(STATIC_CTX, counting_illegal_callback_fn, &ecount);
+    secp256k1_context_set_illegal_callback(STATIC_CTX, counting_illegal_callback_fn, &ecount);
```
jonasnick, line 140: This can all be removed. And instead of
```
CHECK(secp256k1_schnorr_adaptor_presign(CTX, sig, msg, &keypairs[0], t, NULL) == 1);
CHECK(ecount == 0);
```
write 
```
CHECK(secp256k1_schnorr_adaptor_presign(CTX, sig, msg, &keypairs[0], t, NULL) == 1);
```


Instead of 
```
CHECK(secp256k1_schnorr_adaptor_presign(STATIC_CTX, sig, msg, &keypairs[0], t, NULL) == 0);
CHECK(ecount == 1);
```
write
```
CHECK_ILLEGAL(STATIC_CTX, secp256k1_schnorr_adaptor_presign(STATIC_CTX, sig, msg, &keypairs[0], t, NULL) == 0);
```

jonasnick (commented):
The CI tasks fail because `.github/workflows/ci.yml` does not contain a definition of SCHNORRADAPTOR. Just search for `BPPP` and add a definition for SCHNORRADAPTOR right next to it that defines SCHNORRADAPTOR to the same value as `BPPP` (`'yes'` or `'no'`).

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+
+/** Extract the adaptor from an adaptor signature and a schnorr signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:    ctx: pointer to a context object.
+ *  Out:      t32: pointer to a 32-byte array to store the adaptor.
+ *  In:    sig65: pointer to a 65-byte serialized adaptor signature.
+ *           sig64: pointer to a 64-byte adapted schnorr signature.
+ */
+SECP256K1_API int secp256k1_schnorr_adaptor_extract_adaptor(
```
jonasnick, line 127: I think we should rename this to `secp256k1_schnorr_adaptor_extract` for simplicity.

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+/** Extract an adaptor point T from the signature.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:    ctx: pointer to a context object.
+ *  Out:      t33: pointer to a 33-byte array to store the compressed adaptor point.
+ *  In:     sig65: pointer to a 65-byte serialized signature.
+ *           msg32: the 32-byte message being signed.
+ *           pubkey: pointer to an x-only public key to verify with (cannot be NULL)
+ */
+SECP256K1_API int secp256k1_schnorr_adaptor_extract_t(
```
jonasnick, line 96: I think we should rename this because `t` is confusing. Maybe `secp256k1_schnorr_adaptor_preextract`?

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *           t33: pointer to a 33-byte compressed adaptor point.
+ *    aux_rand32: 32 bytes of fresh randomness. While recommended to provide
+ *                this, it is only supplemental to security and can be NULL. A
+ *                NULL argument is treated the same as an all-zero one. See
+ *                BIP-340 "Default Signing" for a full explanation of this
+ *                argument and for guidance if randomness is expensive.
+ */
+SECP256K1_API int secp256k1_schnorr_adaptor_presign(
+    const secp256k1_context *ctx,
+    unsigned char *sig65,
```
jonasnick, line 80: I think it's more clear to rename sig65 -> presig65 (everywhere).

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+    const unsigned char *msg32,
+    const secp256k1_keypair *keypair,
+    const unsigned char *t33,
```
jonasnick, line 83: Wouldn't it be better if this was of type `pubkey`? Converting from bytes to point is costly. If `t33` was already provided as pubkey we wouldn't have to do the conversion.

jonasnick: Also we should consider renaming `t33 -> adaptor`, `t32 -> secadaptor`.

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+    const unsigned char *aux_rand32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Extract an adaptor point T from the signature.
```
jonasnick, line 87: We should consider adding documentation like

> If this funtion passes, `secp256k1_schnorr_adaptor_adapt` called with the secret key corresponding to public key T and `presig65` outputs a valid Schnorr signature.

include/secp256k1_schnorr_adaptor.h
siv2r: Similarly after https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#discussion_r1377766484, `secp256k1_schnorr_adaptor_extract` or `secp256k1_schnorr_adaptor_extract_adaptor`?

ZhePang: I think for simplicity, we can name extract_t as extract and extract_adaptor as extract_sec, so it matches the variable names for adaptor and sec adaptor

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+SECP256K1_API int secp256k1_schnorr_adaptor_presign(
+    const secp256k1_context *ctx,
+    unsigned char *presig65,
+    const unsigned char *msg32,
+    const secp256k1_keypair *keypair,
+    const unsigned char *adaptor,
+    const unsigned char *aux_rand32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Extract an adaptor point T from the signature.
```
jonasnick, line 87:
```suggestion
/** Extract an adaptor point from the signature.
```

ZhePang:
> I was playing around with this, and it's great! Though one issue I encountered is that the `schnorr_adaptor_extract_adaptor` will sometimes lead to a call to the `illegal_callback` for invalid adaptor signatures when the extracted public key is invalid (which leads to a panic at the rust layer). This makes is not a very good method for verifying adaptor signatures.

Hi @Tibo-lg , we expect users to call `schnorr_adaptor_extract` (was `extract_t`) before `schnorr_adaptor_extract_sec` (was `schnorr_adaptor_extract_adaptor`) so in this way the adaptor signature passed to 'extract_sec' is valid and would not raise `illegal_callback`. We'll put that in the documentation, thanks for mentioning it up!

jonasnick:
> Hi @Tibo-lg , we expect users to call schnorr_adaptor_extract (was extract_t) before schnorr_adaptor_extract_sec (was schnorr_adaptor_extract_adaptor) so in this way the adaptor signature passed to 'extract_sec' is valid and would not raise illegal_callback. We'll put that in the documentation, thanks for mentioning it up!

Actually we don't. Sorry, I was wrong when we talked about this. Typically, the party that created the adaptor signature would call `extract_sec` and therefore they know that they created the first byte correctly. But since this cannot be guaranteed in a proper type safe language like rust, I'd prefer if we'd remove the ARG_CHECK and just returned 0 instead.

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,957 @@
...
+    int i;
+    run_adaptor_nonce_function_bip340_tests();
+
+    test_schnorr_adaptor_api();
+    test_schnorrsig_sha256_tagged();
+    test_schnorr_adaptor_vectors();
+    for (i = 0; i < COUNT; i++) {
+        test_schnorr_adaptor_presign();
+        test_schnorr_adaptor_extract();
+        test_schnorr_adaptor_adapt_extract_sec();
```
jonasnick, line 953: I don't quite understand the intentions of these tests. `presign` seems to be tested in both `test_schnorr_adaptor_extract()` and `test_schnorr_adaptor_adapt_extract_sec` so `test_schnorr_adaptor_presign` seems to be redundant. I think it would be sufficient to have one "correctness" test that runs the full protocol (essentially `test_schnorr_adaptor_adapt_extract_sec`) and one test for failures due to bitflips or negating `s` (essentially `test_schnorr_adaptor_extract`).

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *  different size (or the same size but without a context-specific tag
+ *  prefix), it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an context-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple contexts by accident.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:    ctx: pointer to a context object (not secp256k1_context_static).
+ *  Out:   presig65: pointer to a 65-byte array to store the serialized adaptor signature.
+ *  In:    msg32: the 32-byte message being signed.
```
jonasnick, line 69: Can you indent the arguments in this file such that the colon `:` is in the same column; like this:
```suggestion
 *  Out:   presig65: pointer to a 65-byte array to store the serialized adaptor signature.
 *  In:       msg32: the 32-byte message being signed.
```

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *  This function only signs 32-byte messages. If you have messages of a
+ *  different size (or the same size but without a context-specific tag
+ *  prefix), it is recommended to create a 32-byte message hash with
+ *  secp256k1_tagged_sha256 and then sign the hash. Tagged hashing allows
+ *  providing an context-specific tag for domain separation. This prevents
+ *  signatures from being valid in multiple contexts by accident.
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:    ctx: pointer to a context object (not secp256k1_context_static).
+ *  Out:   presig65: pointer to a 65-byte array to store the serialized adaptor signature.
```
jonasnick, line 68: Can you remove "serialized" from "serialized adaptor signature" in this file? Adaptor signatures are always serialized.

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ */
+SECP256K1_API int secp256k1_schnorr_adaptor_extract(
+    const secp256k1_context *ctx,
+    unsigned char *adaptor,
+    const unsigned char *presig65,
+    const unsigned char *msg32,
+    const secp256k1_xonly_pubkey *pubkey
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4) SECP256K1_ARG_NONNULL(5);
+
+/** Adapt an adaptor signature to a schnorr signature.
```
jonasnick, line 104:
```suggestion
/** Adapt an adaptor signature to result in a Schnorr signature.
```

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *
+ *  Returns 1 on success, 0 on failure.
+ *  Args:         ctx: pointer to a context object.
+ *  Out:        sig64: pointer to a 64-byte array to store the adapted schnorr signature.
```
jonasnick, line 108:
```suggestion
 *  Out:        sig64: pointer to a 64-byte array to store the adapted Schnorr signature.
```

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *         secadaptor: pointer to a 32-byte secadaptor.
+ */
+SECP256K1_API int secp256k1_schnorr_adaptor_adapt(
+    const secp256k1_context *ctx,
+    unsigned char *sig64,
+    const unsigned char *presig65,
+    const unsigned char *secadaptor
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
+
+/** Extract the secadaptor from an adaptor signature and a schnorr signature.
```
jonasnick, line 119:
```suggestion
/** Extract the secadaptor from an adaptor signature and a Schnorr signature.
```

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,324 @@
...
+    secp256k1_scalar_set_b32(&s, &sig64[32], &overflow);
+    ret &= !overflow;
+
+    if (presig65[0] == SECP256K1_TAG_PUBKEY_EVEN) {
+        secp256k1_scalar_negate(&s0, &s0);
+        secp256k1_scalar_add(&t, &s, &s0);
+    } else if (presig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
+        secp256k1_scalar_negate(&s, &s);
+        secp256k1_scalar_add(&t, &s0, &s);
+    }
```
jonasnick, line 314:
```suggestion
    if (presig65[0] == SECP256K1_TAG_PUBKEY_EVEN) {
        secp256k1_scalar_negate(&s0, &s0);
    } else if (presig65[0] == SECP256K1_TAG_PUBKEY_ODD) {
        secp256k1_scalar_negate(&s, &s);
    }
    secp256k1_scalar_add(&t, &s0, &s);
```

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,324 @@
...
+    secp256k1_scalar_set_b32(&k, nonce32, NULL);
+    ret &= !secp256k1_scalar_is_zero(&k);
+    secp256k1_scalar_cmov(&k, &secp256k1_scalar_one, !ret);
+
+    /* R = k0*G */
+    secp256k1_ecmult_gen(&ctx->ecmult_gen_ctx, &rj, &k);
+    secp256k1_ge_set_gej(&r, &rj);
+
+    /* T = cpoint(T) */
+    ret &= !!secp256k1_eckey_pubkey_parse(&t, adaptor, 33);
```
jonasnick, line 150: Sorry, this is another misunderstanding. We should really only use `!!` for `noncefp`. What @real-or-random meant in https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#discussion_r1330214126 was that in this particular line there's not a lot of benefit in fixing this. He did not want to suggest usage of this pattern generally.

jonasnick (commented):
I've just completed a review of your code and noticed a few areas that could benefit from refinement. I did not find any serious issues. It would be good if we could add another test vector as mentioned in the review comments. As promised, I'll also follow-up with an attempt at a module level documentation.

To further enhance your code's readability, you may want to tweak some of the code comments in the presign, extract, adapt, and extract_sec functions.

jonasnick:
Here's my suggestion for a module level documentation (to be put at the top of include/secp256k1_schnorr_adaptor.h):

```C
/** This module provides an experimental implementation of a Schnorr adaptor
 *  signature protocol variant.
 *
 *  The test vectors have been generated and cross-verified using a Python
 *  implementation of this adaptor signature variant available at [0].
 *
 *  The protocol involves two parties, Alice and Bob. The general sequence of
 *  their interaction is as follows:
 *  1. Alice calls the `schnorr_adaptor_presign` function for an adaptor point T
 *  and sends the pre-signature to Bob.
 *  2. Bob extracts the adaptor point T from the pre-signature using
 *  `schnorr_adaptor_extract`.
 *  3. Bob provides the pre-signature and the discrete logarithm of T to
 *  `schnorr_adaptor_adapt` which outputs a valid BIP 340 Schnorr signature.
 *  4. Alice extracts the discrete logarithm of T from the pre-signature and the
 *  BIP 340 signature using `schnorr_adaptor_extract_sec`.
 *
 *  In contrast to common descriptions of adaptor signature protocols, this
 *  module does not provide a verification algorithm for pre-signatures.
 *  Instead, `schnorr_adaptor_extract` returns the adaptor point encoded by a
 *  pre-signature, reducing communication cost. If a verification function for
 *  pre-signatures is needed, it can be easily simulated with
 *  `schnorr_adaptor_extract`.
 *
 *  Assuming that BIP 340 Schnorr signatures satisfy strong unforgeability under
 *  chosen message attack, the Schnorr adaptor signature scheme fulfills the
 *  following properties as formalized by [1].
 *
 *  - Witness extractability:
 *    If Alice
 *      1. creates a pre-signature with `schnorr_adaptor_presign` for message m
 *         and adaptor point T and
 *      2. receives a Schnorr signature for message m that she hasn't created
 *         herself,
 *    then Alice is able to obtain the discrete logarithm of T with
 *    `schnorr_adaptor_extract_sec`.
 *
 *  - Pre-signature adaptability:
 *    If Bob
 *      1. receives a pre-signature and extracts an adaptor point T using
 *         `schnorr_adaptor_extract`, and
 *      2. obtains the discrete logarithm of the adaptor point T
 *    Then then Bob is able to adapt the received pre-signature to a valid BIP
 *    340 Schnorr signature using `schnorr_adaptor_adapt`.
 *
 *  - Existential Unforgeability:
 *    Bob is not able to create a BIP 340 signature from a pre-signature for
 *    adaptor T without knowing the discrete logarithm of T.
 *
 *  - Pre-signature existiential unforgeability:
 *    Only Alice can create a pre-signature for her public key.
 *
 *  [0] https://github.com/ZhePang/Python_Specification_for_Schnorr_Adaptor
 *  [1] https://eprint.iacr.org/2020/476.pdf
 */
```

include/secp256k1_schnorr_adaptor.h
```diff
@@ -0,0 +1,138 @@
...
+ *  auxiliary random data as defined in BIP-340. If the data pointer is NULL,
+ *  the nonce derivation procedure follows BIP-340 by setting the auxiliary
+ *  random data to zero. The algo argument must be non-NULL, otherwise the
+ *  function will fail and return 0. The hash will be tagged with algo.
+ *  Therefore, to create BIP-340 compliant signatures, algo must be set to
+ *  "BIP0340/nonce" and algolen to 13.
+ */
+SECP256K1_API const secp256k1_adaptor_nonce_function_hardened secp256k1_adaptor_nonce_function_bip340;
+
+/** Create a Schnorr adaptor signature.
```
jonasnick, line 57: Would it make sense to replace the term "adaptor signature" with "pre-signature" to be consistent with the literature (see https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#issuecomment-1878484766)?

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,324 @@
...
+    secp256k1_sha256_write(&sha, adaptor, 33);
+    secp256k1_sha256_write(&sha, xonly_pk32, 32);
+    secp256k1_sha256_write(&sha, msg32, 32);
+    secp256k1_sha256_finalize(&sha, nonce32);
+    return 1;
+}
+
+const secp256k1_adaptor_nonce_function_hardened secp256k1_adaptor_nonce_function_bip340 = adaptor_nonce_function_bip340;
+
+static int secp256k1_schnorr_adaptor_presign_internal(const secp256k1_context *ctx, unsigned char *presig65, const unsigned char *msg32, const secp256k1_keypair *keypair, secp256k1_adaptor_nonce_function_hardened noncefp, const unsigned char *adaptor, void *ndata) {
```
siv2r, line 100:
```suggestion
static int secp256k1_schnorr_adaptor_presign_internal(const secp256k1_context *ctx, unsigned char *presig65, const unsigned char *msg32, const secp256k1_keypair *keypair, secp256k1_adaptor_nonce_function_hardened noncefp, const secp256k1_pubkey *adaptor, void *ndata) {
```
The current design allows the user to input an invalid adaptor point. By making this as `secp256k1_pubkey`, we can catch an invalid 33-byte adaptor point during `secp256k1_ec_pubkey_parse`. `secp256k1_ecdsa_adaptor_encrypt` API does the same.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,324 @@
...
+
+    /* We declassify R' to allow using it as a branch point. This is fine
+     * because R' is not a secret.  */
+    secp256k1_declassify(ctx, &r0, sizeof(r0));
+    secp256k1_fe_normalize_var(&r0.y);
+    if (secp256k1_fe_is_odd(&r0.y)) {
+        secp256k1_scalar_negate(&k, &k);
+    }
+
+    ret &= !!secp256k1_eckey_pubkey_serialize(&r0, presig65, &size, 1);
```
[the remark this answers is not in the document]
siv2r: Did you mean `ret &= secp256k1_eckey_pubkey_serialize(&r0, presig65, &size, 1);`?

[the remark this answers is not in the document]
siv2r: Would it be better to test for the infinity point immediately after `gej_add_ge_var`?
```
VERIFY_CHECK(secp256k1_ge_is_infinity(&r0))
```
This would match the [Python reference code](https://github.com/bitcoin/bips/blob/master/bip-0340/reference.py#L112).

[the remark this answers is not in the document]
jonasnick: > Did you mean to use `ret &= secp256k1_eckey_pubkey_serialize(&r0, presig65, &size, 1)` instead?

No I meant `ret =` without `&` but when I wrote this suggestion I thought that we could get rid of `ret &=` everywhere else in the function since I missed that this function is required to be constant time. We can write something like
```C
{
    int serialize_ret = secp256k1_eckey_pubkey_serialize(&r0, presig65, &size, 1);
    /* R0 is not the point at infinity with overwhelming probability */
    VERIFY_CHECK(serialize_ret);
}
```

> Would it be better to test for the infinity point immediately after gej_add_ge_var?

We can do that. I think even in that case we should VERIFY_CHECKing the return value of `serialize` to make sure it succeeded. Maybe adding an additional VERIFY_CHECK for infinity helps reading the code but not entirely sure.

src/modules/schnorr_adaptor/main_impl.h
```diff
@@ -0,0 +1,324 @@
...
+    secp256k1_scalar s0;
+    secp256k1_scalar s;
+    secp256k1_scalar t;
+    int overflow;
+    int ret = 1;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(secadaptor != NULL);
+    ARG_CHECK(presig65 != NULL);
+    ARG_CHECK(presig65[0] == SECP256K1_TAG_PUBKEY_EVEN || presig65[0] == SECP256K1_TAG_PUBKEY_ODD);
```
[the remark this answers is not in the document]
siv2r: I second this. libsecp handles other scenarios similarly:
- If the nonce is zero, `shcnorr_sign` sets `ret` to zero and continues with signature creation instead of calling an illegal callback.
- `eckey_pubkey_parse` also returns 0 for an invalid first byte.

This change also needs to be applied to the `schnorr_adaptor_adapt` API. 

At present, [our ctime test](https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#issuecomment-1694655543) considers the `adapt` and `extract_secadaptor` APIs to be non-const-time wrt `presig`. Therefore, it is safe to return 0 immediately. However, if we plan on making these APIs const-time wrt `presig`, we should set `ret` to zero instead.

src/modules/schnorr_adaptor/main_impl.h
siv2r: I am unsure if the change is really necessary since we parse the `adaptor` point using `secp256k1_ec_pubkey_parse` inside the presign function anyway.

siv2r: The essence of this modification is to make a design choice about serialization/deserialization.

- Option 1: user handles the ser/de-ser of the adaptor point
    - we receive the adaptor point as `secp256k1_pubkey` (in presign & extract_adaptor)
- Option 2: API handles the ser/de-ser of the adaptor point
    - we receive the adaptor point as `unsigned_char *` (in presign & extract_adaptor)

I think it makes sense to leave the ser/de-ser to the user since we already do this for the pubkey point. What’s the libsecp norm here?

siv2r: Just realized there is already a comment regarding this: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#discussion_r1442086962

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,957 @@
...
+    secp256k1_testrand256(sk2);
+    secp256k1_testrand256(sk3);
+    secp256k1_testrand256(msg);
+    secp256k1_testrand256(secadaptor);
+    CHECK(secp256k1_keypair_create(CTX, &keypairs[0], sk1) == 1);
+    CHECK(secp256k1_keypair_create(CTX, &keypairs[1], sk2) == 1);
+    CHECK(secp256k1_keypair_create(CTX, &keypairs[2], sk3) == 1);
+    CHECK(secp256k1_keypair_xonly_pub(CTX, &pk[0], NULL, &keypairs[0]) == 1);
+    CHECK(secp256k1_keypair_xonly_pub(CTX, &pk[1], NULL, &keypairs[1]) == 1);
+    CHECK(secp256k1_keypair_xonly_pub(CTX, &pk[2], NULL, &keypairs[2]) == 1);
```
siv2r, line 146: Since we use only one, creating three `sk,` `keypairs,` and `pk` seems unnecessary.

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,957 @@
...
+
+    CHECK(secp256k1_schnorr_adaptor_presign(CTX, sig, msg, &keypairs[0], t, NULL) == 1);
+    CHECK(secp256k1_schnorr_adaptor_extract(CTX, t2, sig, msg, &pk[0]) == 1);
+    CHECK_ILLEGAL(CTX, secp256k1_schnorr_adaptor_extract(CTX, NULL, sig, msg, &pk[0]));
+    CHECK_ILLEGAL(CTX, secp256k1_schnorr_adaptor_extract(CTX, t2, NULL, msg, &pk[0]));
+    CHECK_ILLEGAL(CTX, secp256k1_schnorr_adaptor_extract(CTX, t2, sig, NULL, &pk[0]));
+    CHECK_ILLEGAL(CTX, secp256k1_schnorr_adaptor_extract(CTX, t2, sig, msg, NULL));
+    CHECK_ILLEGAL(CTX, secp256k1_schnorr_adaptor_extract(CTX, t2, sig, msg, &zero_pk));
+
+    CHECK(secp256k1_schnorr_adaptor_adapt(CTX, sig64, sig, secadaptor) == 1);
```
siv2r, line 167: we can check this `sig64` result using `schnorrsig_verify`

src/modules/schnorr_adaptor/main_impl.h
jonasnick: Yes, I'd say the libsecp norm is to allow avoiding unnecessary costly recomputations such as pubkey deserialization.

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,957 @@
...
+        adaptor_nonce_function_bip340_bitflip(args, 4, algolen, algolen);
+        /* Flip algo again */
+        adaptor_nonce_function_bip340_bitflip(args, 4, algolen, algolen);
+        adaptor_nonce_function_bip340_bitflip(args, 5, 32, algolen);
+    }
+
+    /* NULL algo is disallowed */
+    CHECK(adaptor_nonce_function_bip340(nonce, msg, key, t, pk, NULL, 0, NULL) == 0);
+    CHECK(adaptor_nonce_function_bip340(nonce, msg, key, t, pk, algo, algolen, NULL) == 1);
+    /* Other algo is fine */
```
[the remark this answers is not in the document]
jonasnick: I don't think this test significantly increases test coverage because the nonce function does relatively little with algo. If we test algo = 0 an algo 0xff...f, why not test the same for the other nonce function arguments?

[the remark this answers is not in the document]
siv2r: > why not test the same for the other nonce function arguments?

I didn't consider this case. I suggested those tests because the ECDSA adaptor had them. 

I take back my suggestion. The `adaptor_nonce_function_bip340` function passes its arguments to the hash function without checking their values, except for `algo` and `aux`. Thus, a min-max test for each args seems unnecessary.

src/modules/schnorr_adaptor/tests_impl.h
siv2r: The `presign` test could benefit from testing with [different nonce functions](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/ecdsa_adaptor/tests_impl.h#L894) (failing, returns 0, overflowing).

siv2r: _side note:_ The ECDSA adaptor module performs these tests under a single [`adaptor_tests` function](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/ecdsa_adaptor/tests_impl.h#L868).

src/modules/schnorr_adaptor/tests_impl.h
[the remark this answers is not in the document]
siv2r: It would be nice to have a use-case test. For example,

- Schnorrsig module  has [`test_schnorrsig_taproot`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/schnorrsig/tests_impl.h#L932)
- ECDSA adaptor has [`multi_hop_lock_tests`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/ecdsa_adaptor/tests_impl.h#L1068)
- Musig2 adaptor has [`scriptless_atomic_swap`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/musig/tests_impl.h#L541)

siv2r, line 1: It would be nice to have a use-case test. For example,

- Schnorrsig module  has [`test_schnorrsig_taproot`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/schnorrsig/tests_impl.h#L932)
- ECDSA adaptor has [`multi_hop_lock_tests`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/ecdsa_adaptor/tests_impl.h#L1068)
- Musig2 adaptor has [`scriptless_atomic_swap`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/musig/tests_impl.h#L541)

jonasnick: Yes, similar to the "correctness" suggestion here: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268#discussion_r1442197643

src/modules/schnorr_adaptor/tests_impl.h
```diff
@@ -0,0 +1,957 @@
...
+
+/* Helper function for schnorr_adaptor_vectors
+ * Adapt a Schnorr adptor signature to a valid Schnorr signature
+ * and checks if it is the same as expected_sig64. */
+void test_schnorr_adaptor_vectors_check_adapt(const unsigned char *sig, const unsigned char *secadaptor, const unsigned char *expected_sig64, int expected) {
+    unsigned char sig64[64];
+
+    CHECK(secp256k1_schnorr_adaptor_adapt(CTX, sig64, sig, secadaptor));
+    CHECK(expected == (secp256k1_memcmp_var(sig64, expected_sig64, 64) == 0));
+}
```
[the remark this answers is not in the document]
ZhePang: I don't think we need to check this is a valid schnorr sig here since we think that adapt function would return a valid schnorr sig and the schnorr sig in the test vector should be valid. But in the python specification when creating the schnorr sig, we just use the adapt to create the schnorr sig in the test vectors. So I suggest we check if valid schnorr sig in the python specification when creating the test vectors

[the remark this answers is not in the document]
siv2r: The current testing method does this:

- extracts `expected_bip340_sig` and `test_result` from the test vector CSV file
- Then it checks the following:
    - **Case 1:** If `test_result == TRUE`
        - `adapt(presig, secadaptor)` output should match `expected_bip340_sig`
    - **Case 2:** If `test_result == FALSE`
        - `adapt(presig, secadaptor)` output shouldn’t match `expected_bip340_sig`

This way of testing is logically correct. However, it relies on the assumption that the `expected_bip340_sig` mentioned in the test vector is correct (i.e., passes `schnorrsig_verify`).

I think it would be nice to avoid this correctness assumption by directly calling `schnorrsig_verify` on the `adapt(presig, secadaptor)` output.

If we do this, then the new testing method would do:

- extract `test_result` from the test vector CSV file
- Then it will checks:
    - **Case 1:** If `test_result == TRUE`
        - `adapt(presig, secadaptor)` output should pass `schnorr_verify`
    - **Case 2:** If `test_result == FALSE`
        - `adapt(presig, secadaptor)` output should fail `schnorr_verify`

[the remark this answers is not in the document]
jonasnick: Now two months after I initially wrote this, I don't know anymore why I suggested this. I agree with @siv2r that this rules out some sort of human error but that's a very minor thing. In particular, we need a message and public key to verify against which may not be available and complicate things.

Maybe I commented on the wrong line. I definitely wanted to suggest to call `schnorrsig_extract` in the randomized "correctness" tests (and not the test vector tests) where we create a schnorrsig via adapt (such as `test_schnorr_adaptor_adapt_extract_sec`). So imo we can resolve this convo.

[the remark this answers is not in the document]
ZhePang: Will add schnorrsig calls in the tests

include/secp256k1_schnorr_adaptor.h
```diff
@@ -57,7 +57,7 @@ SECP256K1_API const secp256k1_adaptor_nonce_function_hardened secp256k1_nonce_fu
  *  Out:   presig65: pointer to a 65-byte array to store the adaptor signature.
  *  In:       msg32: the 32-byte message being signed.
  *          keypair: pointer to an initialized keypair.
- *        adaptor33: pointer to a 33-byte compressed adaptor point.
+ *        adaptor: pointer to an adaptor point.
```
siv2r, line 60: nit: 344620e renamed `adaptor` to `adaptor33`. Just wanna make sure that this revert is intentional.

siv2r (commented):
There were a few warnings during compilation, but all tests passed successfully.

<details>
<summary>warnings</summary>

```bash
src/ctime_tests.c: In function ‘run_tests’:
src/ctime_tests.c:212:74: warning: passing argument 5 of ‘secp256k1_schnorr_adaptor_presign’ from incompatible pointer type [-Wincompatible-pointer-types]
  212 |         ret = secp256k1_schnorr_adaptor_presign(ctx, sig, msg, &keypair, t, NULL);
      |                                                                          ^
      |                                                                          |
      |                                                                          unsigned char *
In file included from src/ctime_tests.c:35:
src/../include/secp256k1_schnorr_adaptor.h:72:29: note: expected ‘const secp256k1_pubkey *’ {aka ‘const struct <anonymous> *’} but argument is of type ‘unsigned char *’
   72 |     const secp256k1_pubkey *adaptor,
      |     ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
In file included from src/secp256k1.c:896,
                 from src/bench_ecmult.c:8:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/secp256k1.c:896,
                 from src/tests.c:20:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/secp256k1.c:896,
                 from src/bench_internal.c:8:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/secp256k1.c:896,
                 from src/tests_exhaustive.c:23:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/secp256k1.c:896,
                 from src/tests.c:20:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/secp256k1.c:896:
src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
  197 |     size_t size = 33;
      |            ^~~~
In file included from src/tests.c:7474:
src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_api’:
src/modules/schnorr_adaptor/tests_impl.h:141:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
  141 |     secp256k1_ec_pubkey_parse(CTX, &t, adaptor33, 33);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/tests.c:7474:
src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_vectors_check_presigning’:
src/modules/schnorr_adaptor/tests_impl.h:180:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
  180 |     secp256k1_ec_pubkey_parse(CTX, &adaptor, adaptor33, 33);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/tests.c:7474:
src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_api’:
src/modules/schnorr_adaptor/tests_impl.h:141:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
  141 |     secp256k1_ec_pubkey_parse(CTX, &t, adaptor33, 33);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from src/tests.c:7474:
src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_vectors_check_presigning’:
src/modules/schnorr_adaptor/tests_impl.h:180:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
  180 |     secp256k1_ec_pubkey_parse(CTX, &adaptor, adaptor33, 33);
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
</details>

ZhePang:
> There were a few warnings during compilation, but all tests passed successfully.
> 
> warnings
> ```shell
> src/ctime_tests.c: In function ‘run_tests’:
> src/ctime_tests.c:212:74: warning: passing argument 5 of ‘secp256k1_schnorr_adaptor_presign’ from incompatible pointer type [-Wincompatible-pointer-types]
>   212 |         ret = secp256k1_schnorr_adaptor_presign(ctx, sig, msg, &keypair, t, NULL);
>       |                                                                          ^
>       |                                                                          |
>       |                                                                          unsigned char *
> In file included from src/ctime_tests.c:35:
> src/../include/secp256k1_schnorr_adaptor.h:72:29: note: expected ‘const secp256k1_pubkey *’ {aka ‘const struct <anonymous> *’} but argument is of type ‘unsigned char *’
>    72 |     const secp256k1_pubkey *adaptor,
>       |     ~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~
> In file included from src/secp256k1.c:896,
>                  from src/bench_ecmult.c:8:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/secp256k1.c:896,
>                  from src/tests.c:20:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/secp256k1.c:896,
>                  from src/bench_internal.c:8:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/secp256k1.c:896,
>                  from src/tests_exhaustive.c:23:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/secp256k1.c:896,
>                  from src/tests.c:20:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/secp256k1.c:896:
> src/modules/schnorr_adaptor/main_impl.h: In function ‘secp256k1_schnorr_adaptor_extract’:
> src/modules/schnorr_adaptor/main_impl.h:197:12: warning: unused variable ‘size’ [-Wunused-variable]
>   197 |     size_t size = 33;
>       |            ^~~~
> In file included from src/tests.c:7474:
> src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_api’:
> src/modules/schnorr_adaptor/tests_impl.h:141:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
>   141 |     secp256k1_ec_pubkey_parse(CTX, &t, adaptor33, 33);
>       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> In file included from src/tests.c:7474:
> src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_vectors_check_presigning’:
> src/modules/schnorr_adaptor/tests_impl.h:180:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
>   180 |     secp256k1_ec_pubkey_parse(CTX, &adaptor, adaptor33, 33);
>       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> In file included from src/tests.c:7474:
> src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_api’:
> src/modules/schnorr_adaptor/tests_impl.h:141:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
>   141 |     secp256k1_ec_pubkey_parse(CTX, &t, adaptor33, 33);
>       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> In file included from src/tests.c:7474:
> src/modules/schnorr_adaptor/tests_impl.h: In function ‘test_schnorr_adaptor_vectors_check_presigning’:
> src/modules/schnorr_adaptor/tests_impl.h:180:5: warning: ignoring return value of ‘secp256k1_ec_pubkey_parse’, declared with attribute warn_unused_result [-Wunused-result]
>   180 |     secp256k1_ec_pubkey_parse(CTX, &adaptor, adaptor33, 33);
>       |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

I probably forgot to delete some unused variables. And ec_pubkey_parse returns a 0/1 value, I'll put it into a CHECK

About the time_test warning, I forgot to change the code there and it might takes the original input format.

jonasnick:
Replaced by #299. Thanks @ZhePang for your work!

PR #275, opened 2023-11-03
add cmake gears to enable ECDSA sign-to-contract - rangeproof - surjectionproof - whitelist - generator modules

lightyear15:


real-or-random:
Thanks! 

I think this should also cover module dependencies, i.e., some modules should be enabled automatically if others are enabled. See a few lines above your changes in the root `CMakeLists.txt` for an example where schnorrsig depends on extrakeys, and see https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/configure.ac#L447-L449 for all dependencies.

And if you're at it, would you mind adding all modules to CMake? The remaining ones shouldn't be harder to add. (If not, no worries, we can do it in a separate PR.)

lightyear15:
thanks @real-or-random  for the feedback

I think I added all the modules.
Sorry if I reshuffled the existing ones but I thought it would be easier to review if cmake modules declarations mirrored the autoconf ones.

Also, I think that splitting option declarations and option dependencies into two separate sections makes code cleaner

lightyear15:
@real-or-random ping :smiley_cat:

real-or-random:
This looks good, but since we're a fork of libsecp256k1, we tend to stay close to their code to avoid conflicts and additional work when we rebase on libsecp256k1s master. (So it's for example not a good idea to add a trailing "." to the existing options.)  When I reviewed this PR last week, my first thought was "let me just fixup your PR a bit, that's quicker done than explained", but then it turned out to be more complex than I expected. 

But you're totally right, it's cleaner to declare all the module options first and only then process them. https://github.com/bitcoin-core/secp256k1/pull/1482 does this upstream. Let's see, if we can get it quickly it, we can cherry-pick it here and then rebase this PR on top of the result. I can take care of this then, I have some local changes already.

jonasnick:
The upstream commit is has been merged.

lightyear15:
I rebased my branch
it seems there is a bit still missing: exporting header files based on the enabled modules

real-or-random:
See #288.

PR #277, opened 2023-11-17
generator: Save a fe_mul in shallue_van_de_woestijne

real-or-random:


roconnor-blockstream (approved):
LGTM

I ran some randomized testing comparing this to my Haskell implementation.

real-or-random:
> Arguably this whole function should be rewritten. Unfortunately that is risky, so maybe it is better to keep doing these micro-improvements.

You have a point, though. I mean, sure, let's merge this if we get the ACKs, but a proper rewrite is probably a better approach.

real-or-random:
> Haskell implementation.

Is that a full implementation of the SWU map? I guess we could then steal the algorithm.

roconnor-blockstream:
See #286 for my rewrite of this function, which would supersede this PR.

PR #278, opened 2023-11-23
FROST Trusted Dealer

jesseposner:
This PR implements a BIP-340 compatible threshold signature system based on FROST (Flexible Round-Optimized Schnorr Threshold Signatures) with a trusted dealer. This PR does not implement distributed key generation (DKG).

### FROST Paper

This PR is based upon the FROST [paper](https://crysp.uwaterloo.ca/software/frost/) by Chelsea Komlo and Ian Goldberg and the [draft RFC](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-frost-08).

### FROST Signing BIP

This PR implements the FROST signing [BIP](https://github.com/siv2r/bip-frost-signing).

### Prior Work

This PR is patterned on the APIs, and in many instances duplicates code, from the MuSig2 implementation due to their similarities in nonce generation and signing.


include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+ *  `frost_share_serialize` and `frost_share_parse`.
+ */
+typedef struct {
+    unsigned char data[36];
+} secp256k1_frost_share;
+
+/** Serialize a FROST share
+ *
+ *  Returns: 1 when the share could be serialized, 0 otherwise
+ *  Args:    ctx: a secp256k1 context object
```
real-or-random, line 161: nit: see https://github.com/bitcoin-core/secp256k1/pull/1476/commits/aa3dd5280b4a046c03bd344bfd7f1499199a1f3c for consistent wording here

include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+
+/** Creates key generation shares
+ *
+ *  To generate a key, a trusted dealer generates a share for each other
+ *  participant.
+ *
+ *  Each participant _must_ have a secure channel with the trusted dealer with
+ *  which they can transmit shares to each other.
+ *
+ *  A new seed32 _must_ be used for each key generation session. The trusted
```
real-or-random, line 192: nit: s/new/fresh 

By the way, we need to say similar things for musig sessions. Feel free to steal some sentences/phrasing ideas from https://github.com/bitcoin-core/secp256k1/pull/1479.

include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+ *  dealer must NOT REUSE their respective seed32 again for subsequent key
+ *  generation sessions. If a trusted dealer fails to complete this session or
+ *  start a new session to generate a new key, they must NOT REUSE their
+ *  respective seed32 again, but instead generate a new one. It is recommended
```
real-or-random, line 196: s/again//  ("reuse again" is saying the same thing twice)

include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+ *  Out:   share: pointer to a share object
+ *  In:     in32: pointer to the 32-byte share to be parsed
+ */
+SECP256K1_API int secp256k1_frost_share_parse(
+    const secp256k1_context *ctx,
+    secp256k1_frost_share *share,
+    const unsigned char *in32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3);
+
+/** Creates key generation shares
```
real-or-random, line 184: "key shares" or "key generation" shares?

include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+ *
+ *  To generate a key, a trusted dealer generates a share for each other
+ *  participant.
+ *
+ *  Each participant _must_ have a secure channel with the trusted dealer with
+ *  which they can transmit shares to each other.
```
real-or-random, line 190: Hm, this may be slightly confusing because one could read it as "must have established a secure channel already", but that's of course not necessary. Maybe: "The trusted dealer must transmit shares over secure channels to participants."

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,146 @@
...
+
+    /* Commit to threshold, n_participants, and seed */
+    secp256k1_sha256_initialize_tagged(&sha, (unsigned char*)"FROST/trusted-shares-polygen", sizeof("FROST/trusted-shares-polygen") - 1);
+    secp256k1_sha256_write(&sha, seed32, 32);
+    secp256k1_write_be64(&polygen[0], threshold);
+    secp256k1_write_be64(&polygen[8], n_participants);
+    secp256k1_sha256_write(&sha, polygen, 16);
+    secp256k1_sha256_finalize(&sha, polygen);
+
+    /* Derive share */
```
real-or-random, line 127: s/share/shares

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,146 @@
...
+    /* See draft-irtf-cfrg-frost-08#appendix-C.1 */
+    for (i = 0; i < n_participants; i++) {
+        secp256k1_scalar share_i, idx;
+
+        secp256k1_scalar_clear(&share_i);
```
real-or-random, line 132: This function is just for nuking memory. Use `_scalar_set_int(0)` to set to zero. (We should probably have a dedicated function, but we don't have one currently.)

include/secp256k1_frost.h
```diff
@@ -17,6 +17,84 @@ extern "C" {
...
+ *  generation sessions. If a trusted dealer fails to complete this session or
+ *  start a new session to generate a new key, they must NOT REUSE their
+ *  respective seed32 again, but instead generate a new one. It is recommended
+ *  to always choose seed32 uniformly at random to avoid their reuse.
+ *
+ *  Returns: 0 if the arguments are invalid, 1 otherwise
+ *  Args:            ctx: pointer to a context object
+ *  Out:          shares: pointer to the key generation shares
+ *             pubshares: pointer to the public verification shares
+ *                    pk: pointer to the x-only public key
```
real-or-random, line 203: Should this be an x-only key? I'm not entirely sure, I would need to think about it.

real-or-random (commented):
some initial comments, more tomorrow

real-or-random:
In general, I think it would be nice to get this rebase this. (Should be mostly trivial.) Also, CMake support is missing because we have this now for all other -zkp modules (https://github.com/BlockstreamResearch/secp256k1-zkp/pull/288 ), but that's currently not essential.

include/secp256k1_frost.h
jesseposner: The function [conditionally inverts the shares](https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/files/a6af50e9a4dc70768108278438c391900038d980..cf31f983ff6677e8f36aec542d03f585eaeb13be#diff-65b267bfb41db84f9bf003641ec64ef27687583f9692adb7869f561f14da6d08R134) to guarantee that the private key will match the x-only key.

Since the primary use-case will be on-chain signing, I think it's a useful simplification to take care of the negation during share generation and makes it easier to reason about in [secp256k1_frost_pubkey_get](https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/772a0386c1a5f250c974efa6615972cd043f7d7f#diff-65b267bfb41db84f9bf003641ec64ef27687583f9692adb7869f561f14da6d08R187) and [secp256k1_frost_pubkey_tweak](https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/772a0386c1a5f250c974efa6615972cd043f7d7f#diff-65b267bfb41db84f9bf003641ec64ef27687583f9692adb7869f561f14da6d08R203).

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,650 @@
...
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(session != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(pubnonces != NULL);
+    ARG_CHECK(ids != NULL);
+    ARG_CHECK(n_pubnonces > 1);
+    ARG_CHECK(my_id != 0);
+    for (i = 0; i < n_pubnonces; i++) {
+        ARG_CHECK(ids[i] != 0);
```
siv2r, line 427: should we also `ARG_CHECK` for `1 <= ids[i] <= max_participant`?

include/secp256k1_frost.h
```diff
@@ -0,0 +1,606 @@
...
+SECP256K1_API int secp256k1_frost_nonce_gen(
+    const secp256k1_context *ctx,
+    secp256k1_frost_secnonce *secnonce,
+    secp256k1_frost_pubnonce *pubnonce,
+    const unsigned char *session_id32,
+    const secp256k1_frost_share *agg_share,
+    const unsigned char *msg32,
+    const secp256k1_xonly_pubkey *agg_pk,
+    const unsigned char *extra_input32
+) SECP256K1_ARG_NONNULL(1) SECP256K1_ARG_NONNULL(2) SECP256K1_ARG_NONNULL(3) SECP256K1_ARG_NONNULL(4);
```
siv2r, line 387: We can also add `pubshare` as an input argument. MuSig2 hashes the individual pubkey when generating the nonces, but we don't need to make `pubshare` a mandatory argument like MuSig2 does ([more info](https://github.com/siv2r/bip-frost-signing/blob/master/design_decisions.md#no-mandatory-pk-in-nonce-generation)).

How do we decide the args supplied to the hash function for generating a nonce? Should we throw in all available signer information into the hash function? For instance, can we include the participant identifier too?

src/modules/frost/session_impl.h
jesseposner: The function doesn't currently take `max_participant` as an input, and I'm not sure if it's worth requiring the extra state to perform this check.

include/secp256k1_frost.h
jesseposner: Upon further reflection, I'm realizing it's not a useful simplification because in practice the output of the DKG will typically be further tweaked by a BIP32 tweak and then a taproot tweak, rather than used directly.

So I now think it would be better to have the DKG output a 33-byte key rather than an x-only key.

configure.ac
```diff
@@ -470,6 +475,14 @@ if test x"$enable_module_ecdsa_adaptor" = x"yes"; then
   SECP_CONFIG_DEFINES="$SECP_CONFIG_DEFINES -DENABLE_MODULE_ECDSA_ADAPTOR=1"
 fi
 
+if test x"$enable_module_frost" = x"yes"; then
+  if test x"$enable_module_schnorrsig" = x"no"; then
+    AC_MSG_ERROR([Module dependency error: You have disabled the schnorrsig module explicitly, but it is required by the musig module.])
```
jesseposner, line 480: `%s/musig/frost`

include/secp256k1_frost.h
jesseposner: I've taken a stab at implementing this change, however, it requires tracking more state during signing.

Currently, `nonce_process` takes an xonly `agg_pk` and optionally a `tweak_cache` and returns a `session`. `partial_sign` takes a `session` and optionally a `tweak_cache`.

`partial_sign` only needs to handle negation logic if a `tweak_cache` is present because the DKG outputs an x-only key. If we change the DKG to output a 33-byte key, then `nonce_process` will need to take a 33-byte `agg_pk` and it will need to save a parity bit, serialized as a byte, in the `session`, so that `partial_sign` can handle the negation logic for when a `tweak_cache` is not present.

So while I don't think there's any reason for the protocol to output a 33-byte key, it seems useful to have the implementation do so, because it avoids a serialized byte in the `session` and I don't see a downside to doing so.

jesseposner: I need to compare this approach to the one taken in the signing BIP. The BIP proposes that the pubshares be saved to the `session` and the 33 byte public key is derived from the pubshares.

jesseposner: https://github.com/siv2r/bip-frost-signing/issues/12

jesseposner: Perhaps the best reason to not output an xonly key from the DKG is that we don't want to encourage FROST DKG outputs being used directly on-chain without an unspendable script path because unlike MuSig, FROST does not randomize the group public key: https://github.com/bitcoin/bips/blob/master/bip-0341.mediawiki#cite_note-23

configure.ac
```diff
@@ -579,6 +592,9 @@ else
   if test x"$enable_module_generator" = x"yes"; then
     AC_MSG_ERROR([NUMS generator module is experimental. Use --enable-experimental to allow.])
   fi
+  if test x"$enable_module_frost" = x"yes"; then
+    AC_MSG_ERROR([FROST module is experimental. Use --enable-experimental to allow.])
+  fi
```
siv2r, line 597: nit: when we execute `./configure --enable-module-frost`, the following error appears:
```bash
configure: error: MuSig module is experimental. Use --enable-experimental to allow.
``` 
instead of:
```
configure: error: FROST module is experimental. Use --enable-experimental to allow.
```
----
To resolve this issue, we should reposition the FROST module check before the MuSig experimental module check (lines 486-488).

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,146 @@
...
+    /* Commit to threshold, n_participants, and seed */
+    secp256k1_sha256_initialize_tagged(&sha, (unsigned char*)"FROST/trusted-shares-polygen", sizeof("FROST/trusted-shares-polygen") - 1);
+    secp256k1_sha256_write(&sha, seed32, 32);
+    secp256k1_write_be64(&polygen[0], threshold);
+    secp256k1_write_be64(&polygen[8], n_participants);
+    secp256k1_sha256_write(&sha, polygen, 16);
+    secp256k1_sha256_finalize(&sha, polygen);
+
+    /* Derive share */
+    /* See draft-irtf-cfrg-frost-08#appendix-C.1 */
```
siv2r, line 128: nit: we can update this to refer to the completed RFC: https://datatracker.ietf.org/doc/rfc9591/

include/secp256k1_frost.h
siv2r: The signing BIP now [uses signer public shares](https://github.com/siv2r/bip-frost-signing/commit/3f23cf9f2f2177c00048782c93b90caa84b61ca8) to initialize the tweak context (name to be updated) rather than the group public key. Hence, it no longer depends on whether the group pubkey (produced by keygen) is plain or x-only.

siv2r: An alternative design choice here is for `trusted_gen` to always output `tweak_cache` and optionally `group_pk`. This would be similar to how `musig_pubkey_agg` [outputs `keyagg_cache` and optionally `agg_pk`](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/include/secp256k1_musig.h#L203-L207).

src/group.h
```diff
@@ -202,6 +202,10 @@ static void secp256k1_ge_from_bytes(secp256k1_ge *r, const unsigned char *buf);
  */
 static int secp256k1_ge_is_in_correct_subgroup(const secp256k1_ge* ge);
 
+static void secp256k1_point_save_ext(unsigned char *data, secp256k1_ge *ge);
+
+static void secp256k1_point_load_ext(secp256k1_ge *ge, const unsigned char *data);
```
siv2r, line 207: The libsecp repo defines these functions [using different names](https://github.com/bitcoin-core/secp256k1/pull/1479/files#diff-28d9b92418cbddbe487fe84847993cc5da0b096363c83631a525cff25189752cR179-R191):
- `point_save_ext` -> `ge_to_bytes_ext`
- `point_load_ext` -> `ge_from_bytes_ext`

include/secp256k1_frost.h
```diff
@@ -95,6 +153,59 @@ SECP256K1_API int secp256k1_frost_shares_trusted_gen(
...
+ *     it is recommended to always choose session_id32 uniformly at random.
+ *  2. If you already know the seckey, message or aggregate public key, they
+ *     can be optionally provided to derive the nonce and increase
+ *     misuse-resistance. The extra_input32 argument can be used to provide
+ *     additional data that does not repeat in normal scenarios, such as the
+ *     current time.
+ *  3. Avoid copying (or serializing) the secnonce. This reduces the possibility
+ *     that it is used more than once for signing.
+ *
+ *  Remember that nonce reuse will leak the secret key!
```
siv2r, line 359: Would it be clearer to say "will leak the agg_share" instead?

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,233 @@
...
+    /* The group elements can not be infinity because they were just parsed */
+    secp256k1_frost_pubnonce_save(nonce, ge);
+    return 1;
+}
+
+static void secp256k1_nonce_function_frost(secp256k1_scalar *k, const unsigned char *session_id, const unsigned char *msg32, const unsigned char *key32, const unsigned char *pk32, const unsigned char *extra_input32) {
+    secp256k1_sha256 sha;
+    unsigned char seed[32];
+    unsigned char i;
+    enum { n_extra_in = 4 };
```
siv2r, line 210:
```suggestion
    enum { N_EXTRA_IN = 4 };
```

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,233 @@
...
+    extra_in[0] = msg32;
+    extra_in[1] = key32;
+    extra_in[2] = pk32;
+    extra_in[3] = extra_input32;
+    for (i = 0; i < n_extra_in; i++) {
+        unsigned char len;
+        if (extra_in[i] != NULL) {
+            len = 32;
+            secp256k1_sha256_write(&sha, &len, 1);
+            secp256k1_sha256_write(&sha, extra_in[i], 32);
```
siv2r, line 228: I like this approach to hashing optional arguments—it’s clearer than [musig’s approach of using a helper function](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/musig/session_impl.h#L303-L310). However, there's a minor issue: the signing BIP encodes the message length as 8 bytes, but here we use 1 byte.

As stated in the [BIP's noncegen:](https://github.com/siv2r/bip-frost-signing?tab=readme-ov-file#nonce-generation-1)
> - If the optional argument _m_ is not present:
>    - Let _m_prefixed = bytes(1, 0)_
>- Else:
>    - Let _m_prefixed = bytes(1, 1) || bytes(8, len(m)) || m_

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,233 @@
...
+        if (!secp256k1_ge_is_in_correct_subgroup(&ge[i])) {
+            return 0;
+        }
+    }
+    /* The group elements can not be infinity because they were just parsed */
+    secp256k1_frost_pubnonce_save(nonce, ge);
+    return 1;
+}
+
+static void secp256k1_nonce_function_frost(secp256k1_scalar *k, const unsigned char *session_id, const unsigned char *msg32, const unsigned char *key32, const unsigned char *pk32, const unsigned char *extra_input32) {
```
siv2r, line 206: Just a note for the future: when writing tests, we should skip two noncegen test vectors—[(1) empty message and (2) 38-byte message](https://github.com/siv2r/bip-frost-signing/blob/master/reference/vectors/nonce_gen_vectors.json#L15-L34)—since our nonce generation currently only supports 32-byte messages.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,233 @@
...
+        } else {
+            len = 0;
+            secp256k1_sha256_write(&sha, &len, 1);
+        }
+    }
+    secp256k1_sha256_finalize(&sha, seed);
+
+    for (i = 0; i < 2; i++) {
+        unsigned char buf[32];
+        secp256k1_sha256_initialize(&sha);
```
siv2r, line 238: Our method of hashing arguments deviates from the BIP. We compute the nonces as
```math
k_i = \mathrm{H}_{\text{no-tag}}(\mathrm{H}_{\text{FROST/nonce}}(\text{optional args}) \;||\; i - 1)
```
whereas the BIP computes them as
```math
k_i = \mathrm{H}_{\text{FROST/nonce}}(\text{optional args} \;||\; i - 1)
```

src/modules/frost/session_impl.h
```diff
@@ -230,4 +273,192 @@ int secp256k1_frost_nonce_gen(const secp256k1_context* ctx, secp256k1_frost_secn
...
+               de-/serializing) would require a lot of extra code complexity.
+               Instead, we set the aggregate nonce to some arbitrary point (the
+               generator). This is secure, because it only restricts the
+               abilities of the attacker: an attacker that forces the sum of
+               nonces to be infinity by sending some maliciously generated nonce
+               pairs can be turned into an attacker that forces the sum to be
+               the generator (by simply adding the generator to one of the
+               malicious nonces), and this does not change the winning condition
+               of the EUF-CMA game. */
+            aggnonce_pt[i] = secp256k1_ge_const_g;
```
siv2r, line 462: I believe `aggnonce_pt[i]` is allowed to be infinity. The [*NonceAgg*](https://github.com/siv2r/bip-frost-signing?tab=readme-ov-file#nonce-aggregation) specification serializes them using `cybtes_ext`, so we can use `ge_to_bytes_ext` here.

Setting the nonce to the generator point only occurs with [the final nonce](https://github.com/siv2r/bip-frost-signing?tab=readme-ov-file#session-context).

src/modules/frost/session_impl.h
```diff
@@ -230,4 +273,192 @@ int secp256k1_frost_nonce_gen(const secp256k1_context* ctx, secp256k1_frost_secn
...
+        for (j = 0; j < 2; j++) {
+            secp256k1_gej_add_ge_var(&summed_nonces[j], &summed_nonces[j], &nonce_pt[j], NULL);
+        }
+    }
+    return 1;
+}
+
+/* TODO: consider updating to frost-08 to address maleability at the cost of performance */
+/* See https://github.com/cfrg/draft-irtf-cfrg-frost/pull/217 */
+static int secp256k1_frost_compute_noncehash(const secp256k1_context* ctx, unsigned char *noncehash, const unsigned char *msg, const secp256k1_frost_pubnonce * const* pubnonces, size_t n_pubnonces, const unsigned char *pk32, const size_t *ids) {
```
siv2r, line 337: FROST3 computes the nonce coefficient as $b = \mathrm{H}_{\text{FROST/noncecoef}}(T || \widetilde{R_1} || \widetilde{R_2} || \widetilde{X} || msg)$, where $T$ is the set of signer ids.

siv2r: We can remove the `ctx` object from the function args.

src/modules/frost/session_impl.h
```diff
@@ -230,4 +273,192 @@ int secp256k1_frost_nonce_gen(const secp256k1_context* ctx, secp256k1_frost_secn
...
+        }
+        secp256k1_sha256_write(&sha, buf, sizeof(buf));
+    }
+    secp256k1_sha256_write(&sha, pk32, 32);
+    secp256k1_sha256_write(&sha, msg, 32);
+    secp256k1_sha256_finalize(&sha, noncehash);
+    return 1;
+}
+
+static int secp256k1_frost_nonce_process_internal(const secp256k1_context* ctx, int *fin_nonce_parity, unsigned char *fin_nonce, secp256k1_scalar *b, secp256k1_gej *aggnoncej, const unsigned char *msg, const secp256k1_frost_pubnonce * const* pubnonces, size_t n_pubnonces, const unsigned char *pk32, const size_t *ids) {
```
siv2r, line 360: We can remove the `ctx` object from the function args

src/modules/frost/session_impl.h
```diff
@@ -230,4 +273,192 @@ int secp256k1_frost_nonce_gen(const secp256k1_context* ctx, secp256k1_frost_secn
...
+    if (!secp256k1_frost_compute_noncehash(ctx, noncehash, msg, pubnonces, n_pubnonces, pk32, ids)) {
+        return 0;
+    }
+    /* fin_nonce = aggnonce[0] + b*aggnonce[1] */
+    secp256k1_scalar_set_b32(b, noncehash, NULL);
+    secp256k1_ecmult(&fin_nonce_ptj, &aggnoncej[1], b, NULL);
+    secp256k1_gej_add_ge_var(&fin_nonce_ptj, &fin_nonce_ptj, &aggnonce[0], NULL);
+    secp256k1_ge_set_gej(&fin_nonce_pt, &fin_nonce_ptj);
+
+    if (secp256k1_ge_is_infinity(&fin_nonce_pt)) {
```
siv2r, line 377: The BIP sets the final nonce to the generator point if it’s infinity.

src/modules/frost/session_impl.h
```diff
@@ -230,4 +273,192 @@ int secp256k1_frost_nonce_gen(const secp256k1_context* ctx, secp256k1_frost_secn
...
+        if (!secp256k1_scalar_is_zero(&cache_i.tweak)) {
+            secp256k1_scalar e_tmp;
+            secp256k1_scalar_mul(&e_tmp, &session_i.challenge, &cache_i.tweak);
+            if (secp256k1_fe_is_odd(&cache_i.pk.y)) {
+                secp256k1_scalar_negate(&e_tmp, &e_tmp);
+            }
+            secp256k1_scalar_add(&session_i.s_part, &session_i.s_part, &e_tmp);
+        }
+    }
+    /* Update the challenge by multiplying the Lagrange coefficient to prepare
```
siv2r, line 497: I hadn’t considered this option when writing the BIP. So, right now, it computes $\lambda_{i, T}$ during *Sign*, not in *GetSessionValue*. Happy to update the BIP if this approach is better.

include/secp256k1_frost.h
```diff
@@ -335,6 +345,135 @@ SECP256K1_API int secp256k1_frost_nonce_gen(
...
+ */
+SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_frost_nonce_process(
+    const secp256k1_context *ctx,
+    secp256k1_frost_session *session,
+    const secp256k1_frost_pubnonce * const *pubnonces,
+    size_t n_pubnonces,
+    const unsigned char *msg32,
+    const secp256k1_xonly_pubkey *agg_pk,
+    size_t my_id,
+    const size_t *ids,
```
siv2r, line 420: In the documentation, we should probably mention that [1] the `ids` array must be unique, and [2] `my_id` must be included in `ids`. Otherwise, Lagrange interpolation will return incorrect values.

We could also implement `ARG_CHECK`s for these, but that might be a lot of work.

include/secp256k1_frost.h
```diff
@@ -335,6 +345,135 @@ SECP256K1_API int secp256k1_frost_nonce_gen(
...
+
+/** Verifies that the adaptor can be extracted by combining the adaptor
+ *  pre-signature and the completed signature.
+ *
+ *  Returns: 0 if the arguments are invalid or the adaptor signature does not
+ *           verify, 1 otherwise
+ *  Args:         ctx: pointer to a context object
+ *  In:     pre_sig64: 64-byte pre-signature
+ *              msg32: the 32-byte message being verified
+ *             pubkey: pointer to an x-only public key to verify with
```
siv2r, line 535:
```suggestion
 *             agg_pk: pointer to an x-only public key to verify with
```

include/secp256k1_frost.h
```diff
@@ -335,6 +345,135 @@ SECP256K1_API int secp256k1_frost_nonce_gen(
...
+ *            adaptor: pointer to the adaptor point being verified
+ *       nonce_parity: the output of `frost_nonce_parity` called with the
+ *                     session used for producing the pre-signature
+ */
+SECP256K1_API int secp256k1_frost_verify_adaptor(
```
siv2r, line 540: Just pointing out that there's another possible approach here. In [single signer Schnorr adaptor](https://github.com/BlockstreamResearch/secp256k1-zkp/pull/268/files#diff-5b4fed35fac8b6fb8210bd7be8c64663ae9a227e702fb5e7d9f2a02fe2a505d3R85), we use the `extract_adaptor_point` API instead of `verify_adaptor`—[more details](https://excalidraw.com/#room=77bd1b33053edad36dff,hTvbUv8McjWYHvp-mGZv3w). Interestingly, [MuSig](https://github.com/BlockstreamResearch/secp256k1-zkp/blob/master/src/modules/musig/adaptor_impl.h) doesn’t include either `extract_adaptor_point` or `verify_adaptor` (not sure why).

include/secp256k1_frost.h
siv2r: In BIP, the _DeriveInterpolatingValue_ function fails if both these conditions aren't met.

src/modules/frost/session_impl.h
```diff
@@ -461,4 +505,154 @@ int secp256k1_frost_nonce_process(const secp256k1_context* ctx, secp256k1_frost_
...
+
+    if (tweak_cache != NULL) {
+        secp256k1_tweak_cache_internal cache_i;
+        if (!secp256k1_tweak_cache_load(ctx, &cache_i, tweak_cache)) {
+            secp256k1_frost_partial_sign_clear(&sk, k);
+            return 0;
+        }
+        if (secp256k1_fe_is_odd(&cache_i.pk.y) != cache_i.parity_acc) {
+            secp256k1_scalar_negate(&sk, &sk);
+        }
```
siv2r, line 555: We can use this comment here (taken from musig)

```
    /* Negate sk if secp256k1_fe_is_odd(&cache_i.pk.y)) XOR cache_i.parity_acc.
     * This corresponds to the line "Let d = g⋅gacc⋅d' mod n" in the
     * specification. */
```

src/modules/frost/session_impl.h
```diff
@@ -461,4 +505,154 @@ int secp256k1_frost_nonce_process(const secp256k1_context* ctx, secp256k1_frost_
...
+
+int secp256k1_frost_partial_sig_agg(const secp256k1_context* ctx, unsigned char *sig64, const secp256k1_frost_session *session, const secp256k1_frost_partial_sig * const* partial_sigs, size_t n_sigs) {
+    size_t i;
+    secp256k1_frost_session_internal session_i;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(sig64 != NULL);
+    ARG_CHECK(session != NULL);
+    ARG_CHECK(partial_sigs != NULL);
+    ARG_CHECK(n_sigs > 0);
```
siv2r, line 641:
```suggestion
    ARG_CHECK(n_sigs > 1);
```
Should we do this instead? 
The `frost_nonce_process` currently has `ARG_CHECK(n_pubnonce > 1)`.  We might want to use the same value in both functions, either 1 or 0.

examples/frost.c
siv2r, line 1: In musig's example, every function (except `main`) is `static`

examples/frost.c
```diff
@@ -0,0 +1,224 @@
...
+            /* Check if signer has already been assigned */
+            if (!is_signer[signer_id]) {
+                is_signer[signer_id] = 1;
+                signers[i] = signer_id;
+                break;
+            }
+        }
+        /* Mark signer as assigned */
+        pubnonces[i] = &signer[signer_id].pubnonce;
+        /* pubkeys[i] = &signer[signer_id].pubkey; */
```
siv2r, line 146: I believe this comment was meant to be removed

src/modules/frost/tests_impl.h
siv2r, line 1: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278#discussion_r1720955395 applies here too.

src/modules/frost/tests_impl.h
```diff
@@ -0,0 +1,738 @@
...
+    frost_api_tests();
+    frost_nonce_test();
+    for (i = 0; i < COUNT; i++) {
+        /* Run multiple times to ensure that pk and nonce have different y
+         * parities */
+        frost_tweak_test();
+    }
+    for (i = 0; i < COUNT; i++) {
+        frost_multi_hop_lock_tests();
+    }
```
siv2r, line 735: We can move this function call to the previous loop

src/modules/frost/tests_impl.h
```diff
@@ -0,0 +1,738 @@
...
+        secp256k1_ge_neg(&ge[0], &ge[0]);
+        secp256k1_ge_neg(&ge[1], &ge[1]);
+    }
+
+    secp256k1_frost_sum_nonces(CTX, summed_nonces, pubnonce_ptr, 2);
+    CHECK(secp256k1_gej_is_infinity(&summed_nonces[0]));
+    CHECK(secp256k1_gej_is_infinity(&summed_nonces[1]));
+}
+
+int frost_memcmp_and_randomize(unsigned char *value, const unsigned char *expected, size_t len) {
```
siv2r, line 92: The same function is also used in MuSig's tests, so we can make it a helper function in `src/testutil.h`.

configure.ac
jesseposner: Can you try again? I wasn't able to reproduce the issue. 

When I run `./configure --enable-module-frost` I get `configure: error: FROST module is experimental. Use --enable-experimental to allow.`.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,233 @@
...
+
+static void secp256k1_nonce_function_frost(secp256k1_scalar *k, const unsigned char *session_id, const unsigned char *msg32, const unsigned char *key32, const unsigned char *pk32, const unsigned char *extra_input32) {
+    secp256k1_sha256 sha;
+    unsigned char seed[32];
+    unsigned char i;
+    enum { n_extra_in = 4 };
+    const unsigned char *extra_in[n_extra_in];
+
+    /* TODO: this doesn't have the same sidechannel resistance as the BIP340
+     * nonce function because the seckey feeds directly into SHA. */
```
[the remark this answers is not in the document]
jesseposner: I need to bring these changes over from MuSig: https://github.com/BlockstreamResearch/secp256k1-zkp/commit/206017d67d9bb8b21d5cc924ba53e1618274774c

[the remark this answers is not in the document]
jesseposner: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/35f453dfd9690664d0cc50ca2f953bf14d584c94

src/modules/frost/keygen_impl.h
jesseposner: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/64e8beb8d2a9b79e8a1f242da187383e785942a6

src/modules/frost/session_impl.h
jesseposner: I switched over to using the MuSig approach: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/35f453dfd9690664d0cc50ca2f953bf14d584c94

src/modules/frost/session_impl.h
jesseposner: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/35f453dfd9690664d0cc50ca2f953bf14d584c94

src/group.h
jesseposner: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/9fcaf25be7ccd96f6c4da9b292daf6ff4cf10e3b

include/secp256k1_frost.h
jesseposner: https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/commits/ae39a3a377b8b39c4ce1ad58f32115cb77e7cebb

theStack (commented):
Only looked at share generation/verification so far, left a few comments below. Fwiw, I've added CMake support, feel free to pull it in: https://github.com/theStack/secp256k1-zkp/commit/3b72fbd71abafee0aa43ff0a168fd8ae7865f7a1

src/modules/frost/session_impl.h
siv2r: I don't think we need to skip the empty message test vector. The `secp256k1_nonce_function_frost_helper` handles it correctly. It adds the byte `00` to the hash when `msg32 == NULL`, as the BIP specifies.

siv2r: `secp2566k1_nonce_function_frost_helper` doesn't add the prefix of `01` when the message is present, which needs to be updated.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,683 @@
...
+/* TODO: consider updating to frost-08 to address maleability at the cost of performance */
+/* See https://github.com/cfrg/draft-irtf-cfrg-frost/pull/217 */
+static int secp256k1_frost_compute_noncehash(const secp256k1_context* ctx, unsigned char *noncehash, const unsigned char *msg, const secp256k1_frost_pubnonce * const *pubnonces, size_t n_pubnonces, const unsigned char *pk32, const size_t *ids) {
+    unsigned char buf[66];
+    secp256k1_sha256 sha;
+    size_t i;
+
+    secp256k1_frost_compute_noncehash_sha256_tagged(&sha);
+    /* TODO: sort by index */
+    for (i = 0; i < n_pubnonces; i++) {
```
DarkWindman, line 416: The BIP specification defines the function `has_duplicates(id_{1..u})`. This check is crucial for the correct computation of interpolation values. However, the implementation does not enforce this verification.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,683 @@
...
+        secp256k1_scalar idx;
+
+        secp256k1_frost_get_scalar_index(&idx, ids[i]);
+        secp256k1_scalar_get_b32(buf, &idx);
+        secp256k1_sha256_write(&sha, buf, 32);
+        if (!secp256k1_frost_pubnonce_serialize(ctx, buf, pubnonces[i])) {
```
DarkWindman: The implementation uses a different hash input format, namely `ids[i]||pubnonces[i]||ids[i+1]||...`, instead of the `ser_ids||aggnonce...` format specified in the BIP. In addition, as noted in the comments above, the implementation does not sort IDs, as required by the BIP specification.

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+    if (output_pubkey != NULL) {
+        memset(output_pubkey, 0, sizeof(*output_pubkey));
+    }
+    ARG_CHECK(keygen_cache != NULL);
+    ARG_CHECK(tweak32 != NULL);
+
+    if (!secp256k1_keygen_cache_load(ctx, &cache_i, keygen_cache)) {
+        return 0;
+    }
+    secp256k1_scalar_set_b32(&tweak, tweak32, &overflow);
```
DarkWindman, line 406: As I understand it, this function does not perform a 'zero'-tweak check, which can lead to a privacy leak since `Q = Q'`. In contrast, the BIP specification validates this and aborts if the check fails.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,683 @@
...
+    unsigned char pk32[32];
+    secp256k1_scalar l;
+
+    VERIFY_CHECK(ctx != NULL);
+    ARG_CHECK(session != NULL);
+    ARG_CHECK(msg32 != NULL);
+    ARG_CHECK(pubnonces != NULL);
+    ARG_CHECK(ids != NULL);
+    ARG_CHECK(keygen_cache != NULL);
+    ARG_CHECK(n_pubnonces > 1);
```
DarkWindman, line 478: As stated in the specification, the threshold `t` must satisfy `1 <= t <= n`. In this regard, it would be better to verify both boundary conditions.

src/modules/frost/session_impl.h
```diff
@@ -0,0 +1,683 @@
...
+    size_t i;
+    int j;
+
+    secp256k1_gej_set_infinity(&summed_nonces[0]);
+    secp256k1_gej_set_infinity(&summed_nonces[1]);
+
+    for (i = 0; i < n_pubnonces; i++) {
+        secp256k1_ge nonce_pts[2];
+        if (!secp256k1_frost_pubnonce_load(ctx, nonce_pts, pubnonces[i])) {
+            return 0;
```
DarkWindman, line 363: It does not return the blame signer `id_i` for invalid `pubnonce`. A similar issue also applies in the case of an invalid `psig`.

DarkWindman (commented):
Hi! We are new to the secp256k1-zkp library, but we are actively researching bip-frost-signing and have noticed several potential inconsistencies. We would appreciate any feedback to our clarification.

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+    for (i = 0; i < n_participants; i++) {
+        memset(&shares[i], 0, sizeof(shares[i]));
+    }
+    ARG_CHECK(vss_commitment != NULL);
+    ARG_CHECK(seed32 != NULL);
+    ARG_CHECK(threshold > 1);
+    ARG_CHECK(n_participants >= threshold);
+
+    /* Commit to all inputs */
+    secp256k1_sha256_initialize(&sha);
```
Christewart, line 207: Why the hashing here? If i'm following [RFC9591](https://www.rfc-editor.org/rfc/rfc9591.html#appendix-C) correctly it uses the secret directly ?

>  participant_private_keys, coefficients = secret_share_shard(
      **secret_key**, coefficients, MAX_PARTICIPANTS)

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+    secp256k1_sha256_write(&sha, seed32, 32);
+    secp256k1_write_be64(&polygen[0], threshold);
+    secp256k1_write_be64(&polygen[8], n_participants);
+    secp256k1_sha256_write(&sha, polygen, 16);
+    secp256k1_sha256_finalize(&sha, polygen);
+
+    secp256k1_frost_vss_gen(ctx, vss_commitment, polygen, threshold);
```
Christewart, line 214: nit: The RFC does vss gen and share gen in inverted order.

Christewart: nit: Both [`secp256k1_frost_vss_gen`](https://github.com/jesseposner/secp256k1-zkp/blob/frost-trusted-dealer/src/modules/frost/keygen_impl.h#L123) and [`secp256k1_frost_share_gen`](https://github.com/jesseposner/secp256k1-zkp/blob/frost-trusted-dealer/src/modules/frost/keygen_impl.h#L142) rederive the coefficients. Not sure if this is necessary or it would make sense as a parameter.

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+        secp256k1_ge_set_gej(&rp, &rj);
+        secp256k1_pubkey_save(&vss_commitment[threshold - i - 1], &rp);
+    }
+}
+
+static int secp256k1_frost_share_gen(secp256k1_frost_secshare *share, const unsigned char *polygen32, size_t threshold, const size_t id) {
+    secp256k1_scalar idx;
+    secp256k1_scalar share_i;
+    size_t i;
+    int ret = 1;
```
Christewart: It appears `ret = 1` is set, and never modified, so this function always succeeds? It appears that [`secp256k1_frost_shares_gen`](https://github.com/jesseposner/secp256k1-zkp/blob/frost-trusted-dealer/src/modules/frost/keygen_impl.h#L183) expects some validation to be done?

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+    secp256k1_sha256_write(&sha, seed32, 32);
+    secp256k1_write_be64(&polygen[0], threshold);
+    secp256k1_write_be64(&polygen[8], n_participants);
+    secp256k1_sha256_write(&sha, polygen, 16);
+    secp256k1_sha256_finalize(&sha, polygen);
+
+    secp256k1_frost_vss_gen(ctx, vss_commitment, polygen, threshold);
+
+    for (i = 0; i < n_participants; i++) {
+        ret &= secp256k1_frost_share_gen(&shares[i], polygen, threshold, i);
```
Christewart: I believe this will never modify `ret` to `0` because of https://github.com/BlockstreamResearch/secp256k1-zkp/pull/278/changes#r2795329446 , therefore validation in `secp256k1_frost_shares_gen` will always succeed.

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+    /* See RFC 9591, appendix C.1 */
+    secp256k1_scalar_set_int(&share_i, 0);
+    for (i = 0; i < threshold; i++) {
+        secp256k1_scalar coeff_i;
+
+        secp256k1_frost_derive_coeff(&coeff_i, polygen32, i);
+        /* Horner's method to evaluate polynomial to derive shares */
+        secp256k1_scalar_add(&share_i, &share_i, &coeff_i);
+        if (i < threshold - 1) {
+            secp256k1_frost_get_scalar_index(&idx, id);
```
Christewart, line 179: Isn't this the same value on every iteration of the loop? I.e. 

`idx = id +1`

If so couldn't this be moved out of the loop?

src/modules/frost/keygen_impl.h
```diff
@@ -0,0 +1,422 @@
...
+
+    if (!secp256k1_frost_share_load(ctx, &share_i, share)) {
+        return 0;
+    }
+
+    if (!secp256k1_frost_evaluate_vss(ctx, &tmpj, threshold, id, vss_commitment)) {
+        return 0;
+    }
+
+    secp256k1_scalar_negate(&share_neg, &share_i);
```
Christewart: I'm following the[ RFC C.2](https://www.rfc-editor.org/rfc/rfc9591.html#appendix-C.2) and I don't understand where `L282-L286` are coming from. Can you help me understand what this is doing?

apoelstra:
a4ceda50240a4531e11a83d7a2924e8be2ca02d3 needs rebase

apoelstra:
69e9d289ff1b670bbdf080427d306c02a89a4425 needs rebase

apoelstra:
f0d36bbd945afd7f3e642540aef4542142464523 needs rebase

src/modules/frost/session_impl.h
BEULAHEVANJALIN: f0d36bbd945afd7f3e642540aef4542142464523

PR #282, opened 2024-01-11
Avoid using out-of-bounds field elements (in impossible cases)

roconnor-blockstream:
secp256k1_fe_set_b32_limit says that when it returns 0, one is not allowed to use the resulting output value.

This refactors the code so that the existing value of t is maintained in cases where sha256 would output an out-of-bounds hash value.

Note: This situation is cryptographically impossible to occur.

### DIFF
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -187,6 +187,7 @@ static int secp256k1_generator_generate_internal(const secp256k1_context* ctx, s
     static const unsigned char prefix1[17] = "1st generation: ";
     static const unsigned char prefix2[17] = "2nd generation: ";
     secp256k1_fe t = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 4);
+    secp256k1_fe set = SECP256K1_FE_CONST(0, 0, 0, 0, 0, 0, 0, 0);
     secp256k1_ge add;
     secp256k1_gej accum;
     int overflow;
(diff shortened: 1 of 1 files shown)


real-or-random (commented):
Hm, yeah, good catch! This is similar to https://github.com/bitcoin-core/secp256k1/pull/1316 .

I wonder if there's a better fix in this case. Looking at the implementation of `_limit`, the field element is actually a valid one. Just in the tests, we mark it as invalid by setting the magnitude to `-1`, so that `_fe_verify` will catch any read accesses to it.

I guess we could declassify `ret` and return early when it's 0.

roconnor-blockstream:
Ah this bug was introduced in https://github.com/BlockstreamResearch/secp256k1-zkp/pull/256 when `secp256k1_fe_set_b32` was split into `_mod` and `_limit` variants.  The `_limit` version isn't quite a drop in replacement due to the fact the post condition says we cannot use the value if failure is returned.

Anyhow, I think we should seriously entertain just using the `_mod` version and not failing.

Honestly I don't understand why people insist on failing when hashes return values larger than the secp256k1 field size, rather than just wrapping.

real-or-random:
> Yes, it is infinitesimally slower to use `_mod`, but it pales in comparison the the expense of everything else here.

It's the other around! `_limit` first calls `_mod`, so `_mod`  faster...

real-or-random:
> Anyhow, I think we should seriously entertain just using the `_mod` version and not failing.

Agreed, that's the best fix.

> 
> Honestly I don't understand why people insist on failing when hashes return values larger than the secp256k1 field size, rather than just wrapping.

For the record, I fully agree. We have debated negligible cases a lot in the past, but I've personally settled on this conclusion: Since these cases won't occur in practice, we should not discuss what happens if we hit these cases. We instead just should do whatever is convenient for us.

And in this case, this is for sure taking a mod. This avoids a branch (and an annoying branch because we cannot test it, etc...), it makes sure static analysis tools won't complain about invalid uses / out-of-bound fields, and it's faster...

roconnor-blockstream:
Okay.  I will build and alternative PR that uses `_mod`.

roconnor-blockstream:
I wrote up a version that uses `_mod` and the testsuite (correctly) fails.  There is actually a problem with that proposal.

https://github.com/BlockstreamResearch/secp256k1-zkp/blob/03aecafe4c45f51736ce05b339d2e8bcc2e5da55/src/modules/generator/main_impl.h#L200

`shallue_van_de_woestijne` calls `secp256k1_fe_is_odd(t)` which means that `shallue_van_de_woestijne` has a precondition that `t` is normalized.  The `_limit` version guarantees that the the result is normalized but the `_mod` version does not.

Issue #279, opened 2023-11-27
shallue_van_de_woestijne returns an off-curve point when t is 0.


roconnor-blockstream:
My suggestion is to either

A) Rewrite `shallue_van_de_woestijne`.  The natural result for t equal to 0 is the point with x-coordinate `d`.  In particular this is what the cited paper suggests that the function do.

or

B) Have `secp256k1_generator_generate_internal` return 0 if it calls `shallue_van_de_woestijne` with 0.

My preference would be for (A).  The existing elements codebase already just crashes if `secp256k1_generator_generate_internal` returns 0. Therefore doing a cryptographically impossible hard-fork is not materially worse.

Doing (A) totally addresses the issue while, doing (B) requires still doing point addition with off-curve points in violation of that code's preconditions.  While all the functions being called in `secp256k1_generator_generate_internal` should be safe to run on off-curve points, some code, (such as some scalar multiplication operations) really does requires on-curve points to avoid undefined behavior.

jonasnick:
Fixed in merged PR #286.

PR #283, opened 2024-01-12
Add failure case to secp256k1_generator_generate_interal

roconnor-blockstream:
The implementation of shallue_van_de_woestijne currently returns an off-curve point when given a value of t.

While this is cryptographically impossible to occur, none-the-less we should fail in this case rather than simply return a garbage value.

### DIFF
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -90,6 +90,7 @@ int secp256k1_generator_serialize(const secp256k1_context* ctx, unsigned char *o
     return 1;
 }
 
+/* ACHTUNG: Calling with t = 0, will return an off-curve point. */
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
This fixes issue #279 using method (B) by having `secp2561_generator_generate_internal` return 0 in case one of the hashes results in the all 0 value (this is cryptographically impossible).

No however this "fix" still results in computations proceeding with off-curve points, which is somewhere between "strongly discouraged" and "a violation of the preconditions of the secp256k1 operations".

For an alternative solution see #284.

jonasnick (commented):
Concept ACK

I wrote [a commit](https://github.com/jonasnick/secp256k1-zkp/commits/2024-01-generator_generate-jn/) that adds a test for t = 0, and that confirms the off-curve point on master. Feel free to cherry-pick.

jonasnick:
Merged alternative fix #286.

PR #284, opened 2024-01-12
shallue_van_de_woestijne support for t = 0

roconnor-blockstream:
The current implementation returns an off-curve point for the input t=0. This is due to performing a "0/0" computation.

Here we change the code to explicitly return the on-curve point (d, sqrt(8)), which is the point that the paper Indifferentiable Hashing to Barreto–Naehrig Curves suggests returning in this case.

Note: At the moment it is cryptographically impossible for the input t to be 0.

### DIFF
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -90,6 +90,7 @@ int secp256k1_generator_serialize(const secp256k1_context* ctx, unsigned char *o
     return 1;
 }
 
+/* t must be normalized. */
 static void shallue_van_de_woestijne(secp256k1_ge* ge, const secp256k1_fe* t) {
     /* Implements the algorithm from:
      *    Indifferentiable Hashing to Barreto-Naehrig Curves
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
This fixes issue #279 using method (A) by having changing the behavior of shallue_van_de_woestijne at the input t=0.

The downside of this fix is that it alters the behavior of this function.  This should be reasonably safe to do even though it amounts to a hard-fork in Elements consensus code, because it is cryptographically impossible for the t=0 case to arise in practice.  This can be seen because this is a `static` function, it can only be called in this file, and we can see that t is only ever set to the result of a sha256 digest (or perhaps t = 4).

For an alternative solution see #283.

This change is the minimally invasive change that I could think of to fix this function.  However, I think this function could, and perhaps should, be rewritten to be streamlined.  I have [a version written in Haskell](https://github.com/BlockstreamResearch/simplicity/blob/495c9c0f611a189722ebc1644a777a7f6c008cb4/Haskell/Core/Simplicity/LibSecp256k1/Spec.hs#L376-L401) that could be translated to C.  Please advise me if you want me to create such a PR and see how it looks in comparison to this (and #277).

apoelstra:
My thoughts on these PRs:

1. I don't think it's a "hardfork" to have forking behavior on a cryptographically unreachable path.
2. But since this will stop being cryptographically unreachable post-Simplicity we want to fix this issue now.
3. And I think "returning an on-curve point" is the most elegant solution, vs having an error path to deal with or vs having an off-curve point to deal with.

jonasnick:
Merged alternative fix #286.

PR #288, opened 2024-01-25
cmake: Add support for -zkp modules

real-or-random:
Supersedes #275.


real-or-random:
@lightyear15 Can you test if this does the right thing for you?

lightyear15:
@real-or-random
I did a quick test and it seems to be working
Thanks

jonasnick (approved):
utACK 4228fd1124281af57c9872f1de3f4eaac73973f4

PR #289, opened 2024-01-29
Avoid using out-of-bounds field elements (in impossible cases)

real-or-random:
secp256k1_fe_set_b32_limit says that when it returns 0, one is not allowed to use the resulting output value.

This refactors the code so that 0 is returned early (indicating failure) in cases where sha256 would output an out-of-bounds hash value. This makes secp256k1_generator_generate_internal variable-time in its "t" argument, but this not a problem because this value is public in applications.

Note: This situation is cryptographically impossible to occur.

Alternative to #282. cc @roconnor-blockstream What do you think?

### DIFF
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -222,7 +222,9 @@ static int secp256k1_generator_generate_internal(const secp256k1_context* ctx, s
     secp256k1_sha256_write(&sha256, prefix1, 16);
     secp256k1_sha256_write(&sha256, key32, 32);
     secp256k1_sha256_finalize(&sha256, b32);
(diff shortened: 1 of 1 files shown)


roconnor-blockstream:
Seems okay to me if you folks are okay with the variable time in these (impossible) cases.

real-or-random:
> Seems okay to me if you folks are okay with the variable time in these (impossible) cases.

AFAIU, we should be okay even if the cases were more likely. The derivation of generators should only be a public computation (except maybe blinding), but it will be nice if @apoelstra can confirm that I haven't overlooked anything in this argument.

roconnor-blockstream:
I agree with you, except that the code seems to have been clearly and deliberately written to be constant time for some reason.

PR #290, opened 2024-02-05
configure: Clean ups

real-or-random:


jonasnick:
What is the canonical order of modules?

real-or-random:
The one used for the echos, i.e., first the upstream modules (in their order) and then our modules in the ordered they appeared. And that order should be used in the other places, just sometimes reversed...

Let me add a comment. (I just don't want to add a comment everywhere we list modules. This would again need to be done upstream, and I don't want to make that dance yet another time, at least not for an issue of this kind. :D)

real-or-random:
Updated.

real-or-random:
Changed it again, so that the main idea can at least be ported to upstream easily. Really ready for review now. ^^

jonasnick (commented):
ACK 860e3bb29461c05b495c6f027b6c5a62532738b5

PR #292, opened 2024-05-07
doc: fix sage code for deriving alternative generator H

theStack:
The line calculating H (in particular, the expression `G.decode('hex')`) fails with the following error message on Sage 9.5:

```
AttributeError: 'str' object has no attribute 'decode'
```

Fix that by converting the hex-string to bytes using `bytes.fromhex`.

(Noticed while reviewing https://github.com/bitcoin/bitcoin/pull/30048 which picks this code snippet comment up.)


josibake (approved):
ACK https://github.com/BlockstreamResearch/secp256k1-zkp/pull/292/commits/7040a2024795a7e3758c7ab604ab440652c9772f

Verified by running this using the binder instance for Sage 10.3: https://mybinder.org/v2/gh/sagemath/sage-binder-env/master. 

Worth mentioning this is a python2 to python3 change (decoded was removed in python3 iirc).

real-or-random (approved):
utACK 7040a2024795a7e3758c7ab604ab440652c9772f

PR #294, opened 2024-05-10
generator: massively speed up serialization

### LINKED ISSUE #293: `secp256k1_pedersen_commitment_serialize` decompresses the point then recompresses it
Ok, yeah, the issue is in libsecp256k1-zkp, not here. For some bizarre reason `secp256k1_pedersen_commitment_serialize` decompresses the point then recompresses it, when all it needs to do is a memcmp (I think). So this function is wasting a lot of time for no reason.

_Originally posted by @apoelstra in https://github.com/ElementsProject/rust-elements/issues/202#issuecomment-2104503688_

apoelstra:
`secp256k1_pedersen_commit_serialize` would call `_load` (which does a sqrt to fully decompress the key, then a conditional negation based on the flag), then check the Jacobian symbol of the resulting y-coordinate, then re-serialize based on this.

Instead, don't do any of this stuff. Copy the flag directly out of the internal representation and copy the x-coordinate directly out of the internal representation.

Checked that none of the other _serialize methods in the modules do this.

Fixes #293

### DIFF
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -296,17 +296,11 @@ int secp256k1_pedersen_commitment_parse(const secp256k1_context* ctx, secp256k1_
 }
 
 int secp256k1_pedersen_commitment_serialize(const secp256k1_context* ctx, unsigned char *output, const secp256k1_pedersen_commitment* commit) {
-    secp256k1_ge ge;
-
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(output != NULL);
     ARG_CHECK(commit != NULL);
 
-    secp256k1_pedersen_commitment_load(&ge, commit);
-
-    output[0] = 9 ^ secp256k1_fe_is_square_var(&ge.y);
-    secp256k1_fe_normalize_var(&ge.x);
-    secp256k1_fe_get_b32(&output[1], &ge.x);
+    memcpy(output, commit->data, 33);
     return 1;
 }
 
--- a/src/modules/generator/main_impl.h
+++ b/src/modules/generator/main_impl.h
@@ -276,7 +276,6 @@ static void secp256k1_pedersen_commitment_save(secp256k1_pedersen_commitment* co
 
 int secp256k1_pedersen_commitment_parse(const secp256k1_context* ctx, secp256k1_pedersen_commitment* commit, const unsigned char *input) {
     secp256k1_fe x;
-    secp256k1_ge ge;
 
     VERIFY_CHECK(ctx != NULL);
     ARG_CHECK(commit != NULL);
@@ -285,13 +284,11 @@ int secp256k1_pedersen_commitment_parse(const secp256k1_context* ctx, secp256k1_
 
     if ((input[0] & 0xFE) != 8 ||
         !secp256k1_fe_set_b32_limit(&x, &input[1]) ||
-        !secp256k1_ge_set_xquad(&ge, &x)) {
+        !secp256k1_ge_x_on_curve_var(&x)) {
         return 0;
     }
-    if (input[0] & 1) {
-        secp256k1_ge_neg(&ge, &ge);
-    }
-    secp256k1_pedersen_commitment_save(commit, &ge);
+
+    memcpy(commit->data, input, 33);
     return 1;
 }
 
--- a/src/modules/generator/tests_impl.h
+++ b/src/modules/generator/tests_impl.h
@@ -264,7 +264,13 @@ static void test_pedersen(void) {
     }
     CHECK(secp256k1_pedersen_blind_sum(CTX, &blinds[(total - 1) * 32], bptr, total - 1, inputs));
     for (i = 0; i < total; i++) {
+        unsigned char result[33];
+        secp256k1_pedersen_commitment parse;
+
         CHECK(secp256k1_pedersen_commit(CTX, &commits[i], &blinds[i * 32], values[i], secp256k1_generator_h));
+        CHECK(secp256k1_pedersen_commitment_serialize(CTX, result, &commits[i]));
+        CHECK(secp256k1_pedersen_commitment_parse(CTX, &parse, result));
+        CHECK(secp256k1_memcmp_var(&commits[i], &parse, 33) == 0);
     }
     CHECK(secp256k1_pedersen_verify_tally(CTX, cptr, inputs, &cptr[inputs], outputs));
     CHECK(secp256k1_pedersen_verify_tally(CTX, &cptr[inputs], outputs, cptr, inputs));


src/modules/generator/main_impl.h
```diff
@@ -296,17 +296,12 @@ int secp256k1_pedersen_commitment_parse(const secp256k1_context* ctx, secp256k1_
...
     ARG_CHECK(output != NULL);
     ARG_CHECK(commit != NULL);
 
-    secp256k1_pedersen_commitment_load(&ge, commit);
-
-    output[0] = 9 ^ secp256k1_fe_is_square_var(&ge.y);
-    secp256k1_fe_normalize_var(&ge.x);
-    secp256k1_fe_get_b32(&output[1], &ge.x);
+    output[0] = 8 ^ (commit->data[0] & 1);
+    memcpy(&output[1], &commit->data[1], 32);
```
real-or-random, line 304: Couldn't you even memcpy the entire 33 bytes?

real-or-random (commented):
Concept ACK

src/modules/generator/main_impl.h
delta1: > Couldn't you even memcpy the entire 33 bytes? 

I tried this, it does pass the tests :) 

I see `8 ^ (commit->data[0] & 1)` alternates between `1000` and `1001` binary for even/odd values of `data[0]`. 

Out of interest as cryptography noob, what is the significance of this?

real-or-random: It's just a flag byte. `0x02`, `0x03` (and a few others) are already used for public keys, so whoever wrote that code took `0x08` and `0x09` to distinguish commitments from public keys.

delta1: @real-or-random thanks!

apoelstra:
Changed to memcpy the entire 33 bytes. Also added a commit which does the same thing for parsing.

Previously, parsing decoded the whole point, then conditionally negated it, then extracted the jacobi symbol of the y-coordinate (which is entirely determined by the `ge_set_xquad` followed by conditional negation!), then normalized and re-serialized the x-coordinate (which was parsed directly using `fe32_set_b32_limit` so it was already normalized).

But all we need to do is (a) check that the first byte is 8 or 9; (b) check that the remainder is a valid x-coordinate, then (c) memcpy.

Step (b) is still pretty slow, compared to serialization which is just a memcpy, but this is a big improvement.

apoelstra:
CI failure appears to be some unrelated break in our macos image.

jonasnick (commented):
Concept ACK

This PR looks good.

As a sidenote (not this PR) the implementation of `pedersen_commitment_load` and `save` is quite different from `pubkey_load` and `save`. In contrast to pubkey loading, pedersen commitment loading is costly, which results in unnecessary sqrts when loading commitments in `verify_tally`.

src/modules/generator/tests_impl.h
```diff
@@ -264,7 +264,13 @@ static void test_pedersen(void) {
     }
     CHECK(secp256k1_pedersen_blind_sum(CTX, &blinds[(total - 1) * 32], bptr, total - 1, inputs));
     for (i = 0; i < total; i++) {
+        unsigned char result[33];
+        secp256k1_pedersen_commitment parse;
+
         CHECK(secp256k1_pedersen_commit(CTX, &commits[i], &blinds[i * 32], values[i], secp256k1_generator_h));
+        CHECK(secp256k1_pedersen_commitment_serialize(CTX, result, &commits[i]));
+        CHECK(secp256k1_pedersen_commitment_parse(CTX, &parse, result));
+        CHECK(secp256k1_memcmp_var(&commits[i], result, 33) == 0);
```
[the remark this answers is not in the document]
apoelstra: Oh, yeah, it should be checking `commits[i]` against `parse` rather than against `result` :).

This code works but only because of a bunch of C weirdness. Will fix.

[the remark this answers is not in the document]
apoelstra: Fixed.

src/modules/generator/main_impl.h
```diff
@@ -288,10 +288,8 @@ int secp256k1_pedersen_commitment_parse(const secp256k1_context* ctx, secp256k1_
         !secp256k1_ge_set_xquad(&ge, &x)) {
```
real-or-random, line 288:
```suggestion
        !secp256k1_ge_x_on_curve_var(&x)) {
```
and we can drop `ge` entirely.

real-or-random (commented):
We could also consider overwriting the output with zeros (or anything which is initialized but not a valid commitment) when parsing fails. We do this in upstream for other parsing functions. 

Not sure, it may be a bit arbitrary to do it here now. This should maybe go to a proper PR that does it consistently in all -zkp modules.

src/modules/generator/main_impl.h
apoelstra: Force-pushed to do this.

real-or-random (approved):
utACK 6361266013ad14428c89334013c74f8dec6f8e9d

I think I know how to fix CI, I'll open a PR later today or tomorrow

real-or-random:
The CI issue should be resolved once the GitHub Actions image is updated to have brew 4.3.1, which has the fix (https://github.com/Homebrew/brew/pull/17336). The images are updated every week, so the issue should just disappear in a few days.

jonasnick:
I am surprised that the last 31 bytes of the pedersen commitment struct are uninitialized preventing something like
```
memcpy(&com1, &com2, sizeof(secp256k1_pedersen_commitment));
```
However, that is already the case without this PR.

jonasnick (approved):
ACK 6361266013ad14428c89334013c74f8dec6f8e9d

real-or-random:
> I am surprised that the last 31 bytes of the pedersen commitment struct are uninitialized preventing something like
> 
> ```
> memcpy(&com1, &com2, sizeof(secp256k1_pedersen_commitment));
> ```

Yes, that's a bit strange. That means we can always switch to an uncompressed internal representation. I haven't thought about the performance implications, but uncompressed should usually be faster (unless you only deserialize and serialize again without doing actual computations, but why should you do this...)

This dates back to https://github.com/BlockstreamResearch/secp256k1-zkp/commit/edb879f57893c6d815db16d5793ad9143b4bd6b9 (see also the parent commit https://github.com/BlockstreamResearch/secp256k1-zkp/commit/fca4c3b62f386ece6a09de6373b1750e91f04ca9 which does the same for `secp256k1_generator`), but I can't find the corresponding PR. @apoelstra Do you remember why you changed this?

apoelstra:
> Yes, that's a bit strange. That means we can always switch to an uncompressed internal representation. I haven't thought about the performance implications, but uncompressed should usually be faster (unless you only deserialize and serialize again without doing actual computations, but why should you do this...)
> 

This is a somewhat common thing to do with Elements transactions, where you parse the whole transaction (including decompressing the points, because this is how you obtain valid pubkey-type objects) and then only manipulate non-crypto parts before re-serializing.

With normal pubkeys serialization is (almost) just as fast with a compressed vs uncompressed representation because you just need to check the lsb of the y coordinate. But with secp-zkp keys, recompressing an uncompressed key requires recomputing the jacobi symbol :/.

> This dates back to [edb879f](https://github.com/BlockstreamResearch/secp256k1-zkp/commit/edb879f57893c6d815db16d5793ad9143b4bd6b9) (see also the parent commit [fca4c3b](https://github.com/BlockstreamResearch/secp256k1-zkp/commit/fca4c3b62f386ece6a09de6373b1750e91f04ca9) which does the same for `secp256k1_generator`), but I can't find the corresponding PR. @apoelstra Do you remember why you changed this?

Afraid not. And I also cannot find the corresponding PR.

real-or-random:
I see, so you're saying it makes sense to keep the internal representation compressed for performance reasons?

If we keep this, then we should at least write all 64 bytes to avoid passing uninitialized memory to the user. Or simply switch back to an opaque type that has just 33 bytes. The latter is cleaner and simpler IMO.

apoelstra:
Honestly I'm tempted to switch to *65* bytes before switching to 33. So we'd have the uncompressed representation which can be efficiently manipulated by the actual crypto functions, and also an xquad flag bit (probably also useful for crypto actually) so we can quickly serialize.

real-or-random:
Ok, sure, that's also possible...

The issue with switching either to 32 or 65 is that it's a breaks ABI compatibility for any code copying or moving structs. We say this:
https://github.com/BlockstreamResearch/secp256k1-zkp/blob/168377204d5a826df0123093b592147c66cf4218/include/secp256k1_generator.h#L14-L16

Strictly speaking, all of this is in an experimental module, but yeah, not sure if we want to risk breaking things.

apoelstra:
Agreed. I think for now we should leave it alone. Maybe if/when we implement BP++ and switch to it in Elements, we can consider also breaking the generator API.

PR #295, opened 2024-06-19
rangeproof: add unit test for malleating single-value proofs

apoelstra:
I was a bit confused reading `secp256k1_rangeproof_getheader_impl` because in the case of single-value proofs (`has_nz_range == 0`) some bits of the header are unconstrained. At first I thought this was a malleability vector. And I think I've had this same confusion in the past.

But in fact it is not a malleability vector because the whole header gets hashed into the proof.

Add a unit test to confirm this to reduce future confusion.

### DIFF
--- a/src/modules/rangeproof/tests_impl.h
+++ b/src/modules/rangeproof/tests_impl.h
@@ -422,6 +422,7 @@ static void test_single_value_proof(uint64_t val) {
 
     uint64_t val_out = 0;
     size_t m_len_out = 0;
+    size_t i;
 
     secp256k1_testrand256(blind);
     secp256k1_testrand256(nonce);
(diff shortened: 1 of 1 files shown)


src/modules/rangeproof/tests_impl.h
```diff
@@ -463,6 +464,30 @@ static void test_single_value_proof(uint64_t val) {
         CHECK(plen == 73);
     }
 
+    /* Test if trailing bytes are rejected. */
+    proof[plen] = 0;
+    CHECK(secp256k1_rangeproof_verify(
+        CTX,
+        &min_val_out, &max_val_out,
+        &commit,
+        proof, plen + 1,
```
real-or-random, line 473: `proof` should probably be fully initialized to avoid a read of uninitialized data here. (I think `secp256k1_rangeproof_verify` will not read the last byte, but the test shouldn't rely on that.)

apoelstra: I do initialize the last byte in the line above this check.

real-or-random: oh, nevermind

real-or-random (approved):
utACK 3a1c39625e7f9c05234df71c35bee3dae5b785e6

PR #296, opened 2024-08-28
Musig: Minor comment corrections

siv2r:
Changes:
- PartialSigVerify mistakenly refers to `pubnonce` as `aggnonce`.
    - _Edit:_ Just noticed that this correction was already pointed out here: https://github.com/bitcoin-core/secp256k1/pull/1479#discussion_r1731398318
- KeyAgg comments inconsistently use both `x` and `pk` to refer to the public key argument. Updated to use only `pk`.

### DIFF
--- a/src/modules/musig/keyagg_impl.h
+++ b/src/modules/musig/keyagg_impl.h
@@ -129,7 +129,7 @@ static void secp256k1_musig_keyaggcoef_sha256(secp256k1_sha256 *sha) {
 }
 
 /* Compute KeyAgg coefficient which is constant 1 for the second pubkey and
- * otherwise tagged_hash(pk_hash, x) where pk_hash is the hash of public keys.
+ * otherwise tagged_hash(pk_hash, pk) where pk_hash is the hash of public keys.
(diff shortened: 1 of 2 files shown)


siv2r:
https://github.com/BlockstreamResearch/secp256k1-zkp/blob/6152622613fdf1c5af6f31f74c427c4e9ee120ce/src/modules/musig/session_impl.h#L579-L587

I'm not sure why `pk.y` is being normalized here. Shouldn't `cache_i.pk.y` be normalized instead? That's the one passed to the `fe_is_odd` function, which requires the field element to be normalized.

Just a side note: all tests still pass even after removing this function call (line 579).

jonasnick:
Thanks, I think all the corrections were either addressed or pointed out in the PR to upstream libsecp256k1.

jonasnick:
I think it's best if we merge the PR in upstream libsecp first and then merge the changes to this repo.

real-or-random:
> I think it's best if we merge the PR in upstream libsecp first and then merge the changes to this repo.

Yes, this can be closed then. Updating to the upstream musig module is tracked in #301.

PR #300, opened 2024-10-26
fix and verify `compressed` argument in _eckey_pubkey_serialize calls

theStack:
In several calls of the internal function `secp256k1_eckey_pubkey_serialize`, the public API flag `SECP256K1_EC_COMPRESSED` is passed, which is meant to be only used for the public function `secp256k1_ec_pubkey_serialize`. It works as intended in all of those cases (it wouldn't for `SECP256K1_EC_UNCOMPRESSED` though), but it's still kind of a type mismatch that can't be detected by the compiler.  To avoid cases like this in the future, a VERIFY_CHECK is added that the `compressed` parameter needs to be either 0 or 1.

### DIFF
--- a/src/eckey_impl.h
+++ b/src/eckey_impl.h
@@ -9,6 +9,7 @@
 
 #include "eckey.h"
 
+#include "util.h"
 #include "scalar.h"
 #include "field.h"
 #include "group.h"
@@ -35,6 +36,8 @@ static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char 
 }
 
 static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size, int compressed) {
+    VERIFY_CHECK(compressed == 0 || compressed == 1);
+
     if (secp256k1_ge_is_infinity(elem)) {
         return 0;
     }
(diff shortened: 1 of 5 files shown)


apoelstra:
LGTM

src/eckey_impl.h
```diff
@@ -35,6 +36,8 @@ static int secp256k1_eckey_pubkey_parse(secp256k1_ge *elem, const unsigned char
 }
 
 static int secp256k1_eckey_pubkey_serialize(secp256k1_ge *elem, unsigned char *pub, size_t *size, int compressed) {
+    VERIFY_CHECK(compressed == 0 || compressed == 1);
```
real-or-random, line 39: Not sure if we want to touch this upstream file. It's also a nice goal to keep the diff to upstream minimal. Ideally, the diff to upstream would just be added modules and changes to build system, README, etc... But yeah, we've modified upstream files in the past. If we think that this is a reasonable change, we could PR it to upstream, of course.

Same is true for the modification to secp256k1.c

src/modules/musig/session_impl.h
real-or-random, line 1: We should just rebase on upstream's musig module now.

real-or-random (commented):
Concept ACK

theStack:
Closing this, as a corresponding PR was opened and merged in upstream (see https://github.com/bitcoin-core/secp256k1/pull/1642), i.e. it will be available here once a rebase on v0.7.0 (or later) is done.

PR #302, opened 2025-09-08
extrakeys: fix pubkey_sort_cmp test

jonasnick:
Instead of providing CTX directly, pass a cmp_data object containing CTX. Otherwise, memory sanitizer fails with "use-of-uninitialized-value".

### DIFF
--- a/src/modules/extrakeys/tests_impl.h
+++ b/src/modules/extrakeys/tests_impl.h
@@ -605,7 +605,9 @@ static void test_sort(void) {
         }
         secp256k1_pubkey_sort(CTX, pk_ptr, 5);
         for (j = 1; j < 5; j++) {
-            CHECK(secp256k1_pubkey_sort_cmp(&pk_ptr[j - 1], &pk_ptr[j], CTX) <= 0);
+            secp256k1_pubkey_sort_cmp_data cmp_data;
+            cmp_data.ctx = CTX;
+            CHECK(secp256k1_pubkey_sort_cmp(&pk_ptr[j - 1], &pk_ptr[j], &cmp_data) <= 0);
         }
     }
 }


src/modules/extrakeys/tests_impl.h
```diff
@@ -605,7 +605,9 @@ static void test_sort(void) {
         }
         secp256k1_pubkey_sort(CTX, pk_ptr, 5);
         for (j = 1; j < 5; j++) {
-            CHECK(secp256k1_pubkey_sort_cmp(&pk_ptr[j - 1], &pk_ptr[j], CTX) <= 0);
+            secp256k1_pubkey_sort_cmp_data cmp_data;
```
BEULAHEVANJALIN, line 608: I confirmed that declaring the `cmp_data` variable at the beginning of the function does not cause any issues. This means we do not need to create a new instance for each iteration.

I reviewed the `secp256k1_pubkey_sort_cmp` function, and it appears that the `ctx` variable is not utilized within the `secp256k1_pubkey_load` function. My understanding of this usage is limited, so please correct me if I am mistaken.

apoelstra: It's correct that `ctx` is never used (which is why the old, incorrect code could pass the tests).

As for scoping, IMO we should keep `cmp_data` as tightly scoped as we can, because it makes the code easier to read.

BEULAHEVANJALIN: Got it, thanks!

apoelstra (approved):
ACK 83d0fa25a826d716f809e00ef0a67385b776aec9; successfully ran local tests

BEULAHEVANJALIN:
Follow-up opened in #303 to fix the remaining CI failures in. This should stabilize ci and unblock FROST (#278)

PR #357, opened 2026-04-03
Upstream PRs 1824, 1777, 1834, 1837, 1839, 1841

DarkWindman:
[https://github.com/bitcoin-core/secp256k1/pull/1824]: util: introduce and use ARRAY_SIZE macro
[https://github.com/bitcoin-core/secp256k1/pull/1777]: Make SHA256 compression runtime pluggable
[https://github.com/bitcoin-core/secp256k1/pull/1834]: ecmult: Document and test ng=NULL in ecmult
[https://github.com/bitcoin-core/secp256k1/pull/1837]: tests: Fix function pointer initialization C89 error in ellswift tests
[https://github.com/bitcoin-core/secp256k1/pull/1839]: ecdsa: VERIFY_CHECK result of _fe_set_b32_limit
[https://github.com/bitcoin-core/secp256k1/pull/1841]: gha: Bump deprecated GHA workflow dependencies

Tips:

    Use git show --remerge-diff <pr-branch> to show the conflict resolution in the merge commit.
    Use git read-tree --reset -u <pr-branch> to replay these resolutions during the conflict resolution stage when recreating the PR branch locally.
    Be aware that this may discard your index as well as the uncommitted changes and untracked files in your worktree.


real-or-random:
@mllwchrry Do you want to review this?

src/modules/ecdsa_s2c/tests_impl.h
```diff
@@ -171,7 +172,7 @@ static void test_ecdsa_s2c_fixed_vectors(void) {
     };
     size_t i;
 
-    for (i = 0; i < sizeof(ecdsa_s2c_tests) / sizeof(ecdsa_s2c_tests[0]); i++) {
+    for (i = 0; i < ARRAY_SIZE(ecdsa_s2c_tests) ; i++) {
```
mllwchrry: Trailing space before a semicolon

src/modules/rangeproof/main_impl.h
```diff
@@ -46,13 +47,14 @@ int secp256k1_rangeproof_rewind(const secp256k1_context* ctx,
...
-    return secp256k1_rangeproof_verify_impl(&ctx->ecmult_gen_ctx,
+    return secp256k1_rangeproof_verify_impl(hash_ctx, &ctx->ecmult_gen_ctx,
      blind_out, value_out, message_out, outlen, nonce, min_value, max_value, &commitp, proof, plen, extra_commit, extra_commit_len, &genp);
 }
 
 int secp256k1_rangeproof_verify(const secp256k1_context* ctx, uint64_t *min_value, uint64_t *max_value,
  const secp256k1_pedersen_commitment *commit, const unsigned char *proof, size_t plen, const unsigned char *extra_commit, size_t extra_commit_len, const secp256k1_generator* gen) {
-    secp256k1_ge commitp;
+    const secp256k1_hash_ctx *hash_ctx = secp256k1_get_hash_context(ctx);
+   secp256k1_ge commitp;
```
mllwchrry: Invalid indentation

src/modules/rangeproof/main_impl.h
```diff
@@ -63,14 +65,15 @@ int secp256k1_rangeproof_verify(const secp256k1_context* ctx, uint64_t *min_valu
...
-    return secp256k1_rangeproof_verify_impl(NULL,
+    return secp256k1_rangeproof_verify_impl(hash_ctx, NULL,
      NULL, NULL, NULL, NULL, NULL, min_value, max_value, &commitp, proof, plen, extra_commit, extra_commit_len, &genp);
 }
 
 int secp256k1_rangeproof_sign(const secp256k1_context* ctx, unsigned char *proof, size_t *plen, uint64_t min_value,
  const secp256k1_pedersen_commitment *commit, const unsigned char *blind, const unsigned char *nonce, int exp, int min_bits, uint64_t value,
  const unsigned char *message, size_t msg_len, const unsigned char *extra_commit, size_t extra_commit_len, const secp256k1_generator* gen){
-    secp256k1_ge commitp;
+   const secp256k1_hash_ctx *hash_ctx = secp256k1_get_hash_context(ctx);
```
mllwchrry: Invalid indentation in lines 75 and 76

src/eccommit_impl.h
```diff
@@ -25,42 +25,42 @@ static int secp256k1_ec_commit_pubkey_serialize_const(secp256k1_ge *pubp, unsign
 }
 
 /* Compute an ec commitment tweak as hash(pubp, data). */
-static int secp256k1_ec_commit_tweak(unsigned char *tweak32, secp256k1_ge* pubp, secp256k1_sha256* sha, const unsigned char *data, size_t data_size)
+static int secp256k1_ec_commit_tweak(const secp256k1_hash_ctx *hash_ctx,unsigned char *tweak32, secp256k1_ge* pubp, secp256k1_sha256* sha, const unsigned char *data, size_t data_size)
```
mllwchrry: Missing space after comma

src/modules/bppp/main_impl.h
```diff
@@ -37,11 +38,11 @@ secp256k1_bppp_generators *secp256k1_bppp_generators_create(const secp256k1_cont
     secp256k1_fe_get_b32(&seed[0], &secp256k1_ge_const_g.x);
     secp256k1_fe_get_b32(&seed[32], &secp256k1_ge_const_g.y);
 
-    secp256k1_rfc6979_hmac_sha256_initialize(&rng, seed, 64);
+    secp256k1_rfc6979_hmac_sha256_initialize(hash_ctx,&rng, seed, 64);
```
mllwchrry: Missing space after comma

mllwchrry (commented):
The PR is good overall. A few style nits and one missed place for #1834. 
Claude also caught a hard-to-catch bug in the #1777 port: `secp256k1_ecdsa_adaptor_encrypt` falls back to a static-context wrapper for nonce generation instead of dispatching to the `_impl` function with the actual context. This way a custom SHA256 compression function set on the context is silently ignored for adaptor signature nonce generation. We discussed it internally, and the fix will be included in this PR.

mllwchrry:
> Claude also caught a hard-to-catch bug in the #1777 port: `secp256k1_ecdsa_adaptor_encrypt` falls back to a static-context wrapper for nonce generation instead of dispatching to the `_impl` function with the actual context. This way a custom SHA256 compression function set on the context is silently ignored for adaptor signature nonce generation. We discussed it internally, and the fix will be included in this PR.

After the investigation, it turned out that the fix is more involved than expected. The nonce generation for the dleq proof inside `secp256k1_dleq_prove` also goes through the same static context wrapper, so both call sites need to be addressed.
Our fix plan is to restructure where the nonce is computed: moving it out of `dleq_impl.h` into `main_impl.h` and passing a pre-computed scalar to `secp256k1_dleq_prove`. This is required since we decided to apply the approach used for schnorrsig from #1777. The idea is to modify the internal `nonce_function_ecdsa_adaptor_impl` function to take hash context and call it if `noncefp` is NULL, keeping the public API wrapper unchanged and uncalled. The complication is that `dleq_impl.h` is included at the top of `main_impl.h`, so `nonce_function_ecdsa_adaptor_impl` isn't defined yet when `dleq_impl.h` is processed. That's why we decided to move nonce computation to `main_impl.h`. If you have any other thoughts on this, please share. 

@real-or-random Would it make sense to merge this PR as-is once style fixes are applied and address the bug in another PR?

real-or-random:
> 
> @real-or-random Would it make sense to merge this PR as-is once style fixes are applied and address the bug in another PR?

My understanding is that the bug is just that the externally-provided SHA256 callback is not used. While this classifies as a  bug, it's just about performance and not about correctness. If this understanding is right, then yes, I think it's okay to fix this in a follow-up PR.

mllwchrry:
ACK baac08d

real-or-random (approved):
ACK baac08d207003151d095423487d3954503a52aeb

src/modules/schnorrsig/tests_impl.h
```diff
@@ -849,6 +852,29 @@ static void test_schnorrsig_sign_internal(void) {
     CHECK(secp256k1_memcmp_var(sig, sig2, sizeof(sig)) == 0);
 }
 
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_schnorrsig)
+static void test_schnorrsig_ctx_sha256(void) {
+    /* Check ctx-provided SHA256 compression override takes effect */
```
real-or-random, line 857: Do you think it would make sense to port these tests to the -zkp  modules now that #361 has been merged?

mllwchrry: Good catch, I'll work on a PR for this.

PR #360, opened 2026-04-24
sync-upstream: Simplify

real-or-random:
This is a half-rewrite of the sync-upstream script because I still wasn't happy with it. Looking for early feedback on this.

Advantages:
 - The script does not hardcode any repo URLs, branches, etc. 
 - Nicer separation of responsibilities between the script and the GHA workflow. This brings back the usefulness of the script when used locally outside GHA.
 - Much simpler code, and we could simplify also the GHA workflow.

Still to do:
 - Rewrite the "Tips" section in the generated PR description, e.g., add how to resolve merge conflicts. 
 - Change the GHA workflow accordingly.

### DIFF
--- a/.github/workflows/sync.yml
+++ b/.github/workflows/sync.yml
@@ -9,11 +9,10 @@ jobs:
   sync-upstream:
     runs-on: ubuntu-latest
     env:
-      UPSTREAM: "https://github.com/bitcoin-core/secp256k1.git" 
-      UPSTREAM_BRANCH: "master"
-      ORIGIN_BRANCH: "master"
-      MIN_UPSTREAM_MERGES: 1
-    
+      UPSTREAM: "https://github.com/bitcoin-core/secp256k1.git"
+      BASE_BRANCH: "master"
+      UPSTREAM_REF: "upstream/master"
+
     steps:
       - name: Checkout repository
         uses: actions/checkout@v6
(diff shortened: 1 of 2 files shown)


mllwchrry:
Nice cleanup! A few issues and suggestions:

1. The new pattern replaces the fixed `bitcoin-core/secp256k1` with `.*`, but `.*` will consume up to the last # in the line, capturing the wrong number on commit messages that have multiple # occurrences.
For example, in the lib, there exists a commit message:
`Merge bitcoin-core/secp256k1#1819: tests: Improve tests (Issue #1812)`
New pattern captures 1812 instead of 1819.
I suggest using this: `sed s/'Merge [^#]*#\([0-9]*\).*'/'\1'/`. This way `[^#]*` stops at the first #. And since * already handles the zero-characters case (e.g., `Merge #123`), `\?` is not needed, so omitting it will make sed supported on macOS as well.

2. The generated script sets `BASE`, but the `gh pr create` command references `$BASE_BRANCH`, which is never defined and expands to an empty string. The fix is to change `"\$BASE_BRANCH"` to `"\$BASE"` in the `gh pr create` line in sync-upstream.sh.

3. The branch is created before `$BASE_BRANCH` is validated. `$UPSTREAM_REF` is implicitly validated by the first `git rev-parse` call, so the script exits if it's invalid. But `$BASE_BRANCH` isn't used until `git merge-base`, which runs after the branch is already created. A typo in `$BASE_BRANCH` can lead to an orphaned branch that must be manually deleted before retrying. Although in the case of automatic sync, this may not be that important, the fix is quick and easy, and still worth applying. For this, `RANGESTART_COMMIT=$(git merge-base ...)` has to be moved above the branch creation.

4. Inside the loop, `git log -1` is called twice per commit (once for `PRNUM` and once for the body). Since they are identical, we can store the result once and derive both title and body from it.

real-or-random:
All of this makes sense, thanks! Though at this stage, I was more looking for conceptual feedback. Is this the right way and does it give us the claimed advantages? Any disadvantages you see?

mllwchrry:
Yes, it definitely gives us the advantages you mentioned, and I like the concept. That's why I went ahead and looked at the code more closely as well. As for disadvantages, I don't see any right now. I'm only wondering about placing gh-pr-create.sh in CWD instead of contrib/. What's the reason for that change?

DarkWindman:
Generally, I agree that this concept is more useful! Local sync support is always a plus, just in case we need to do sync earlier than next month.

Also, from what I understand, we are moving the branch creation logic back to the script, right?

real-or-random:
>  1. The new pattern replaces the fixed `bitcoin-core/secp256k1` with `.*`, but `.*` will consume up to the last # in the line, (...) 

fixed, though your (LLM?) suggestion was wrong here.


> 2. The generated script sets `BASE`, but the `gh pr create` command references `$BASE_BRANCH`, which is never defined and expands to an empty string. The fix is to change `"\$BASE_BRANCH"` to `"\$BASE"` in the `gh pr create` line in sync-upstream.sh.

fixed

> 3. (...) A typo in `$BASE_BRANCH` can lead to an orphaned branch that must be manually deleted before retrying. (...) `RANGESTART_COMMIT=$(git merge-base ...)` has to be moved above the branch creation.

fixed 

> 4. Inside the loop, `git log -1` is called twice per commit (once for `PRNUM` and once for the body). Since they are identical, we can store the result once and derive both title and body from it.

I tried this but the result doesn't really look simpler, so I kept it as is.


> Also, from what I understand, we are moving the branch creation logic back to the script, right?

indeed

real-or-random:
I addressed all comments by @mllwchrry and copilot.

real-or-random:
@mllwchrry @DarkWindman Want to re-review and/or test this in one of your forks? I think you have more experience with testing this in a fork? Maybe we can use it already for the June sync.

DarkWindman:
I have tested the sync script both locally and using a PAT. Overall, all jobs are working! 
However, I have one question: the current setup throws an error if there is nothing to merge or the sync PR already exists; but, in the previous version, we just skipped the jobs in these cases. Was this logic removed for simplicity? It is not critical, but if there is nothing to synk, it feels more like a skipped step rather than an actual error. But maybe I am overthinking it...

mllwchrry:
Everything looks good. Found only one issue with local use on macOS. The script uses a heredoc inside `$() (BODY+=$(cat <<EOF ... EOF))`, which bash 3.2 fails to parse:

`./contrib/sync-upstream.sh: line 68: unexpected EOF while looking for matching `)'`

macOS ships bash 3.2 as the default, so `#!/usr/bin/env` bash will pick it up unless the user has a newer bash earlier in PATH.

We could either document the requirement, add a version guard at the top, or split the heredoc into a separate assignment:

USAGE_HINTS=$(cat <<'EOF'
...
EOF
)
BODY+="$USAGE_HINTS"

And I’d like to also re-ask my previous question regarding the placement of the gh-pr-create.sh file:

> I'm only wondering about placing gh-pr-create.sh in CWD instead of contrib/. What's the reason for that change?

real-or-random:
> if there is nothing to synk, it feels more like a skipped step rather than an actual error.

I agree. But I think skipping is what the workflow does. `sync-upstream.sh` exits with 0 if there are no commits, and so the `exit 1` inside the GHA workflow is not hit in this case. (Or which case do you have in mind?)

> The script uses a heredoc inside `$()` [...] We could either document the requirement, add a version guard at the top, or split the heredoc into a separate assignment:

Just to get this right: What's the actual problem with the heredoc? You write that it's "inside `$()`" but in your code example it's still inside "inside `$()`"? Happy to cherry-pick a commit if you have one that you tested on macOS.

> I'm only wondering about placing gh-pr-create.sh in CWD instead of contrib/. What's the reason for that change?

No big reason. It feels a bit more natural to put it in the working dir because it's a "temporary" script anyway. But I don't have a strong opinion here. Do you want me to move it back?

DarkWindman:
> the exit 1 inside the GHA workflow is not hit in this case. (Or which case do you have in mind?)

Yes, I mean this case (as well as when the PR already exists). When I triggered the GHA, the run failed with the following error:
`Run ./gh-pr-create.sh`
`pull request create failed: GraphQL: No commits between master and sync-8363a2d8 (createPullRequest)`
`Error: Process completed with exit code 1.`
It does not cause any real issues except for the email alert. But it does not look good.

real-or-random:
> > the exit 1 inside the GHA workflow is not hit in this case. (Or which case do you have in mind?)
> 
> Yes, I mean this case (as well as when the PR already exists). When I triggered the GHA, the run failed with the following error: `Run ./gh-pr-create.sh` `pull request create failed: GraphQL: No commits between master and sync-8363a2d8 (createPullRequest)` `Error: Process completed with exit code 1.` It does not cause any real issues except for the email alert. But it does not look good.

Okay yes, this shouldn't happen. My intention was that `sync-upstream.sh` does not even create a `gh-pr-create.sh` script in that case. It should exit with 0 and GHA should skip then. Do you have a GHA log?

mllwchrry:
> What's the actual problem with the heredoc?

Sorry for the wrong diagnosis. The issue is not heredoc, but the single quote in `you're`, which confuses bash 3.2's parser. Replacing it with `you are` is sufficient to fix it.

> No big reason. It feels a bit more natural to put it in the working dir because it's a "temporary" script anyway. But I don't have a strong opinion here. Do you want me to move it back?

As for gh-pr-create.sh in CWD, no strong objection, we're just used to it being in contrib/. The only minor annoyance is that it tends to get lost among logs and other files in the repo root, but that's not a good reason to revert it. Keeping it in CWD is a good idea.

DarkWindman:
> Do you have a GHA log?

I hope this is [it](https://github.com/mllwchrry/secp256k1-zkp/actions/runs/26635383270/job/78499858960). Or the full [log](https://productionresultssa12.blob.core.windows.net/actions-results/af8680c6-c99e-4998-8c0e-7ea37bbd80b2/workflow-job-run-b30fe5f9-0e43-5889-af71-75fe5daa3218/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-05-29T15%3A26%3A10Z&sig=yZdOb5Ylu3M8QRmpscWvx6JjBtbI11MuWFYvHzpsxJY%3D&ske=2026-05-29T18%3A00%3A25Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-05-29T14%3A00%3A25Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-05-29T15%3A16%3A05Z&sv=2025-11-05).

real-or-random:
> I hope this is [it](https://github.com/mllwchrry/secp256k1-zkp/actions/runs/26635383270/job/78499858960). Or the full [log](https://productionresultssa12.blob.core.windows.net/actions-results/af8680c6-c99e-4998-8c0e-7ea37bbd80b2/workflow-job-run-b30fe5f9-0e43-5889-af71-75fe5daa3218/logs/job/job-logs.txt?rsct=text%2Fplain&se=2026-05-29T15%3A26%3A10Z&sig=yZdOb5Ylu3M8QRmpscWvx6JjBtbI11MuWFYvHzpsxJY%3D&ske=2026-05-29T18%3A00%3A25Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2026-05-29T14%3A00%3A25Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-11-05&sp=r&spr=https&sr=b&st=2026-05-29T15%3A16%3A05Z&sv=2025-11-05).

I can't reproduce this error locally. The log is from attempt 3 of that run. I believe what happened here is:
 1. You manually triggered a run on master when master was on 77eb1817c797407ecd0fa2da4a6c43e723b892e3. This revision doesn't include upstream PR 1854.
 2. Then master was updated to include upstream PR 1854 by merging https://github.com/mllwchrry/secp256k1-zkp/pulls/30 (or https://github.com/mllwchrry/secp256k1-zkp/pulls/31) .
 3. Then you triggered a re-run. Since it's a re-run, it ran on the same 77eb1817c797407ecd0fa2da4a6c43e723b892e3. That's why `sync-upstream.sh` concluded that there's work to do. But then trying to create a PR with `gh` failed because master had already been updated.

real-or-random:
> Sorry for the wrong diagnosis. The issue is not heredoc, but the single quote in `you're`, which confuses bash 3.2's parser. Replacing it with `you are` is sufficient to fix it.

Should be fixed.

mllwchrry:
ACK c5be752

Issue #359, opened 2026-04-23
ecdsa_adaptor: SHA256 override not always used (follow-up to #357) 


real-or-random:
> Claude also caught a hard-to-catch bug in the #1777 port: `secp256k1_ecdsa_adaptor_encrypt` falls back to a static-context wrapper for nonce generation instead of dispatching to the `_impl` function with the actual context. This way a custom SHA256 compression function set on the context is silently ignored for adaptor signature nonce generation. We discussed it internally, and the fix will be included in this PR.

After the investigation, it turned out that the fix is more involved than expected. The nonce generation for the dleq proof inside `secp256k1_dleq_prove` also goes through the same static context wrapper, so both call sites need to be addressed.
Our fix plan is to restructure where the nonce is computed: moving it out of `dleq_impl.h` into `main_impl.h` and passing a pre-computed scalar to `secp256k1_dleq_prove`. This is required since we decided to apply the approach used for schnorrsig from #1777. The idea is to modify the internal `nonce_function_ecdsa_adaptor_impl` function to take hash context and call it if `noncefp` is NULL, keeping the public API wrapper unchanged and uncalled. The complication is that `dleq_impl.h` is included at the top of `main_impl.h`, so `nonce_function_ecdsa_adaptor_impl` isn't defined yet when `dleq_impl.h` is processed. That's why we decided to move nonce computation to `main_impl.h`. If you have any other thoughts on this, please share. 

_Originally posted by @mllwchrry in https://github.com/BlockstreamResearch/secp256k1-zkp/issues/357#issuecomment-4287777183_

PR #361, opened 2026-04-27
ecdsa_adaptor: make DLEQ nonce generation pluggable

DarkWindman:
Addresses #359.

### DIFF
--- a/src/modules/ecdsa_adaptor/dleq_impl.h
+++ b/src/modules/ecdsa_adaptor/dleq_impl.h
@@ -31,32 +31,6 @@ static void secp256k1_dleq_hash_point(const secp256k1_hash_ctx *hash_ctx, secp25
     secp256k1_sha256_write(hash_ctx, sha, buf, 33);
 }
 
-static int secp256k1_dleq_nonce(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *k, const unsigned char *sk32, const unsigned char *gen2_33, const unsigned char *p1_33, const unsigned char *p2_33, secp256k1_nonce_function_hardened_ecdsa_adaptor noncefp, void *ndata) {
-    secp256k1_sha256 sha;
-    unsigned char buf[32];
-    unsigned char nonce[32];
-
-    if (noncefp == NULL) {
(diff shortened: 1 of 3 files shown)


mllwchrry:
`secp256k1_ecdsa_adaptor_encrypt` is getting quite large because the dleq nonce generation was moved to main_impl.h. A DLEQ nonce helper could be potentially useful. The issue is that it would naturally belong in dleq_impl.h, but can't go there due to the ordering constraint (the root cause leading to the current approach). As an alternative, a separate file (e.g., nonce_impl.h) can be created. This way, we can keep responsibilities clear, but it will require more restructuring.
@real-or-random, could you please share your thoughts on the approach you prefer?

real-or-random:
> The issue is that it would naturally belong in dleq_impl.h, but can't go there due to the ordering constraint (the root cause leading to the current approach).

The usual approach in C is to declare the function without defining it. Then you call it. The actual definition can come later.

real-or-random (approved):
utACK https://github.com/BlockstreamResearch/secp256k1-zkp/pull/361/changes/b36c4ab71797907b0f83d1549c25cbca27906e97

mllwchrry:
@real-or-random, thank you! We overcomplicated the fix before.

real-or-random:
> The usual approach in C is to declare the function without defining it. Then you call it. The actual definition can come later.

To expand on this a bit (at the risk of explaining stuff you already know):

What is often done in C is to have header files even for internal function belonging to the same translation unit, and then you don't have this issue because the header files contain all the declarations. This is what we do for "internal" modules, e.g., `hash.h` is the header file for `hash_impl.h`. (And a normal naming convention would name the latter simply `hash.c`; this is a bit special in libsecp256k1.) 

For some reason (laziness?), this convention hasn't carried over to API modules (the stuff in `modules`). These only have `_impl.h`. And that's why it might be necessary to have these extra forward declarations you now introduced here. The cleanest way would be to refactor all modules to have proper headers, or at least those modules for which we have multiple `_impl.h` files. But as long as missing declarations are not a problem we run into very often, I lean towards continuing to err on the side of laziness...

PR #365, opened 2026-06-03
tests: Port hash context tests to zkp modules

mllwchrry:
Ports the hash context tests from bitcoin-core/secp256k1#1777 to the zkp-specific modules.

### DIFF
--- a/src/modules/bppp/tests_impl.h
+++ b/src/modules/bppp/tests_impl.h
@@ -662,6 +662,34 @@ static void norm_arg_test_all(void) {
     norm_arg_test(64, 64);
 }
 
+DEFINE_SHA256_TRANSFORM_PROBE(sha256_bppp)
+static void test_bppp_ctx_sha256(void) {
+    /* Check ctx-provided SHA256 compression override takes effect */
+    secp256k1_context *ctx = secp256k1_context_clone(CTX);
+    unsigned char out_default[66], out_custom[66];
+    secp256k1_bppp_generators *gens;
+    size_t len = sizeof(out_default);
+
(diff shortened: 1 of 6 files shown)


mllwchrry:
I noticed that musig also uses `hash_ctx` but has no equivalent ctx_sha256 test. Is this intentional or an oversight worth fixing in a separate PR to upstream?

real-or-random:
> Is this intentional or an oversight worth fixing in a separate PR to upstream?

I don't think it's intentional? @furszy

furszy:
Yeah, not intentional. There are three API functions that could/should have a custom compression test `secp256k1_musig_pubkey_agg`, `secp256k1_musig_nonce_process` and `secp256k1_musig_nonce_gen`.

If my memory doesn't fail me, the last one is slightly special because the session secret randomness is cleared internally, which will affect the result for the comparison check. But it is just a matter of caching the value and re-seeding it.

Happy to review the PR if you want to tackle it.

mllwchrry:
Thanks for the answer, @furszy! I'll work on the PR.

real-or-random:
@mllwchrry needs rebase.

> Thanks for the answer, @furszy! I'll work on the PR.

 Is this still planned on your side? (Tracking this now in https://github.com/bitcoin-core/secp256k1/issues/1835)

apoelstra:
801467a537a630d8459afeb8e0f379ae52f8e325 needs rebase

apoelstra (approved):
ACK 72867fd682279ff2c79cf13f4f8d8484048d2527; successfully ran local tests

PR #369, opened 2026-08-12
surjection: bind genrand nonce to the full statement

mllwchrry:
`secp256k1_surjection_genrand` derived each per-counter scalar solely from the witness (`blinding_key = r_out − r_in`). Two proofs sharing the same witness under a different statement drew identical nonces, enabling exact recovery of the witness from the pair.

The nonce is consumed by `secp256k1_borromean_sign`, so the fix binds exactly that function's inputs. `genrand` now commits once to the ring keys (`pubs`), the ring size, the honest index, and the message (`msg32`), and derives each s-value from that single statement commitment together with the witness and counter. Mirroring `borromean_sign`'s argument list makes it directly auditable that the whole statement is bound, instead of having to trace each `surjectionproof_generate` argument through `msg32` and `genrand` separately.

A second commit zeroes the full `sec_input` buffer on exit. The previous clear covered only the first 32 bytes, leaving the tail of the witness on the stack.

No format change. Verification is untouched. Previously generated proofs remain valid. The change is prover-side only.

### DIFF
--- a/src/modules/surjection/main_impl.h
+++ b/src/modules/surjection/main_impl.h
@@ -337,7 +337,20 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     rsizes[0] = (int) n_used_pubkeys;
     indices[0] = (int) ring_input_index;
     secp256k1_surjection_genmessage(hash_ctx, msg32, ephemeral_input_tags, n_total_pubkeys, ephemeral_output_tag);
-    if (secp256k1_surjection_genrand(hash_ctx, borromean_s, n_used_pubkeys, &blinding_key) == 0) {
+    /* Derive every s-value, including the one used as the signing nonce, from
+     * every proof-relevant input to
+     * secp256k1_surjectionproof_generate. Except with negligible hash-collision
+     * probability, this prevents distinct proof inputs from reusing any
+     * s-value.
+     *
+     * The proof-relevant arguments to secp256k1_surjectionproof_generate
+     * correspond as follows: proof supplies n_total_pubkeys (proof->n_inputs),
+     * proof->used_inputs, and n_used_pubkeys, while proof->data is output and
+     * proof->initialized is VERIFY-only validation state; ephemeral_input_tags
+     * is committed by msg32; n_ephemeral_input_tags equals n_total_pubkeys as
+     * checked above; ephemeral_output_tag is committed by msg32; input_index
+     * and both blinding keys are passed directly. */
+    if (secp256k1_surjection_genrand(hash_ctx, borromean_s, n_used_pubkeys, n_total_pubkeys, proof->used_inputs, msg32, input_index, input_blinding_key, output_blinding_key) == 0) {
         return 0;
     }
     /* Borromean sign will overwrite one of the s values we just generated, so use
--- a/src/modules/surjection/surjection_impl.h
+++ b/src/modules/surjection/surjection_impl.h
@@ -35,31 +35,52 @@ SECP256K1_INLINE static void secp256k1_surjection_genmessage(const secp256k1_has
     secp256k1_sha256_clear(&sha256_en);
 }
 
-SECP256K1_INLINE static int secp256k1_surjection_genrand(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *s, size_t ns, const secp256k1_scalar *blinding_key) {
+/* Derive the ring's s-values, one of which is used as the signing nonce, from a
+ * seed that hashes the passed-in arguments. See the call site for how these
+ * correspond to the proof inputs. */
+SECP256K1_INLINE static int secp256k1_surjection_genrand(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *s, size_t ns, size_t n_inputs, const unsigned char *used_inputs, const unsigned char *msg32, size_t input_index, const unsigned char *input_blinding_key, const unsigned char *output_blinding_key) {
     size_t i;
-    unsigned char sec_input[36];
+    size_t used_inputs_len;
+    unsigned char n_inputs_ser[4];
+    unsigned char index_ser[4];
+    unsigned char counter[4];
+    unsigned char seed[32];
+    unsigned char out[32];
     secp256k1_sha256 sha256_en;
 
+    used_inputs_len = (n_inputs + 7) / 8;
+    secp256k1_write_be32(n_inputs_ser, (uint32_t)n_inputs);
+    secp256k1_write_be32(index_ser, (uint32_t)input_index);
+
+    /* Hash the arguments into the seed. */
+    secp256k1_sha256_initialize(&sha256_en);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, n_inputs_ser, 4);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, used_inputs, used_inputs_len);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, msg32, 32);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, index_ser, 4);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, input_blinding_key, 32);
+    secp256k1_sha256_write(hash_ctx, &sha256_en, output_blinding_key, 32);
+    secp256k1_sha256_finalize(hash_ctx, &sha256_en, seed);
+    secp256k1_sha256_clear(&sha256_en);
+
     /* compute s values */
-    secp256k1_scalar_get_b32(&sec_input[4], blinding_key);
     for (i = 0; i < ns; i++) {
         int overflow = 0;
-        sec_input[0] = i;
-        sec_input[1] = i >> 8;
-        sec_input[2] = i >> 16;
-        sec_input[3] = i >> 24;
-
+        secp256k1_write_be32(counter, (uint32_t)i);
         secp256k1_sha256_initialize(&sha256_en);
-        secp256k1_sha256_write(hash_ctx, &sha256_en, sec_input, 36);
-        secp256k1_sha256_finalize(hash_ctx, &sha256_en, sec_input);
+        secp256k1_sha256_write(hash_ctx, &sha256_en, counter, 4);
+        secp256k1_sha256_write(hash_ctx, &sha256_en, seed, 32);
+        secp256k1_sha256_finalize(hash_ctx, &sha256_en, out);
         secp256k1_sha256_clear(&sha256_en);
-        secp256k1_scalar_set_b32(&s[i], sec_input, &overflow);
+        secp256k1_scalar_set_b32(&s[i], out, &overflow);
         if (overflow == 1) {
-            secp256k1_memclear_explicit(sec_input, 32);
+            secp256k1_memclear_explicit(out, sizeof(out));
+            secp256k1_memclear_explicit(seed, sizeof(seed));
             return 0;
         }
     }
-    secp256k1_memclear_explicit(sec_input, 32);
+    secp256k1_memclear_explicit(out, sizeof(out));
+    secp256k1_memclear_explicit(seed, sizeof(seed));
     return 1;
 }
 
--- a/src/modules/surjection/tests_impl.h
+++ b/src/modules/surjection/tests_impl.h
@@ -643,6 +643,147 @@ static void test_gen_verify_all(void) {
     test_gen_verify(SECP256K1_SURJECTIONPROOF_MAX_N_INPUTS, SECP256K1_SURJECTIONPROOF_MAX_USED_INPUTS);
 }
 
+static int surjection_genrand_streams_equal(const secp256k1_scalar *a, const secp256k1_scalar *b, size_t n) {
+    size_t i;
+    for (i = 0; i < n; i++) {
+        if (!secp256k1_scalar_eq(&a[i], &b[i])) {
+            return 0;
+        }
+    }
+    return 1;
+}
+
+static int surjection_genrand_stream_all_differ(const secp256k1_scalar *a, const secp256k1_scalar *b, size_t n) {
+    size_t i;
+    for (i = 0; i < n; i++) {
+        if (secp256k1_scalar_eq(&a[i], &b[i])) {
+            return 0;
+        }
+    }
+    return 1;
+}
+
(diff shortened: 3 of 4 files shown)


src/modules/surjection/surjection_impl.h
```diff
@@ -35,11 +35,39 @@ SECP256K1_INLINE static void secp256k1_surjection_genmessage(const secp256k1_has
     secp256k1_sha256_clear(&sha256_en);
 }
 
-SECP256K1_INLINE static int secp256k1_surjection_genrand(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *s, size_t ns, const secp256k1_scalar *blinding_key) {
+/* Generate the random s-values for the ring signature. The nonce (one of these
+ * s-values) must be unique per secp256k1_borromean_sign invocation, so this
+ * binds exactly borromean_sign's inputs: the ring keys (pubs), the ring size
+ * (ns), the honest index (secidx) and the message (msg32). The secret key is
+ * folded in per s-value. Mirroring borromean_sign's argument list makes it
+ * self-evident that the whole statement is committed. */
```
apoelstra: In ef36ab9a802973b34d7f8b79354e8adf7cabba77:

I don't really buy this comment. We aren't mirroring `borromean_sign`'s argument list, nor are we even mirroring the argument list to `secp256k1_surjectionproof_generate` (which is the API function the user calls), so we can't easily argue that we're committing to all the input.

What we should really say is:
* to generate the proof the user provides ephemeral_input_tags, n_ephemeral_input_tags, ephemeral_output_tag, input_index, input_blinding_key, output_blinding_key (also `ctx` which is independent of the proof and `proof` which is an outptr)
* the first three arguments are hashed into `msg32` by `secp256k1_surjection_genmessage` and therefore are committed here
* the last two arguments are subtracted and their difference appears as `blinding_key`; this is okay because only this difference is actually used
* similarly, the user's `input_index` is reduced(?) to compute `ring_input_index` by `secp256k1_surjection_compute_public_keys`, but only the reduced value is actually used

Alternately we could just directly hash `msg32`, `input_index`, `input_blinding_key`, `output_blinding_key`, which would then let us easily argue that we're committing to every bit of input that the user provides, and this function is deterministic, so the nonce can only be repeated if the entire proof is repeated.

apoelstra: I believe this code is correct but it was pretty hard to convince myself of it and the existing comments are distracting at best.

mllwchrry:
Thanks, @apoelstra!

Switched to the direct approach you suggested. `genrand` now hashes every argument of `secp256k1_surjectionproof_generate` into a single statement (`msg32`, `n_inputs`, `used_inputs`, `input_index`, `input_blinding_key`, `output_blinding_key`) and derives each s-value from it, so the nonce repeats only if the whole proof repeats.

I also dropped the second commit (the `sec_input` buffer clear): the new `genrand` no longer has that 36-byte buffer: the secret now lives in the statement hash.

src/modules/surjection/surjection_impl.h
jonasnick: I was about to post a long response that I believe the original version of the PR was much easier to review because `genrand` did pretty much match the `borromean_sign` arguments (and I actually changed `genrands` argument list in my local branch so it matched `borromean_sign` much more closely). This would be sufficient if we only cared about guaranteeing a unique nonce per borromean_sign challenge. However, we want a stronger property, namely that the s-values are different for any two invocations of surjectionproof_generate to avoid linkability. Therefore, I think you suggested the right approach.

src/modules/surjection/main_impl.h
```diff
@@ -337,7 +337,7 @@ int secp256k1_surjectionproof_generate(const secp256k1_context* ctx, secp256k1_s
     rsizes[0] = (int) n_used_pubkeys;
     indices[0] = (int) ring_input_index;
     secp256k1_surjection_genmessage(hash_ctx, msg32, ephemeral_input_tags, n_total_pubkeys, ephemeral_output_tag);
-    if (secp256k1_surjection_genrand(hash_ctx, borromean_s, n_used_pubkeys, &blinding_key) == 0) {
+    if (secp256k1_surjection_genrand(hash_ctx, borromean_s, n_used_pubkeys, msg32, proof->used_inputs, n_total_pubkeys, input_index, input_blinding_key, output_blinding_key) == 0) {
```
jonasnick: I think this can match both the order of surjectionproof_generate arguments, as well as the order of hashing in `genrand` by using
```C
SECP256K1_INLINE static int secp256k1_surjection_genrand(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *s, size_t ns, size_t n_inputs, const unsigned char *used_inputs, const unsigned char *msg32, size_t input_index, const unsigned char *input_blinding_key, const unsigned char *output_blinding_key);
```

Moreover, we should add a comment here that describes 1) what we're trying to achieve and 2) how the arguments correspond to surjectionproof_generate. Something like:
```C
    /* Derive every s-value, including the one used as the signing nonce, from
     * every proof-relevant input to
     * secp256k1_surjectionproof_generate. Except with negligible hash-collision
     * probability, this prevents distinct proof inputs from reusing any
     * s-value.
     *
     * The proof-relevant arguments to secp256k1_surjectionproof_generate
     * correspond as follows: proof supplies n_total_pubkeys (proof->n_inputs),
     * proof->used_inputs, and n_used_pubkeys, while proof->data is output and
     * proof->initialized is VERIFY-only validation state; ephemeral_input_tags
     * is committed by msg32; n_ephemeral_input_tags equals n_total_pubkeys as
     * checked above; ephemeral_output_tag is committed by msg32; input_index
     * and both blinding keys are passed directly. */
```

The doc domment above genrand in surjecitonproof_impl.h can be simplified significantly and shouldn't refer to surjectionproof_generate.

src/modules/surjection/surjection_impl.h
```diff
@@ -35,31 +35,56 @@ SECP256K1_INLINE static void secp256k1_surjection_genmessage(const secp256k1_has
...
+ * directly. Since the derivation is deterministic, the nonce can only repeat if
+ * the entire proof repeats. */
+SECP256K1_INLINE static int secp256k1_surjection_genrand(const secp256k1_hash_ctx *hash_ctx, secp256k1_scalar *s, size_t ns, const unsigned char *msg32, const unsigned char *used_inputs, size_t n_inputs, size_t input_index, const unsigned char *input_blinding_key, const unsigned char *output_blinding_key) {
     size_t i;
-    unsigned char sec_input[36];
+    size_t used_inputs_len;
+    unsigned char n_inputs_ser[4];
+    unsigned char index_ser[4];
+    unsigned char counter[4];
+    unsigned char statement[32];
```
jonasnick: Better to call this `seed` because it contains not only the statement but also the witness.

src/modules/surjection/surjection_impl.h
```diff
@@ -35,31 +35,56 @@ SECP256K1_INLINE static void secp256k1_surjection_genmessage(const secp256k1_has
...
+    unsigned char out[32];
     secp256k1_sha256 sha256_en;
 
+    used_inputs_len = (n_inputs + 7) / 8;
+    secp256k1_write_be32(n_inputs_ser, (uint32_t)n_inputs);
+    secp256k1_write_be32(index_ser, (uint32_t)input_index);
+
+    /* Commit once to every argument of secp256k1_surjectionproof_generate. */
```
jonasnick: Better: "Hash the arguments into the seed".

src/modules/surjection/tests_impl.h
```diff
@@ -643,6 +643,86 @@ static void test_gen_verify_all(void) {
...
+    return 1;
+}
+
+/* Regression test for the surjection-proof nonce-reuse fix.
+ *
+ * secp256k1_surjection_genrand produces the ring's s-values, one of which is
+ * repurposed as the signing nonce. It commits to every argument of
+ * secp256k1_surjectionproof_generate, so two proofs draw the same s-value stream
+ * (reusing the nonce) only when every input is identical. We vary each committed
+ * argument in turn and check the stream changes. */
```
jonasnick: We can test a stronger property: not only that the streams are unequal, but that every individual element is unequal using a function like this:

```C
static int surjection_genrand_stream_all_differ(const secp256k1_scalar *a, const secp256k1_scalar *b, size_t n) {
    size_t i;
    for (i = 0; i < n; i++) {
        if (secp256k1_scalar_eq(&a[i], &b[i])) {
            return 0;
        }
    }
    return 1;
}
```

The doc comment would need to be updated accordingly and we can simplify it (no need to mention that this is a regression test):
```C
/* Test that changing any proof-relevant argument to
 * secp256k1_surjectionproof_generate changes every s-value produced by
 * secp256k1_surjection_genrand. */
```

src/modules/surjection/tests_impl.h
```diff
@@ -643,6 +643,86 @@ static void test_gen_verify_all(void) {
...
+static void test_surjection_genrand_binds_statement(void) {
+    const secp256k1_hash_ctx *hash_ctx = secp256k1_get_hash_context(CTX);
+    const size_t ns = 2;
```
jonasnick: Better to use 4 because that corresopnds with our used_input bitmaps.

src/modules/surjection/tests_impl.h
```diff
@@ -643,6 +643,86 @@ static void test_gen_verify_all(void) {
...
+}
+
+/* Regression test for the surjection-proof nonce-reuse fix.
+ *
+ * secp256k1_surjection_genrand produces the ring's s-values, one of which is
+ * repurposed as the signing nonce. It commits to every argument of
+ * secp256k1_surjectionproof_generate, so two proofs draw the same s-value stream
+ * (reusing the nonce) only when every input is identical. We vary each committed
+ * argument in turn and check the stream changes. */
+static void test_surjection_genrand_binds_statement(void) {
```
jonasnick: This is not about the statement anymore and "binds" sounds a bit strange. Maybe just `test_surjection_genrand`.

src/modules/surjection/tests_impl.h
```diff
@@ -651,6 +731,7 @@ static const struct tf_test_entry tests_surjection[] = {
     CASE1(test_input_selection_all),
     CASE1(test_input_selection_distribution),
     CASE1(test_gen_verify_all),
+    CASE1(test_surjection_genrand_binds_statement),
```
jonasnick: We could also add a minimal test for surjectionproof_generate, like this:
```C
/* Changing an unused input tag changes the proof message. Check that every
 * s-value changes and that both proofs verify. */
static void test_surjectionproof_generate_changes_s_values(void) {
    const size_t n_inputs = 3;
    const size_t n_used = 2;
    unsigned char seed[32];
    unsigned char input_blinding_key[3][32] = {{ 0 }};
    unsigned char output_blinding_key[32] = { 0 };
    unsigned char reblind[32] = { 0 };
    secp256k1_fixed_asset_tag fixed_input_tags[3];
    secp256k1_generator ephemeral_input_tags[3];
    secp256k1_generator ephemeral_output_tag;
    secp256k1_surjectionproof proof_a;
    secp256k1_surjectionproof proof_b;
    size_t input_index;
    size_t unused_index = n_inputs;
    size_t i;

    testrand256(seed);
    for (i = 0; i < n_inputs; i++) {
        testrand256(fixed_input_tags[i].data);
        input_blinding_key[i][31] = (unsigned char)i + 1;
        CHECK(secp256k1_generator_generate_blinded(CTX, &ephemeral_input_tags[i], fixed_input_tags[i].data, input_blinding_key[i]));
    }
    output_blinding_key[31] = 4;
    reblind[31] = 5;
    CHECK(secp256k1_generator_generate_blinded(CTX, &ephemeral_output_tag, fixed_input_tags[1].data, output_blinding_key));

    CHECK(secp256k1_surjectionproof_initialize(CTX, &proof_a, &input_index, fixed_input_tags, n_inputs, n_used, &fixed_input_tags[1], 100, seed) > 0);
    CHECK(input_index == 1);
    CHECK(secp256k1_surjectionproof_n_used_inputs(CTX, &proof_a) == n_used);
    proof_b = proof_a;

    for (i = 0; i < n_inputs; i++) {
        if (!(proof_a.used_inputs[i / 8] & (1 << (i % 8)))) {
            unused_index = i;
        }
    }
    CHECK(unused_index < n_inputs);

    CHECK(secp256k1_surjectionproof_generate(CTX, &proof_a, ephemeral_input_tags, n_inputs, &ephemeral_output_tag, input_index, input_blinding_key[input_index], output_blinding_key) == 1);
    CHECK(secp256k1_surjectionproof_verify(CTX, &proof_a, ephemeral_input_tags, n_inputs, &ephemeral_output_tag) == 1);

    /* Regenerate the unused ephemeral input tag with reblind instead of
     * input_blinding_key[unused_index]. The modified tag is the only proof input
     * that differs between the two secp256k1_surjectionproof_generate calls. */
    CHECK(secp256k1_generator_generate_blinded(CTX, &ephemeral_input_tags[unused_index], fixed_input_tags[unused_index].data, reblind));
    CHECK(secp256k1_surjectionproof_generate(CTX, &proof_b, ephemeral_input_tags, n_inputs, &ephemeral_output_tag, input_index, input_blinding_key[input_index], output_blinding_key) == 1);
    CHECK(secp256k1_surjectionproof_verify(CTX, &proof_b, ephemeral_input_tags, n_inputs, &ephemeral_output_tag) == 1);

    for (i = 0; i < n_used; i++) {
        CHECK(secp256k1_memcmp_var(&proof_a.data[32 + 32 * i], &proof_b.data[32 + 32 * i], 32) != 0);
    }
}
```

jonasnick (commented):
I suggest a more descriptive commit message that describes the problem being fixed:
```
surjection: prevent s-value reuse for different proof inputs

The s-values produced by secp256k1_surjection_genrand previously
depended only on their indices and the difference between
input_blinding_key and output_blinding_key. Calls with the same
difference therefore reused s-values even when their proof inputs
differed.

For proofs with the same used-input selection and honest input index,
the same generated s-value was used as the signing nonce. Reusing this
nonce across different proof messages allowed recovery of the
blinding-key difference. The remaining repeated s-values also revealed
the honest input index.

This commit affects proof generation only; verification is unchanged.
```

jonasnick (approved):
ACK 65093e1444e8e90e0cf3166633602bf17d2dd028

PR #370, opened 2026-08-12
rangeproof: warn that nonce must not be reused across differing arguments

mllwchrry:
`secp256k1_rangeproof_genrand` seeds RFC6979 with nonce, commit, genp, and the proof header. `extra_commit` and message are not in the seed. Changing either while holding the nonce fixed changes every challenge but leaves every per-ring nonce (`k_i`) unchanged.

No derivation change can close this: rewind reconstructs the prover's stream before it knows the message, so the message can never enter the seed without breaking rewind. The fix is to state the constraint precisely in the API.

### DIFF
--- a/include/secp256k1_rangeproof.h
+++ b/include/secp256k1_rangeproof.h
@@ -125,7 +125,14 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_rangeproof_rewind(
  *          commit: the commitment being proved.
  *          blind:  32-byte blinding factor used by commit. The blinding factor may be all-zeros as long as min_bits is set to 3 or greater.
  *                  This is a side-effect of the underlying crypto, not a deliberate API choice, but it may be useful when balancing CT transactions.
- *          nonce:  32-byte secret nonce used to initialize the proof (value can be reverse-engineered out of the proof if this secret is known.)
+ *          nonce:  32-byte secret nonce used to initialize the proof.
+ *
+ *                  Each call to this function must have a UNIQUE nonce that
+ *                  MUST NOT BE REUSED in subsequent calls. The nonce must be
+ *                  KEPT SECRET except from parties authorized to rewind the
+ *                  proof. Anyone who knows the nonce can recover `value` and
+ *                  `blind` from the proof. Reusing the nonce may expose `blind`
+ *                  even to parties that do not know the nonce.
  *          exp:    Base-10 exponent. Digits below above will be made public, but the proof will be made smaller. Allowed range is -1 to 18.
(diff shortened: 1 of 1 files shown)


apoelstra:
Yeah, this looks good to me. Frustratingly we *could* stick `extra_commit` in there, but it'd break existing wallets which would become unable to decrypt new messages and vice-versa.

But it's true that, in general, we are producing these rangeproofs determistically then xoring a message into them. So if you call the function twice with same nonce but different messages, you will have a classic "reusing an OTP" attack where the attacker can just xor the two rangeproofs and extract the xor of the two messages.

So even if we stuck `extra_commit` in there (which I am not recommending we do) we'd still need this warning.

apoelstra (approved):
ACK 92603e2b5f0eb5e5327a408ec9ce8c8ab1304334; successfully ran local tests

include/secp256k1_rangeproof.h
```diff
@@ -125,7 +125,16 @@ SECP256K1_API SECP256K1_WARN_UNUSED_RESULT int secp256k1_rangeproof_rewind(
...
+ *          nonce:  32-byte secret nonce used to initialize the proof (value can be
+ *                  reverse-engineered out of the proof if this secret is known).
+ *
+ *                  The nonce must not be reused across two calls whose arguments differ in any
+ *                  way. The nonce is not derived from extra_commit or message, so two
+ *                  proofs for the same commitment under the same nonce but a different
+ *                  extra_commit or message reuse a nonce across differing challenges,
+ *                  and anyone holding both proofs recovers 'blind' exactly. Calling
+ *                  again with identical arguments is safe: the derivation is
+ *                  deterministic and reproduces the same proof.
```
jonasnick: This current text tries to characterize precisely when reuse is unsafe, but I don't think we should make that guarantee in the API. I'd prefer being more direct and giving the caller maximally precise instructions:
```
nonce:  32-byte secret nonce used to initialize the proof.

        Each call to this function must have a UNIQUE nonce that
        MUST NOT BE REUSED in subsequent calls. The nonce must be
        KEPT SECRET except from parties authorized to rewind the
        proof. Anyone who knows the nonce can recover `value` and
        `blind` from the proof. Reusing the nonce may expose `blind`
        even to parties that do not know the nonce.
```
This roughly matches the wording about nonces in other parts of the lib (in particular musig_nonce_gen).

apoelstra (approved):
ACK cde28971a2274beeeb2e9df27342d956b5664d44; successfully ran local tests

jonasnick (approved):
ACK cde28971a2274beeeb2e9df27342d956b5664d44

PR #371, opened 2026-08-14
whitelist, bppp: defensive hardening and documentation

DarkWindman:
Small defensive fixes + docs for whitelist/bppp:

- bppp: overflow check in `generators_create` allocation.
- whitelist: signature_parse always leaves `sig` in a valid-or-canonically-invalid state; guard serialize against untrusted `n_keys`.
- whitelist: document that `n_keys` is attacker-controlled and must not be used to size the arrays passed to `_verify`.
- whitelist: document the accepted degenerate `W = -P_i` destination.

No API/format changes. New tests cover theallocation overflow and the parse-failure guarantee.

### DIFF
--- a/include/secp256k1_whitelist.h
+++ b/include/secp256k1_whitelist.h
@@ -68,6 +68,10 @@ SECP256K1_API int secp256k1_whitelist_signature_parse(
  *
  *  Returns: the number of keys for the given signature
  *  In: sig: pointer to a signature object
+ *
+ *  This count is a property of the signature only. It might not match the
+ *  number of public keys the caller has, and the caller should not truncate
+ *  their key set to match it.
  */
 SECP256K1_API size_t secp256k1_whitelist_signature_n_keys(
     const secp256k1_whitelist_signature *sig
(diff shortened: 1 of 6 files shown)


include/secp256k1_whitelist.h
```diff
@@ -68,6 +68,12 @@ SECP256K1_API int secp256k1_whitelist_signature_parse(
  *
  *  Returns: the number of keys for the given signature
  *  In: sig: pointer to a signature object
+ *
+ *  The returned count comes from the parsed input and is therefore
```
apoelstra: In b9d43ad596b57f0206392d12fc8a16d83ed3c5ed:

This comment is kinda silly. Obviously a function that takes `sig` and returns a `size_t` is getting the `size_t` out of the signature.

At best, the comment should be rewritten to just remind the reader that this key count is a property of the signature only, that it might not match the count of the public keys, and that (obviously) you should not truncate your key set to match the count of keys. But I don't think even that's necessary.

As written this comment is distracting because it sounds like a serious warning but actually it's saying nothing at all.

DarkWindman: Fair point, fixed in commit 7ea12c2dca07b82520f396fb10a76b128f232282.

We can leave this short comment as a reminder, but if you think it's unnecessary, I can just remove it.

apoelstra (approved):
ACK 598e22dcde2152003c6a60f04196cc5fe2e2a090; successfully ran local tests

mllwchrry:
ACK 598e22d.

PR #373, opened 2026-09-09
rangeproof: expose secp256k1_borromean_verify as public API

fametrano:
btclib-org/btclib needs to verify that its own Borromean ring signature implementation interoperates with this library's. That claim (btclib-org/btclib#1853) currently can't be checked against real code: the verification primitive, secp256k1_borromean_verify, is static inside the rangeproof module and has no public entry point. The only way to check interoperability today is to read borromean_impl.h by hand and transcribe its constants.

This renames the internal function to secp256k1_borromean_verify_impl (unchanged otherwise) and adds a public secp256k1_borromean_verify that validates its arguments and delegates to it — the same split this module already uses for secp256k1_rangeproof_verify / secp256k1_rangeproof_verify_impl. It is verify-only: secp256k1_borromean_sign is untouched and stays internal, since a public signer would need its own API design (nonce handling, misuse-resistance) that is out of scope here.

A verify-only oracle for this primitive is useful to any consumer of this library building on Borromean ring signatures, not just btclib — it lets them test their own implementation against this one without depending on rangeproof's proof format.

On why this is verify-only rather than a sign/verify pair: verification is a predicate over public data — no secret key, no nonce, no constant-time requirement — so it is safe to expose as a stable entry point. A public signer is deliberately left out, not merely deferred: secp256k1_borromean_sign takes its nonces directly, so exposing it would first require a nonce-generation API and the misuse-resistance that comes with it. That is a larger design question, and a different one from secp256k1_rangeproof_sign, whose nonce is derived internally — so the asymmetry with the rangeproof pair is intentional. The motivating use case does not need a signer either: a fixed, checked-in signature verifies in both directions, so verify-only already covers interoperability testing without a public entry point that produces signatures.

build/bin/tests passes, including a new test_borromean_verify_api that signs with the existing secp256k1_borromean_sign over random rings, verifies through the new public API, and cross-checks the result against a direct call to secp256k1_borromean_verify_impl.

🤖 Generated with [Claude Code](https://claude.com/claude-code), under the close supervision of the human fametrano, who declares full accountability for it.


fametrano:
The one red check here is `ARM64: macOS Sonoma`, and it looks unrelated to this PR. Could someone re-run it?

The failure is in the surjection module:

```
src/modules/surjection/tests_impl.h:222: test condition failed: result > 0
FAIL tests (exit status: 134)
```

That line is in `test_input_selection_distribution_helper`, which calls `secp256k1_surjectionproof_initialize` 10000 times on a fresh `testrand256(seed)` and asserts the result is positive, so it can fail on an unlucky seed.

This PR only renames the internal `secp256k1_borromean_verify` to `..._verify_impl` and updates its three call sites, all on verify paths; `surjectionproof_initialize` does not call borromean at all. The other 101 checks pass, and master was green on 2026-09-14.

fametrano:
Rebased onto master (`8e1f96c`). Three lines had to follow upstream renames: `secp256k1_get_hash_context(ctx)` is now `&ctx->hash_ctx` (c8745f6b) in `rangeproof/main_impl.h` and `tests_impl.h`, and `secp256k1_eckey_pubkey_serialize33` is now `secp256k1_ge_serialize33` (2577fb18) in `tests_impl.h`. The CHANGELOG entry moved back under `[Unreleased]` after the 0.8.0 release. Nothing else changed.

CI is green on the new tip, all 122 jobs including ARM64 macOS, so the re-run I asked for above is no longer needed. Ready for review.

PR #368, opened 2026-08-01
Upstream PRs 1887, 1889, 1894, 1890, 1765, 1896, 1899, 1776, 1869, 1886, 1901, 1900, 1863, 1902, 1709

real-or-random:
*Note: This PR has been created by a GitHub Actions workflow without human involvement.*

This PR syncs the following upstream PRs:
 * bitcoin-core/secp256k1#1887: Make theStack a maintainer and a security contact
 * bitcoin-core/secp256k1#1889: field: serialize elements by word
 * bitcoin-core/secp256k1#1894: extrakeys: check invariant that x-only pubkeys have even Y
 * bitcoin-core/secp256k1#1890: nonce: terminate RFC6979 loop at UINT_MAX
 * bitcoin-core/secp256k1#1765: Add "silentpayments" module implementing BIP352 (take 4, limited to full-node scanning)
 * bitcoin-core/secp256k1#1896: silentpayments: address #1765 follow-ups, add changelog entry
 * bitcoin-core/secp256k1#1899: changelog: add missing entries for #1777 and #1860
 * bitcoin-core/secp256k1#1776: Remove deprecated `secp256k1_schnorrsig_sign` alias
 * bitcoin-core/secp256k1#1869: Remove SECP256K1_GNUC_PREREQ macro
 * bitcoin-core/secp256k1#1886: Remove deprecated `secp256k1_context_no_precomp` pointer
 * bitcoin-core/secp256k1#1901: changelog: Add entry for #1869
 * bitcoin-core/secp256k1#1900: Use SHA256 override for pointers to known aux functions
 * bitcoin-core/secp256k1#1863: ellswift: don't declassify or leave sk in sha256 buffer
 * bitcoin-core/secp256k1#1902: changelog: add entry for #1859
 * bitcoin-core/secp256k1#1709: release-process: Add signing of tarball plus minor improvements

Usage hints:
 * If this PR has merge conflicts, resolve these by switching to the PR branch and merging the base branch into it using `git merge <base-branch>`.
 * To show the conflict resolution diff from an existing merge commit, use `git show --remerge-diff <merge-commit>`.
 * In case you are recreating the PR branch locally, you can (during the conflict resolution state) replay this conflict resolution diff using `git read-tree --reset -u <merge-commit>`.
   Be aware that this may discard your index as well as the uncommitted changes and untracked files in your worktree.

PR #374, opened 2026-09-15
tests: Add hash context tests for generator, whitelist, ecdsa_s2c

mllwchrry:
Adds the ctx-provided SHA256 compression tests (upstream bitcoin-core/secp256k1#1777, ported to zkp modules in #365) to the zkp-only modules that were still missing them: `generator`, `whitelist`, and `ecdsa_s2c`'s anti-exfil signer commit.

`ecdsa_s2c` already had `test_ecdsa_s2c_ctx_sha256`, but that exercises `ecdsa_s2c_sign`, whose nonce goes through the shared core signing routine. `anti_exfil_signer_commit` generates its nonce via a separate call, so it needs its own guard.

`whitelist` can't use the usual perturbing-probe pattern: its message and secret key are themselves derived through `hash_ctx`, so "output differs" holds even if the nonce hashing bypasses the caller's ctx. The test instead uses a non-perturbing probe that counts compressions and pins the exact count (104 for the fixed inputs).


mllwchrry:
Want input on the ctx-SHA256 override tests:
- The #1777/#365 pattern is module-level: it proves a module routes some hashing through the caller's ctx.
- `whitelist` (my recent test) is strict: it guards that the nonce specifically routes through the ctx (#368 fix), via a compression-count check (a perturbing probe can't isolate the nonce there).
- `ecdsa_adaptor` is the same shape but only has the module-level test (#365): I verified it still passes when the DLEQ nonce (#361 fix) is diverted to the static ctx, so it doesn't guard that fix.

Should we use defense-in-depth (make adaptor strict like whitelist), or keep it module-level (make whitelist match the rest)? Leaning defense-in-depth since these are nonce paths.

apoelstra:
My feeling is that we should just do the "normal" test for the whitelist. It's true that we invoke the hash context multiple times, so if we forget to do it once then the test will still pass. But this seems generally true.

A better strategy, if we're worried about it, is to generalize `DEFINE_SHA256_TRANSFORM_PROBE`  so that it only tweaks the final compression round (or the nth round, where n can be specified, or a random round up to the `nth`, or something). And this change we should push upstream.

Meanwhile for our code let's just use the existing infrastructure.

